PR #5, opened 2016-09-11
Rewrite concat to start producers iteratively

mdiep:
Recursively starting the inner producers can blow the stack if they
complete immediately.

Reopened from https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3161.

### DIFF
--- a/Sources/Flatten.swift
+++ b/Sources/Flatten.swift
@@ -341,25 +341,55 @@ extension SignalProtocol where Value: SignalProducerProtocol, Error == Value.Err
 			return disposable
 		}
 	}
-
+	
 	fileprivate func observeConcat(_ observer: Observer<Value.Value, Error>, _ disposable: CompositeDisposable? = nil) -> Disposable? {
-		let state = ConcatState(observer: observer, disposable: disposable)
+		let state = Atomic(ConcatState<Value.Value, Error>())
+		
+		func startNextIfNeeded() {
+			while let producer = state.modify({ $0.dequeue() }) {
+				producer.startWithSignal { signal, inner in
+					let handle = disposable?.add(inner) ?? nil
 
-		return self.observe { event in
+					signal.observe { event in
+						switch event {
+						case .completed, .interrupted:
+							handle?.remove()
+							
+							let shouldStart: Bool = state.modify {
+								$0.active = nil
+								return !$0.isStarting
+							}
+
+							if shouldStart {
+								startNextIfNeeded()
+							}
+
+						case .next, .failed:
+							observer.action(event)
+						}
+					}
+				}
+				state.modify { $0.isStarting = false }
+			}
+		}
+		
+		return observe { event in
 			switch event {
 			case let .next(value):
-				state.enqueueSignalProducer(value.producer)
+				state.modify { $0.queue.append(value.producer) }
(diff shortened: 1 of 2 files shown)


Sources/Flatten.swift
```diff
@@ -341,25 +341,55 @@ extension SignalProtocol where Value: SignalProducerProtocol, Error == Value.Err
...
-
+	
 	fileprivate func observeConcat(_ observer: Observer<Value.Value, Error>, _ disposable: CompositeDisposable? = nil) -> Disposable? {
-		let state = ConcatState(observer: observer, disposable: disposable)
+		let state = Atomic(ConcatState<Value.Value, Error>())
+		
+		func startNextIfNeeded() {
+			while let producer = state.modify({ $0.dequeue() }) {
+				producer.startWithSignal { signal, inner in
+					let handle = disposable?.add(inner) ?? nil
```
NachoSoto: Is `?? nil` needed here

Sources/Flatten.swift
```diff
@@ -341,25 +341,55 @@ extension SignalProtocol where Value: SignalProducerProtocol, Error == Value.Err
...
 
 			case .completed:
-				// Add one last producer to the queue, whose sole job is to
-				// "turn out the lights" by completing `observer`.
-				state.enqueueSignalProducer(SignalProducer.empty.on(completed: {
-					observer.sendCompleted()
-				}))
-
+				state.modify { state in
+					state.queue.append(SignalProducer.empty.on(completed: {
```
NachoSoto: This can be `completed: observer.sendCompleted`.

Tests/ReactiveSwiftTests/SignalProducerSpec.swift
```diff
@@ -1067,6 +1067,27 @@ class SignalProducerSpec: QuickSpec {
 					outerObserver.sendFailed(TestError.default)
 					expect(error) == TestError.default
 				}
+				
+				it("should not overflow the stack if inner producers complete immediately") {
+					typealias Inner = SignalProducer<(), NoError>
+					
+					let depth = 10000
```
NachoSoto: Perfect

NachoSoto:
:shipit: The test is perfect, and the implementation seems reasonable!

Sources/Flatten.swift
mdiep: lol, no. Not sure what I was thinking here. 😆

NachoSoto: If it's a nested optional I can see why it would be needed, but I'd solve that by declaring the type instead I think.

craigzour:
@NachoSoto Is there any plan to release a new ReactiveCocoa version with this fix ?

mdiep:
ReactiveSwift 1.0.0-alpha.1 has this fix.

craigzour:
@mdiep Yeah I noticed that but I wanted to know if we were going to have a ReactiveCocoa version (which targets XCode8 and Swift2.3) with this fix.

mdiep:
There's no plan to back port this to RAC 4.x. It will be included with RAC 5.x since that depends on ReactiveSwift.

craigzour:
@mdiep Alright ! Thank you :)

PR #8, opened 2016-09-12
Rename `next` to `value`

mdiep:
As discussed in https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3013

I rather like this. I don’t think `next` is a particularly great name. `value` seems much better to me.

`Observer.sendValue()` becomes `Observer.send(value: )`
`Observer.sendFailed()` becomes `Observer.send(error: )`
`SignalProducer.startWithNext` becomes `SignalProducer.startWithValues`
`SignalProducer.observeNext` becomes `SignalProducer.startWithValues`

### DIFF
--- a/Sources/Property.swift
+++ b/Sources/Property.swift
@@ -502,7 +502,7 @@ public final class Property<Value>: PropertyProtocol {
 		let producer = unsafeProducer.replayLazily(upTo: 1)
 
 		let atomic = Atomic<Value?>(nil)
-		disposable = producer.startWithNext { atomic.value = $0 }
+		disposable = producer.startWithValues { atomic.value = $0 }
 
 		// Verify that an initial is sent. This is friendlier than deadlocking
 		// in the event that one isn't.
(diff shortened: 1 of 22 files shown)


NachoSoto:
I feel like we're better off leaving breaking changes like this for a version after the initial one for `ReactiveSwift`. Otherwise it's going to make he transition even more painful.

andersio:
This should be easily migrated by API renaming.

andersio:
What about shortening the rest of `send*` to `complete` and `interrupt`?

mdiep:
Yeah, the API renaming makes the transition pretty easy IMO.

neilpa:
I would also think it's better to batch up all the possible (even if it's slightly more) than spreading it out into multiple changes.

liscio:
I'll add my vote to the "break everything now rather than later" pile. Seeing how it's already a huge job to port the growing codebase that I have to swift 3, I might as well get all this renaming handled together in one shot. 😛

mdiep:
Seems like people are mostly on board with making all the breaking changes now.

Is everyone 👍🏻 on the change itself?

Sources/Property.swift
```diff
@@ -599,7 +599,7 @@ public final class MutableProperty<Value>: MutablePropertyProtocol {
 		/// underlying producer prevents sending recursive events.
 		atomic = RecursiveAtomic(initialValue,
 		                          name: "org.reactivecocoa.ReactiveSwift.MutableProperty",
-		                          didSet: observer.sendNext)
+		                          didSet: observer.send(value:))
```
NachoSoto: :+1:

NachoSoto:
Ship it then! I'm also on board.

mdiep:
🎉

PR #22, opened 2016-09-19
Add initialisers to create an Action from an input property

### LINKED ISSUE #88: Action: What is it good for
_This has all spawned from @sharplet's #22._

`Action` is based on RAC 2's `RACCommand`, which is itself based on [ReactiveUI](https://github.com/reactiveui/ReactiveUI)'s `ReactiveCommand`. The model was originally created to model UI actions—which are serial and occur on the main thread. But `Action` was meant to be decoupled from this dependency. The RAC 3 changelog says:

> Unlike commands, actions are not bound to or dependent upon the main thread, making it easier to reason about when they can be executed and when they will generate notifications.

I think this is incomplete: `Action`s aren't truly thread safe.

The `enabledIf` property is one of the key features of actions. However, that property is inherently racey with the body of the action itself. If the property is altered on a different thread/scheduler, it can switch to `false` after `Action` checks the value but before the body of the action executes. You can't trust that the world hasn't changed in the execute closure.

`RACCommand` and ReactiveUI's `ReactiveCommand` don't suffer from this issue because they're tied to the main thread. Altering the world on a background thread is prohibited.

I think we should do something to address this: either explicitly document this limitation or modify `Action` so that it's truly thread safe and reactive.

One we could achieve the later would be to base `Action`'s inputs on signals. Currently, `Action`s often read external properties (i.e. state), rather than relying solely on their input. In the typical login example, the login action would read the username and password properties directly from the view model:

```swift
class LoginViewModel {
    let username = Property<String>()
    let password = Property<String>()
    let action: Action<(), (), LoginError>

    init() {
        let enabled = Property
            .combineLatest(username, password)
            .map { !$0.isEmpty && !$1.isEmpty }
        action = Action(enabledIf: enabled) { _ in
          let username = self.username // These could be empty, even though they should be disabled
          let password = self.password // The perils of reading state
          …
        }
    }
}

// elsewhere
loginViewModel.action.apply(())
```

But actions could be restructured like so:

```

class LoginViewModel {
    let username = Property<String>()
    let password = Property<String>()
    let action: Action<(), LoginError> // input type isn't part of the type

    init() {
        let input = Property.combineLatest(username, password)
        action = Action(input, enabledIf: { !$0.isEmpty && !$1.isEmpty }) { (username, password) in
          …
        }
    }
}


// elsewhere
loginViewModel.action.apply()
```

This would have a few benefits that I see:

1. Value consistency between `enabledIf` and `execute`
1. `apply()` can easily be bound to a trigger signal
1. `Action`s become usable only in the reactive model

But this does require (1) removing the input type from `Action` and (2) removing the input parameter from `apply()`. (2) could be especially problematic, depending on how people are currently using `Action`. But in the standard bind-this-action-to-a-button case, this model seems like it would be simpler.

I'm curious what people think. @ReactiveCocoa/reactiveswift

sharplet:
**Update:** Closes #88.

---

`PropertyAction` encapsulates the pattern where an `Action` should be enabled only if its input is non-nil.

The prime example of this pattern is form validation: validate some input text, and store the validated results in a property. If the form is valid, the property will be non-nil, and it should be safe to grab the current value and submit the form.

I've personally written a bunch of `Action`-based code that basically just implemented this logic in an ad-hoc fashion, so I think it would be super useful to have a convenient shorthand for it.

---

Ideally this could just be a convenience initialiser on `Action`:

```
extension Action where Input == Void {
  // ...
}
```

But unfortunately this triggers Swift's "same-type requirement makes generic parameter non-generic" error.

### DIFF
--- a/Sources/Action.swift
+++ b/Sources/Action.swift
@@ -104,7 +104,7 @@ public final class Action<Input, Output, Error: Swift.Error> {
 	}
 
 	/// Initializes an action that will be conditionally enabled, and creates a
-	/// SignalProducer for each input.
+	/// `SignalProducer` for each input.
 	///
 	/// - parameters:
 	///   - enabledIf: Boolean property that shows whether the action is
@@ -261,15 +261,15 @@ extension Action: ActionProtocol {
 }
 
 extension ActionProtocol where Input == Void {
-	/// Initializes an action that uses an optional property for its input,
-	/// and is disabled whenever the input is nil. When executed, a SignalProducer
+	/// Initializes an action that uses an `Optional` property for its input,
+	/// and is disabled whenever the input is `nil`. When executed, a `SignalProducer`
 	/// is created with the current value of the input.
 	///
 	/// - parameters:
-	///   - input: An Optional property whose current value is used as input
+	///   - input: An `Optional` property whose current value is used as input
 	///            whenever the action is executed. The action is disabled
-	///            whenever the value is nil.
-	///   - execute: A closure to return a new SignalProducer based on the
+	///            whenever the value is `nil`.
+	///   - execute: A closure to return a new `SignalProducer` based on the
 	///              current value of `input`.
 	public init<P: PropertyProtocol, T>(input: P, _ execute: @escaping (T) -> SignalProducer<Output, Error>) where P.Value == T? {
 		self.init(state: input, enabledIf: { $0 != nil }) { input, _ in
@@ -278,12 +278,12 @@ extension ActionProtocol where Input == Void {
 	}
 
 	/// Initializes an action that uses a property for its input. When executed,
-	/// a SignalProducer is created with the current value of the input.
+	/// a `SignalProducer` is created with the current value of the input.
 	///
 	/// - parameters:
 	///   - input: A property whose current value is used as input
 	///            whenever the action is executed.
-	///   - execute: A closure to return a new SignalProducer based on the
+	///   - execute: A closure to return a new `SignalProducer` based on the
 	///              current value of `input`.
 	public init<P: PropertyProtocol, T>(input: P, _ execute: @escaping (T) -> SignalProducer<Output, Error>) where P.Value == T {
 		self.init(input: input.map(Optional.some), execute)
--- a/Sources/Action.swift
+++ b/Sources/Action.swift
@@ -12,7 +12,7 @@ import enum Result.NoError
 public final class Action<Input, Output, Error: Swift.Error> {
 	private let deinitToken: Lifetime.Token
 
-	private let executeClosure: (Input) -> SignalProducer<Output, Error>
+	private let executeClosure: (Input) -> SignalProducer<Output, ActionError<Error>>
 	private let eventsObserver: Signal<Event<Output, Error>, NoError>.Observer
 	private let disabledErrorsObserver: Signal<(), NoError>.Observer
 
@@ -57,16 +57,20 @@ public final class Action<Input, Output, Error: Swift.Error> {
 	/// Initializes an action that will be conditionally enabled, and creates a
 	/// SignalProducer for each input.
 	///
+	/// - warning: This is an internal implementation detail of `Action.init(input:_:)`
+	///            and will be removed in a future version of ReactiveSwift.
+	///            Use `Action.init(enabledIf:_:)` instead.
+	///
 	/// - parameters:
 	///   - enabledIf: Boolean property that shows whether the action is
 	///                enabled.
 	///   - execute: A closure that returns the signal producer returned by
 	///              calling `apply(Input)` on the action.
-	public init<P: PropertyProtocol>(enabledIf property: P, _ execute: @escaping (Input) -> SignalProducer<Output, Error>) where P.Value == Bool {
+	public init<P: PropertyProtocol>(_internalEnabledIf property: P, tryExecute: @escaping (Input) -> SignalProducer<Output, ActionError<Error>>) where P.Value == Bool {
 		deinitToken = Lifetime.Token()
 		lifetime = Lifetime(deinitToken)
 
-		executeClosure = execute
+		executeClosure = tryExecute
 
 		(events, eventsObserver) = Signal<Event<Output, Error>, NoError>.pipe()
 		(disabledErrors, disabledErrorsObserver) = Signal<(), NoError>.pipe()
@@ -87,6 +91,20 @@ public final class Action<Input, Output, Error: Swift.Error> {
 
 		isEnabled = state.map { $0.isEnabled }
 		isExecuting = state.map { $0.isExecuting }
+	}
+
+	/// Initializes an action that will be conditionally enabled, and creates a
+	/// SignalProducer for each input.
+	///
+	/// - parameters:
+	///   - enabledIf: Boolean property that shows whether the action is
+	///                enabled.
+	///   - execute: A closure that returns the signal producer returned by
+	///              calling `apply(Input)` on the action.
+	public convenience init<P: PropertyProtocol>(enabledIf property: P, _ execute: @escaping (Input) -> SignalProducer<Output, Error>) where P.Value == Bool {
+		self.init(_internalEnabledIf: property) {
+			execute($0).mapError(ActionError.producerFailed)
+		}
 	}
 
 	/// Initializes an action that will be enabled by default, and creates a
@@ -136,8 +154,20 @@ public final class Action<Input, Output, Error: Swift.Error> {
 				disposable += signalDisposable
 
 				signal.observe { event in
-					observer.action(event.mapError(ActionError.producerFailed))
-					self.eventsObserver.send(value: event)
+					observer.action(event)
+
+					switch event {
+					case .failed(.disabled):
+						self.disabledErrorsObserver.send(value: ())
+					case let .failed(.producerFailed(error)):
+						self.eventsObserver.send(value: .failed(error))
+					case let .value(value):
+						self.eventsObserver.send(value: .value(value))
+					case .completed:
+						self.eventsObserver.send(value: .completed)
+					case .interrupted:
+						self.eventsObserver.send(value: .interrupted)
+					}
 				}
 			}
 
@@ -173,12 +203,16 @@ public protocol ActionProtocol: BindingTargetProtocol {
 	/// Initializes an action that will be conditionally enabled, and creates a
 	/// SignalProducer for each input.
 	///
+	/// - warning: This is an internal implementation detail of `Action.init(input:_:)`
+	///            and will be removed in a future version of ReactiveSwift.
+	///            Use `Action.init(enabledIf:_:)` instead.
+	///
 	/// - parameters:
 	///   - enabledIf: Boolean property that shows whether the action is
 	///                enabled.
 	///   - execute: A closure that returns the signal producer returned by
 	///              calling `apply(Input)` on the action.
-	init<P: PropertyProtocol>(enabledIf property: P, _ execute: @escaping (Input) -> SignalProducer<Output, Error>) where P.Value == Bool
+	init<P: PropertyProtocol>(_internalEnabledIf property: P, tryExecute: @escaping (Input) -> SignalProducer<Output, ActionError<Error>>) where P.Value == Bool
 
 	/// Whether the action is currently enabled.
 	var isEnabled: Property<Bool> { get }
@@ -224,8 +258,9 @@ extension ActionProtocol where Input == Void {
 	///   - execute: A closure to return a new SignalProducer based on the
 	///              current value of `input`.
 	public init<P: PropertyProtocol, T>(input: P, _ execute: @escaping (T) -> SignalProducer<Output, Error>) where P.Value == T? {
-		self.init(enabledIf: input.map { $0 != nil }) {
-			execute(input.value!)
+		self.init(_internalEnabledIf: input.map { $0 != nil }) {
+			guard let input = input.value else { return SignalProducer(error: .disabled) }
+			return execute(input).mapError(ActionError.producerFailed)
 		}
 	}
 
--- a/Sources/Action.swift
+++ b/Sources/Action.swift
@@ -88,7 +88,7 @@ public final class Action<Input, Output, Error: Swift.Error> {
 		errors = events.map { $0.error }.skipNil()
 		completed = events.filter { $0.isCompleted }.map { _ in }
 
-		let initial = ActionState(isExecuting: false, value: property.value, isEnabled: { isEnabled($0 as! State.Value) })
+		let initial = ActionState(value: property.value, isEnabled: { isEnabled($0 as! State.Value) })
 		state = MutableProperty(initial)
 
 		property.signal
@@ -179,12 +179,11 @@ public final class Action<Input, Output, Error: Swift.Error> {
 }
 
 private struct ActionState {
-	var isExecuting: Bool
+	var isExecuting: Bool = false
 	var value: Any
 	private let userEnabled: (Any) -> Bool
 
-	init(isExecuting: Bool, value: Any, isEnabled: @escaping (Any) -> Bool) {
-		self.isExecuting = isExecuting
+	init(value: Any, isEnabled: @escaping (Any) -> Bool) {
 		self.value = value
 		self.userEnabled = isEnabled
 	}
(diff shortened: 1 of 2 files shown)


Sources/Action.swift
```diff
@@ -174,6 +174,42 @@ extension Action: ActionProtocol {
...
+
+	public var values: Signal<Output, NoError> {
+		return action.values
+	}
+
+	public var errors: Signal<Error, NoError> {
+		return action.errors
+	}
+
+	public func apply(_ input: Void = ()) -> SignalProducer<Output, ActionError<Error>> {
```
sharplet: I initially tried to use `func apply() -> SignalProducer` here, but Swift complained about this not matching the signature of `apply` as defined in `ActionProtocol`. The workaround was to at least provide a default value of `()`, but this still feels like a bug to me.

mdiep:
> Ideally this could just be a convenience initialiser on Action

Did you try defining it in an extension on `ActionProtocol`?

sharplet:
I did think about defining it on `ActionProtocol`, but in order to define a convenience initialiser in a protocol extension, the protocol itself would have to define a designated (required?) initialiser. Moving the initialiser to the protocol didn't feel like the right tradeoff to me, but interested to hear your thoughts.

mdiep:
Yeah, I think adding an initializer to `ActionProtocol` would be fine. The point of the protocol is to make constrained extensions possible. I think that's preferable to adding another type.

sharplet:
Ok, so I tried moving it to an `ActionProtocol` extension, which _almost_ worked (but didn't): 1341fdba055dc9f60fbe79468b96d0831019d18e.

FWIW, I don't mind the idea of making `ActionProtocol` more abstract. The main use-case for me is the ability to constrain one or more of the associated types to be a concrete type. Three type parameters kind of makes `Action` unwieldy to use sometimes, and perhaps makes it unfriendly to beginners.

Also, just off the top of my head, you could conceivably introduce another implementation that enables concurrent execution of the inner signal producers, or one that cancels an executing producer when executed (i.e., the "merge" and "latest" flatten strategies).

Just some ideas!

andersio:
Try to make `ActionProtocol` class bound, and see if your convenience initializer works?

sharplet:
I did try that at one point, but it's not valid to define a `convenience init` in a protocol extension.

However ultimately the error isn't that you can't _define_ the error, it's Swift choosing the wrong initialiser at the call site. I think it's to do with the way `where` clauses affect type checking.

mdiep:
I'd like to play around with it to understand it better and see if I can find a way to make it work in an extension. But that may mean it takes me a little longer to get to this PR.

ikesyo:
To meet the requirement of `public init<P: PropertyProtocol, T>(input: P, _ execute: @escaping (T) -> SignalProducer<Output, Error>) where P.Value == T?`, you need `let input = MutableProperty<Int?>(0)` instead of `let input = MutableProperty(0)`.

sharplet:
@ikesyo Of course! I accidentally deleted the variant where `P.Value == T`.

Well, now that it works as expected, I've squashed those changes into the original commit. For now we can achieve this purely with `ActionProtocol`.

Sources/Action.swift
```diff
@@ -174,6 +184,36 @@ extension Action: ActionProtocol {
...
+	///            whenever the value is nil.
+	///		- execute: A closure to return a new SignalProducer based on the
+	///							 current value of `input`.
+	public init<P: PropertyProtocol, T>(input: P, _ execute: @escaping (T) -> SignalProducer<Output, Error>) where P.Value == T? {
+		self.init(enabledIf: input.map { $0 != nil }) {
+			execute(input.value!)
+		}
+	}
+
+	/// Initializes an action that uses an optional property for its input.
```
mdiep: This one isn't optional.

sharplet: Good catch! Also fixed up some weird tab vs. space problems Xcode inserted for me.

Sources/Action.swift
```diff
@@ -174,6 +184,36 @@ extension Action: ActionProtocol {
...
+	///
+	/// - parameters:
+	///   - input: An Optional property whose current value is used as input
+	///            whenever the action is executed. The action is disabled
+	///            whenever the value is nil.
+	///   - execute: A closure to return a new SignalProducer based on the
+	///              current value of `input`.
+	public init<P: PropertyProtocol, T>(input: P, _ execute: @escaping (T) -> SignalProducer<Output, Error>) where P.Value == T? {
+		self.init(enabledIf: input.map { $0 != nil }) {
+			execute(input.value!)
```
mdiep: Thinking about this more, I don't think this is safe.

The property's value could change between the enabled check and the time execute is called. If the value has switched back to `nil`, then this will assert.

I think this will require some changes to `apply` and maybe a different initializer. I don't see how we can make the current approach work.

sharplet: What do you think about https://github.com/ReactiveCocoa/ReactiveSwift/pull/22/commits/07ea7e7c18c04c19275e3ddcb3dc1e2bde9bc47e? It synchronises checking for `isEnabled` and calling `executeClosure`, so I think should solve this particular problem.

It feels like there could be a more fundamental refactoring of Action that would remove the need for force-unwrapping (like you were hinting at), but this is the smallest change I could think of.

sharplet: I was actually having trouble reasoning about whether changes to `isExecuting` and `userEnabled` were truly atomic, so I resolved that by unifying the two properties in a single struct that can be modified atomically via `MutableProperty`. See 25a3024.

Sources/Action.swift
```diff
@@ -31,34 +31,12 @@ public final class Action<Input, Output, Error: Swift.Error> {
...
-		label: "org.reactivecocoa.ReactiveSwift.Action.executingQueue",
-		attributes: []
-	)
-
-	/// Whether the action should be enabled for the given combination of user
-	/// enabledness and executing status.
-	private static func shouldBeEnabled(userEnabled: Bool, executing: Bool) -> Bool {
-		return userEnabled && !executing
-	}
+	private let _state: MutableProperty<ActionState>
```
mdiep: Our style has been to not prefix private variables with an underscore unless there's a public variable with the same name.

Sources/Action.swift
```diff
@@ -109,21 +95,21 @@ public final class Action<Input, Output, Error: Swift.Error> {
...
+			var executedProducer: SignalProducer<Output, Error>?
 
-			self.executingQueue.sync {
-				if self._isEnabled.value {
-					self._isExecuting.value = true
-					startedExecuting = true
+			self._state.modify { state in
+				if state.isEnabled {
+					state.isExecuting = true
+					executedProducer = self.executeClosure(input)
```
mdiep: You can `return` this from the `modify` directly instead of assigning to a `var`.

Sources/Action.swift
mdiep: I'm still working through whether it's safe (trying to verify that it won't deadlock), but it's definitely easier to understand the code here now. 👍🏻

Sources/Action.swift
mdiep: I don't feel good about calling the `executeClosure` inside a lock. I think we need to find a different solution. 😕

andersio: `let hasStartedExecution: Bool = self.state.modify { ... }` should do.

Sources/Action.swift
```diff
@@ -148,6 +147,16 @@ public protocol ActionProtocol {
...
 
+	/// Initializes an action that will be conditionally enabled, and creates a
+	/// SignalProducer for each input.
+	///
+	/// - parameters:
+	///   - enabledIf: Boolean property that shows whether the action is
+	///                enabled.
+	///   - execute: A closure that returns the signal producer returned by
+	///              calling `apply(Input)` on the action.
+	init<P: PropertyProtocol>(enabledIf property: P, _ execute: @escaping (Input) -> SignalProducer<Output, Error>) where P.Value == Bool
```
mdiep: If we changed this initializer to this:

```swift
init<P: PropertyProtocol>(enabledIf property: P, _ execute: @escaping (Input) -> SignalProducer<Output, ActionError<Error>>) where P.Value == Bool
```

Then the execute closure could bail if required. This seems like it'd be a nice safety addition in general—for cases when you would need make guarantees.

🤔

Sources/Action.swift
mdiep: The reason we started on this is so that this:

```swift
	public init<P: PropertyProtocol, T>(input: P, _ execute: @escaping (T) -> SignalProducer<Output, Error>) where P.Value == T? {
		self.init(enabledIf: input.map { $0 != nil }) {
			execute(input.value!)
		}
	}
```

Could guarantee that `input.value` was non-nil. I actually don't think the code as written solves even this.

Sources/Action.swift
sharplet: I tried making this change (having the execute closure return `SignalProducer<Output, ActionError<Error>>`), but I couldn't figure out a nice way to make the API work without breaking the existing use case.

However, I tried making the return value an optional signal producer, which seems to work nicely: https://github.com/ReactiveCocoa/ReactiveSwift/commit/c8791b7a75123e4885262b7da5cdbc5f7bb8668f. 

There's something still wrong to me about this though. I think if we want to truly prove this is safe, we need to make checking the enabled state and grabbing the input value an atomic operation, which means moving the input property into `ActionState` somehow...

sharplet: I really like the idea of making `ActionState` more powerful by allowing it to carry an optional value, but I can't think of a way to achieve that without leaking implementation details out via `ActionProtocol`.

That is, I'd like to make a private initialiser, but because `ActionProtocol.init(input:_:)` is in a protocol extension it has to call a public initialiser. Starting to wonder whether the only way to make this safe & atomic is to reintroduce `PropertyAction`, so it can redefine `isEnabled` in terms of `input` while still providing atomic access to the current value.

mdiep: I was think that this:

```swift
private let executeClosure: (Input) -> SignalProducer<Output, ActionError<Error>>

// new init that uses ActionError
public init<P: PropertyProtocol>(enabledIf property: P, _ execute: @escaping (Input) -> SignalProducer<Output, ActionError<Error>>) where P.Value == Bool { … }

// existing init
public init<P: PropertyProtocol>(enabledIf property: P, _ execute: @escaping (Input) -> SignalProducer<Output, Error>) where P.Value == Bool {
    self.init(enabledIf: property) { return execute.promoteErrors(ActionError.ProducerError) }
}
```

would work.

sharplet: The problem is that the only difference between the two initialisers is the return type of `execute`: this makes trailing closures with inferred return types ambiguous, and I don't think forcing users to always explicitly annotate the return type is a good idea.

sharplet:
@mdiep I managed to work out a way to introduce an unambiguous overload, so that we can safely unwrap the optional. However I couldn't work out a way to do so without leaking implementation details, so I'd appreciate your input on the solution I came up with.

**IMO**, given the current limitations of extensions of generic types and the need to use protocol extensions, I think it would ultimately be a _cleaner API **today**_ to implement this as `PropertyAction`. It may have to duplicate some logic in `Action`, but we probably wouldn't have to leak implementation details.

Having said that, I think this version of the API would work out just fine in practice, so I ultimately don't mind which way we go. And I'm prepared to admit there might be an even cleaner way to refactor this without leaking implementation details!

andersio:
Concrete same-type requirement would land in 3.1 though. If these can be converted without breaking the source, IMO it is okay to tolerate an exposed private API for a while.

mdiep:
This is still on my list!

Sorry for the delay. I'll try to get to it this week.

sharplet:
I think I've managed to make the input property thread-safe with https://github.com/ReactiveCocoa/ReactiveSwift/pull/22/commits/fea3dc61596a6f3c5a6e38eddfc4884b59e30608!

Related discussion: https://github.com/ReactiveCocoa/ReactiveSwift/issues/88#issuecomment-261449006

andersio:
Sounds great. The underscore label of the designated initializer can be stripped now, no?

andersio:
I have made a patch to streamline the implementation, which looks thread safe to me.
https://github.com/ReactiveCocoa/ReactiveSwift/commit/4b5ac118b3d9c31bd516c640e4aa392452316e67

Sources/Action.swift
```diff
@@ -170,6 +204,28 @@ public protocol ActionProtocol: BindingTargetProtocol {
...
+	///
+	/// - parameters:
+	///   - state: A property that provides the current state of the action
+	///            whenever `apply()` is called.
+	///   - enabledIf: A predicate that, given the current value of `state`,
+	///                returns whether the action should be enabled.
+	///   - execute: A closure that returns the signal producer returned by
+	///              calling `apply(Input)` on the action, optionally using
+	///              the current value of `state`.
+	init<State: PropertyProtocol>(_internalState property: State, enabledIf isEnabled: @escaping (State.Value) -> Bool, _ execute: @escaping (State.Value, Input) -> SignalProducer<Output, Error>)
```
andersio: Sounds great. The underscore label of the designated initializer can be stripped now, can't it? It has a valid use case now as a public API (state with any arbitrary isEnabled predicate).

Sources/Action.swift
```diff
@@ -155,12 +169,32 @@ public final class Action<Input, Output, Error: Swift.Error> {
...
 	}
 }
 
+private struct ActionState {
+	var isExecuting: Bool
+	var value: Any
+	private let userEnabled: (Any) -> Bool
+
+	init(isExecuting: Bool, value: Any, isEnabled: @escaping (Any) -> Bool) {
+		self.isExecuting = isExecuting
```
mdiep: This could probably default to `false` instead of getting passed in.

Sources/Action.swift
```diff
@@ -202,6 +258,36 @@ extension Action: ActionProtocol {
...
+extension ActionProtocol where Input == Void {
+	/// Initializes an action that uses an optional property for its input,
+	/// and is disabled whenever the input is nil. When executed, a SignalProducer
+	/// is created with the current value of the input.
+	///
+	/// - parameters:
+	///   - input: An Optional property whose current value is used as input
+	///            whenever the action is executed. The action is disabled
+	///            whenever the value is nil.
+	///   - execute: A closure to return a new SignalProducer based on the
```
mdiep: `nil`, `Optional`, and `SignalProducer` should all be backtick-quoted here.

Sources/Action.swift
```diff
@@ -155,12 +169,32 @@ public final class Action<Input, Output, Error: Swift.Error> {
...
+	init(isExecuting: Bool, value: Any, isEnabled: @escaping (Any) -> Bool) {
+		self.isExecuting = isExecuting
+		self.value = value
+		self.userEnabled = isEnabled
+	}
+
+	/// Whether the action should be enabled for the given combination of user
+	/// enabledness and executing status.
+	fileprivate var isEnabled: Bool {
+		return userEnabled(value) && !isExecuting
```
mdiep: We should cache the value of `userEnabled(value)` and only call it when the value changes.

Sources/Action.swift
```diff
@@ -94,12 +87,33 @@ public final class Action<Input, Output, Error: Swift.Error> {
...
+					$0.value = newValue
+				}
+			}
+
+		self.isEnabled = state.map { $0.isEnabled }
+		self.isExecuting = state.map { $0.isExecuting }
+	}
+
+	/// Initializes an action that will be conditionally enabled, and creates a
+	/// SignalProducer for each input.
```
mdiep: `SignalProducer` should be backtick-quoted.

sharplet:
@andersio In order not to block this PR any further, how do you feel about submitting your proposed changes as a separate PR after this is merged?

sharplet:
@mdiep I've addressed your feedback, thanks! Hopefully good to go once we're green.

mdiep:
This looks great. Thanks for all the hard work on this, @sharplet!

andersio:
@andersio Sure.

sharplet:
📗 🍏 💚 :shipit:

ikesyo:
Looks like the `state property: State` parameter of the new initializer must be kept alive outside an initialized action: https://github.com/ReactiveCocoa/ReactiveObjCBridge/pull/10#issuecomment-264860959. Is that an intentional behavior change? Formerly the given property is captured by the action by `private let isUserEnabled: Property<Bool>`.

mdiep:
I don't think that was intentional.

andersio:
It would need to be retained even without this PR due to #117, unless we use `Property(capturing:)`.

sharplet:
Confirmed, that was unintentional. I experimented with a fix for that but shelved it until #117 was closed.

PR #23, opened 2016-09-20
[ScopedDisposable] Fix infinite loop in `init(_ disposable: Disposable)`

### LINKED ISSUE #17: Getting infinite loop on ScopedDisposable.init
Hi, i'm getting an infinite loop when doing the following:

`let disposable = someProducer.startWithNext ...`
`let scoped = ScopedDisposable(disposable)`

It loops on ScopedDisposable's extensions' `init(_ disposable: Disposable)`, which is here: https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/Disposable.swift#L230

This init has the same signature as ScopedDisposable's non-extension init (i'm not sure how this compiles without error), and when the extension calls `self.init` it calls the extension init instead of the non-extension init, which causes infinite recursion.

Recommend changing the signature of the extension init to have a named argument, or even remove it? I'm not sure what the need for the extension init is, given that the normal init is fully generic already.

For reference for anyone else having this issue, you can workaround for now with: `let scoped  = ScopedDisposable(AnyDisposable(disposable))`

Thanks

ikesyo:
Fixes #17.

### DIFF
--- a/Sources/Disposable.swift
+++ b/Sources/Disposable.swift
@@ -227,7 +227,7 @@ extension ScopedDisposable where InnerDisposable: AnyDisposable {
 	///   - disposable: A disposable to dispose of when deinitializing, which
 	///                 will be wrapped in an `AnyDisposable`.
 	public convenience init(_ disposable: Disposable) {
-		self.init(AnyDisposable(disposable))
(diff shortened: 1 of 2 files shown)


mdiep:
👍🏻

PR #27, opened 2016-09-21
[WIP] Merging from Rex.

andersio:
Still WIP.


NachoSoto:
Hmm is this to bring `Rex` to `ReactiveSwift`? If so I'm a big :-1::-1::-1:

andersio:
@mdiep seems leaning to preserve the history when I last asked, so I worked it out. The option to squash and merge is still here, anyway.

NachoSoto:
It's not about the history, it's about bringing (part of?) `Rex`. What's the reasoning behind this?
What logic did you follow to decide what to include in this PR?

andersio:
The separation is solely based on the dependencies or its nature (Cocoa or ObjC, etc), as I thought it was about merging in everything from Rex. Perhaps I should focus on bringing the bindings and Cocoa extensions instead, and leave the operators alone.

mdiep:
I think the confusion is that Rex is meant to go into ReactiveCocoa, not ReactiveSwift.

andersio:
There are quite a few bits in Rex that is not Cocoa specific, and is extending the core ReactiveSwift.

ikesyo:
Those are __extensions__ and should not be the core ReactiveSwift API I think.

neilpa:
Yea, the idea I was to extract the Cocoa-specific bindings and use that as a new baseline for a Swift-centric Reactive Cocoa. The rest of Rex was mostly experimental operators that were ports from other Objective-C projects of mine. Those are probably too specific to be included in ReactiveSwift.

PR #28, opened 2016-09-21
Add the @discardableResult attribute to a Disposable.add overload

sibljon:
As implied by the documentation comment (and correctly so, IMO), it seems like the return value of this `Disposable.add` function overload should be discardable.


NachoSoto:
Makes sense

mdiep:
Thanks! These were really easy to miss in the Swift 3 conversion.

PR #30, opened 2016-09-22
Introduce `BindingTarget` (class).

andersio:
For representing write-only binding targets. Renamed `BindingTarget` the protocol to `BindingTargetProtocol`.

As suggested by @mdiep, a special variant `BindingTarget(setter:lifetime:on:)` is also included. It always synchronously triggers the setter action on the specified queue, or elides the dispatch if it is already on the same queue.

It should not cause any issues with the async operators, but only interoperating with explicit but perhaps strange uses of dispatch queues. e.g.

```
mainQueue.sync { anotherQueue.sync { mainQueueTarget <~ myProperty } }
```

### DIFF
--- a/Sources/UnidirectionalBinding.swift
+++ b/Sources/UnidirectionalBinding.swift
@@ -1,3 +1,4 @@
+import Dispatch
 import enum Result.NoError
 
 precedencegroup BindingPrecedence {
@@ -276,3 +277,63 @@ extension BindingTargetProtocol where Value: OptionalProtocol {
 		return target <~ property.producer
 	}
 }
+
+/// A binding target that can be used with the `<~` operator.
+public final class BindingTarget<Value>: BindingTargetProtocol {
+	public let lifetime: Lifetime
+	private let setter: (Value) -> Void
+
+	/// Wraps a `BindingTarget`.
+	///
+	/// - parameters:
+	///   - target: The binding target to wrap.
+	public init<T: BindingTargetProtocol>(_ target: T) where T.Value == Value {
+		lifetime = target.lifetime
+		setter = target.consume
+	}
+
+	/// Creates a binding target.
+	///
+	/// - parameters:
+	///   - setter: The action to receive values.
+	///   - lifetime: The expected lifetime of any bindings against the resulting
+	///               target.
+	public init(setter: @escaping (Value) -> Void, lifetime: Lifetime) {
+		self.setter = setter
+		self.lifetime = lifetime
+	}
+
+	/// Creates a binding target which consumes values synchronously on the
+	/// supplied queue.
+	///
+	/// - parameters:
+	///   - setter: The action to receive values.
+	///   - queue: The dispatch queue on which the values are consumed.
+	///   - lifetime: The expected lifetime of any bindings against the resulting
+	///               target.
+	public convenience init(setter: @escaping (Value) -> Void, lifetime: Lifetime, on queue: DispatchQueue) {
+		let queueId = ObjectIdentifier(queue)
+
+		/// Ensures the queue has been setup property.
+		if nil == queue.getSpecific(key: specificKey) {
+			queue.setSpecific(key: specificKey, value: queueId)
+		}
+
+		let setter: (Value) -> Void = { value in
+			if queueId == DispatchQueue.getSpecific(key: specificKey) {
+				setter(value)
+			} else {
+				queue.sync {
+					setter(value)
+				}
+			}
+		}
+		self.init(setter: setter, lifetime: lifetime)
+	}
+
+	public func consume(_ value: Value) {
+		setter(value)
+	}
+}
+
+private let specificKey = DispatchSpecificKey<ObjectIdentifier>()
--- a/Sources/UnidirectionalBinding.swift
+++ b/Sources/UnidirectionalBinding.swift
@@ -325,6 +325,15 @@ public final class BindingTarget<Value>: BindingTargetProtocol {
 	public func consume(_ value: Value) {
 		setter(value)
 	}
+
+	@discardableResult
+	public static func <~ <Source: SignalProtocol>(target: BindingTarget<Value>, signal: Source) -> Disposable? where Source.Value == Value, Source.Error == NoError {
+		return signal
+			.take(during: target.lifetime)
+			.observeValues { [setter = target.setter] value in
+				setter(value)
+			}
+	}
 }
 
 private let specificKey = DispatchSpecificKey<ObjectIdentifier>()
--- a/Sources/UnidirectionalBinding.swift
+++ b/Sources/UnidirectionalBinding.swift
@@ -10,7 +10,7 @@ precedencegroup BindingPrecedence {
 infix operator <~ : BindingPrecedence
 
 /// Describes a target to which can be bound.
-public protocol BindingTarget: class {
+public protocol BindingTargetProtocol: class {
 	associatedtype Value
 
 	/// The lifetime of `self`. The binding operators use this to determine when
@@ -53,7 +53,7 @@ public protocol BindingTarget: class {
 	static func <~ <Source: SignalProtocol>(target: Self, signal: Source) -> Disposable? where Source.Value == Value, Source.Error == NoError
 }
 
-extension BindingTarget {
+extension BindingTargetProtocol {
 	/// Binds a signal to a target, updating the target's value to the latest
 	/// value sent by the signal.
 	///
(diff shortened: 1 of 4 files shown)


Sources/UnidirectionalBinding.swift
```diff
@@ -276,3 +276,62 @@ extension BindingTarget where Value: OptionalProtocol {
...
+	}
+
+	public func consume(_ value: Value) {
+		setter(value)
+	}
+}
+
+private let specificKey = DispatchSpecificKey<Void>()
+
+private var bindingTargetToken: Void = {
```
NachoSoto: This can be a constant.

andersio: Done.

Sources/UnidirectionalBinding.swift
```diff
@@ -276,3 +277,62 @@ extension BindingTarget where Value: OptionalProtocol {
...
+	public let lifetime: Lifetime
+	private let setter: (Value) -> Void
+
+	/// Wraps a `BindingTarget`.
+	///
+	/// - parameters:
+	///   - target: The binding target to wrap.
+	public init<T: BindingTarget>(_ target: T) where T.Value == Value {
+		lifetime = target.lifetime
+		setter = target.consume
```
sharplet: I think this is going to retain `target`.

Sources/UnidirectionalBinding.swift
```diff
@@ -276,3 +277,62 @@ extension BindingTarget where Value: OptionalProtocol {
 		return target <~ property.producer
 	}
 }
+
+/// A pseudo `BindingTarget`, constructed from a closure and a lifetime.
```
sharplet: What do you think about dropping the word "pseudo"? To me it makes it sound like this isn't a "real" binding target, even though it's as real as any other binding target (and arguably purer).

Sources/UnidirectionalBinding.swift
```diff
@@ -276,3 +277,62 @@ extension BindingTarget where Value: OptionalProtocol {
...
+public final class AnyBindingTarget<Value>: BindingTarget {
```
sharplet: Have we considered any other names than `AnyBindingTarget`? I can see from the initialiser below that this can be used as a type-erased wrapper, but I'm not sure that's its primary purpose. Given that this is perfectly useful as just a pure sink for a binding, what about calling it `BindingSink` or `Sink` or even just `BindingTarget` (the latter implying a rename of the protocol to `BindingTargetProtocol`)?

I'm not necessarily opposed to the name, just thought I'd throw some alternatives out there :)

Sources/UnidirectionalBinding.swift
```diff
@@ -276,3 +277,62 @@ extension BindingTarget where Value: OptionalProtocol {
...
+	}
+
+	/// Creates a pseudo `BindingTarget` which consumes values synchronously on
+	/// the main queue.
+	///
+	/// - parameters:
+	///   - setter: The action to receive values.
+	///   - lifetime: The expected lifetime of any bindings against the resulting
+	///               target.
+	public convenience init(mainQueueSetter: @escaping (Value) -> Void, lifetime: Lifetime) {
```
sharplet: Any reason not to allow any old queue to be passed here instead of tying this to the main queue?

Sources/UnidirectionalBinding.swift
andersio: An existential wrapper retains what it is wrapping, isn't it?

Sources/UnidirectionalBinding.swift
andersio: Removed.

Sources/UnidirectionalBinding.swift
andersio: Renamed.

Sources/UnidirectionalBinding.swift
andersio: Changed this to `init(setter:lifetime:on:)`.

Sources/UnidirectionalBinding.swift
sharplet: Hmm, I suppose? But given that we're referencing `target.lifetime`, I was assuming we wouldn't want to extend the lifetime by also retaining `target`. I don't know what the correct semantic is.

sharplet: However, if I have a text field, and a binding target that sets its `text` property, retaining the binding target is not the same as retaining the text field. So I think you're right, the important thing is that the binding target doesn't affect the lifetime of the subject of that binding.

andersio: Hmm... in its current state, it is actually a problem after a second thought for cases like `DynamicProperty`.

andersio: I have removed it, since it isn't needed now for the Rex merger.

Anyway, implementing this correctly would probably require type erasing via inheritance and runtime type check, especially for the `<~` requirement.

mdiep:
What's the problem this is solving?

I see you mentioned this from https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3210, but I don't see `BindingTarget` referenced anywhere in that PR.

sharplet:
Some context is here: https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3210#pullrequestreview-1197439

Many bindable properties provided by Rex are write-only, despite using `MutableProperty` which should be read-write. My understanding is that the goal of this PR is to provide a first-class way to provide a write-only bindings. IMO that's a useful feature on its own, although in the case of the Rex bindings it might ultimately be more useful (if more complex) to provide proper `MutableProperties` with the expected semantics.

andersio:
@mdiep It hasn't been adopted in that branch yet. Like how @sharplet said, this is primarily for binding targets wrapping properties of UIKit components which was said not to be KVO compliant. In other words, we can't make them emitting changes, and this means it should not be a `Property` at all.

andersio:
@sharplet AFAIU we need only two classes of properties for UIKit bindings:

1. write-only for non-observable properties or method calls, e.g. `UILabel.text`, `UIViewController.dismissAnimated`, etc; and
2. read-write, e.g. `UITextField.text`, `UISwitch.on`, etc.

The rest would be `Signal`s. `BindingTarget` in this PR specifically targets the case 1, while case 2 would require a similar construct conforming `MutablePropertyProtocol`.

mdiep:
It seems like cases like `UIViewController.dismissAnimated` would be better served by an `Action`—since that already abstracts the idea of something that can be called to perform some task. If `Action` were made to conform to `BindingTarget`, that seems like it'd be a pretty natural extension of what we already have.

andersio:
Yeah, `Action` can for sure be extended as a binding target, although how the errors and the outputs should be handled is debatable.

I am just not sure if the serialisation/abstraction overhead of `Action` should be carried over in these binding points - they are already implicitly serialised due to the main queue requirement.

andersio:
Yeah, but the problem is that a straightforward `<~` towards an action could still potentially ignore `ActionError.disabled`, although in practice for bindings on the main queue it would never happen.

mdiep:
> We also have no conditional conformances which allows only Action where Error == NoError to be a binding target.

I think we'd only need to verify that the input signal/producer/property has `NoError`—it doesn't matter if the `Action` can error. I'm also not convinced that it matters that inputs would be ignored while the action is disabled.

This is perhaps most true, from RAC's perspective, if you could configure an `Action` to execute on the main queue.

It may not work out, but it seems like this approach is worth exploring? Perhaps in a separate PR?

andersio:
@mdiep It matters if the action is asynchronous. But for the case of `dismissAnimated` specifically, it does not matter due to the main queue requirement.

liscio:
Yikes. `dismissAnimated` was a `MutableProperty`?  That's smellier than the Objective-C 2.0 anti-pattern of calling methods that don't take parameters using dot notation…

@andersio I'm not sure I fully understand the need for *any* write-only properties, to be honest. Do you have any examples where you'd need to "bind" something to a `dismissAnimated` call?

Also, your example of a `UILabel` doesn't quite work as a write-only binding. It would still have to be a `MutableProperty` because there is an underlying value in the control that has a "current value," and does still change (albeit for different reasons.)

andersio:
Well, `dismissAnimated` is just a case inherited from the codebase. As I might have reiterated for times, the intention of a "write-only" construct, i.e. `BindingTarget` in this PR, is to model things ike `UILabel.text`. I am not sure why it doesn't make sense to you, but `UILabel` is just a "read-only" component that displays the string value it holds. The same applies to many properties like `UIView.isHidden`, `UIView.alpha`, `UILabel.font`, etc.

andersio:
`BindingTarget` in action:
https://github.com/ReactiveCocoa/ReactiveCocoa/compare/rex...rex-landing?expand=1

liscio:
> BindingTarget in action:

Seeing it in use is very helpful, thanks! I much prefer this model to the `MutableProperty` values that get bound at runtime using objc_setAssociatedObject.

PR #31, opened 2016-09-23
32bit arch crash: timer producer fails when interval is >= 3 seconds

taquitos:
Nanos conversion overflows on 32 bit devices because new dispatch API takes an Int and not Int64.
When interval and leeway get high enough and then we cast to Int, it overflows.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,10 @@ matrix:
         - XCODE_SDK=iphonesimulator
         - XCODE_ACTION="build-for-testing test-without-building"
         - XCODE_DESTINATION="platform=iOS Simulator,name=iPhone 6s"
+      env:
+        - XCODE_SDK=iphonesimulator
+        - XCODE_ACTION="build-for-testing test-without-building"
+        - XCODE_DESTINATION="platform=iOS Simulator,name=iPhone 5"
     - xcode_scheme: ReactiveSwift-tvOS
       env:
         - XCODE_SDK=appletvsimulator
--- a/Sources/Scheduler.swift
+++ b/Sources/Scheduler.swift
@@ -264,16 +264,16 @@ public final class QueueScheduler: DateSchedulerProtocol {
 		precondition(interval >= 0)
 		precondition(leeway >= 0)
 
-		let nsecInterval = interval * Double(NSEC_PER_SEC)
-		let nsecLeeway = leeway * Double(NSEC_PER_SEC)
+		let msecInterval = interval * 1000
+		let msecLeeway = leeway * 1000
 
 		let timer = DispatchSource.makeTimerSource(
 			flags: DispatchSource.TimerFlags(rawValue: UInt(0)),
 			queue: queue
 		)
 		timer.scheduleRepeating(wallDeadline: wallTime(with: date),
-		                        interval: .nanoseconds(Int(nsecInterval)),
-		                        leeway: .nanoseconds(Int(nsecLeeway)))
+		                        interval: .milliseconds(Int(msecInterval)),
+		                        leeway: .milliseconds(Int(msecLeeway)))
 		timer.setEventHandler(handler: action)
 		timer.resume()
 
--- a/Tests/ReactiveSwiftTests/SignalProducerSpec.swift
+++ b/Tests/ReactiveSwiftTests/SignalProducerSpec.swift
@@ -794,7 +794,13 @@ class SignalProducerSpec: QuickSpec {
 			}
 
 			it("shouldn't overflow on a real scheduler") {
-				let scheduler = QueueScheduler()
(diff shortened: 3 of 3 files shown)


Sources/Scheduler.swift
```diff
@@ -264,16 +264,16 @@ public final class QueueScheduler: DateSchedulerProtocol {
...
+		let msecLeeway = leeway * 1000
 
 		let timer = DispatchSource.makeTimerSource(
 			flags: DispatchSource.TimerFlags(rawValue: UInt(0)),
 			queue: queue
 		)
 		timer.scheduleRepeating(wallDeadline: wallTime(with: date),
-		                        interval: .nanoseconds(Int(nsecInterval)),
-		                        leeway: .nanoseconds(Int(nsecLeeway)))
+		                        interval: .milliseconds(Int(msecInterval)),
```
NachoSoto: Wait, why not just use `.seconds(Int(interval))` now? This conversion is legacy from pre-Swift 3, but shouldn't be needed anymore.

Tests/ReactiveSwiftTests/SignalProducerSpec.swift
```diff
@@ -793,6 +793,14 @@ class SignalProducerSpec: QuickSpec {
 				expect(dates) == [tick1, tick2, tick3]
 			}
 
+			it("shouldn't overflow on a real scheduler") {
```
NachoSoto: Thanks for adding a test!

Tests/ReactiveSwiftTests/SignalProducerSpec.swift
```diff
@@ -793,6 +793,14 @@ class SignalProducerSpec: QuickSpec {
...
+				let scheduler = QueueScheduler()
+				let producer = timer(interval: 3, on: scheduler)
+				let disposable = producer.start()
```
NachoSoto: Maybe just
```swift
producer
  .start()
  .dispose()
```

Tests/ReactiveSwiftTests/SignalProducerSpec.swift
```diff
@@ -793,6 +793,14 @@ class SignalProducerSpec: QuickSpec {
 				expect(dates) == [tick1, tick2, tick3]
 			}
 
+			it("shouldn't overflow on a real scheduler") {
+				let scheduler = QueueScheduler()
+				let producer = timer(interval: 3, on: scheduler)
```
NachoSoto: If the crash happened with `> 9` seconds, why not use `10` here to make sure this catches a potential regression? :)

.travis.yml
```diff
@@ -22,6 +22,10 @@ matrix:
         - XCODE_SDK=iphonesimulator
         - XCODE_ACTION="build-for-testing test-without-building"
         - XCODE_DESTINATION="platform=iOS Simulator,name=iPhone 6s"
+      env:
```
NachoSoto: Cool

Sources/Scheduler.swift
taquitos: I think that would prevent people from being able to specify like 0.250 seconds

Tests/ReactiveSwiftTests/SignalProducerSpec.swift
taquitos: True dat, I thought it was > 9, but it's actually > 2 that causes issues. Will update title.

Tests/ReactiveSwiftTests/SignalProducerSpec.swift
taquitos: yeah, great point.

Sources/Scheduler.swift
NachoSoto: Oh I forgot the parameter is a `TimeInterval`. Okay, then I think the right fix, which is not backwards compatible, but definitely best long term, is to change the parameter to take a `DispatchTimeInterval`. It's actually a much better API because it self-documents the magnitude. With these old APIs taking `Double`s it's never clear whether it's seconds or what.

NachoSoto: We can keep the original method and deprecate it, and point users to the new one.

taquitos: I love the idea, I started working through what that would take, and we have a few options. If we deprecate this function, there are A LOT of places that will need further deprecations and significant changes as we roll out the newer API including ```DispatchTimeInterval``` instead of ```TimeInterval```
How do you feel about keeping this change to milliseconds, deprecating this function, and then creating an issue to adjust all the other functions that rely on this now deprecated API?

NachoSoto: Sounds perfect 👍 let's not deprecate it in this PR then.

taquitos:
Yey!

PR #32, opened 2016-09-23
[Proposal][Action] Add `disabledErrors` signal

ikesyo:
~~One more breaking change for ReactiveSwift 1.0.0.~~

Revisits https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2882.


sharplet:
My only concern with this is that it makes `Action.errors` harder to use in the usual case in order to make this new case possible.

What if we made the change non-breaking by introducing a separate `Signal<(), NoError>` that sends an event whenever the action is executed while disabled? That would achieve the same use case described in https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2882 without adding complexity.

mdiep:
> My only concern with this is that it makes Action.errors harder to use in the usual case in order to make this new case possible.

I agree. I think we should expose this somehow, but I don't think we should make the ordinary case harder.

ikesyo:
Thanks for your thoughts. I added `let disabledErrors: Signal<(), NoError>` instead.

Sources/Action.swift
```diff
@@ -11,25 +11,29 @@ import enum Result.NoError
...
 	/// In other words, this will send every value from every signal generated
-	/// by each SignalProducer returned from apply().
+	/// by each SignalProducer returned from apply() except `ActionError.disabled`.
 	public let values: Signal<Output, NoError>
 
 	/// A signal of all errors generated from applications of the Action.
 	///
 	/// In other words, this will send errors from every signal generated by
-	/// each SignalProducer returned from apply().
+	/// each SignalProducer returned from apply() except `ActionError.disabled`.
```
mdiep: I think this should be a separate `-note: ` section. It doesn't actually makes sense here because `ActionError.disabled` never comes from the `SignalProducer` returned from `apply()`.

ikesyo: `apply()` returns `SignalProducer<Output, ActionError<Error>>` and the producer sends `ActionError.disabled` if the action is still exectuing. Am I misunderstanding?

ikesyo: You may mean the `execute` closure of `init`.

mdiep: Oh, you're right. I was thinking it was the signal returned from the execute closure. 🙈

PR #38, opened 2016-09-24
`Action` as a binding target.

andersio:
https://github.com/ReactiveCocoa/ReactiveSwift/pull/30#issuecomment-249382191

This PR also includes a new property operator `PropertyProtocol.take(until:)`.


andersio:
A new initializer `Action(lifetime:enabledIf:_:)` is added, so that `Action` does not need to be intentionally retained (and uniqued) for bindings to work properly if it eventually depends on some other's lifetime.

andersio:
It is possible to make `Action` be able to wrap an arbitrary `Lifetime`. Say `rac_dismissAnimated: Action<...>` can wrap the view controller's lifetime, so that bindings towards the action can still be teared down even if somehow the action is being retained.

mdiep:
This seems pretty interesting to me.

Are you still planning to work on tests for this?

andersio:
About the ability to "pass through" an arbitrary lifetime, there are two possible design choices:

1. Allow `Action` to be a throwaway. (CURRENT)
    Bindings are compile-time safe in the sense that even the API consumer unintentionally forgets to retain an `Action`, the bindings would still work.

2. Bindings are valid only if `Action` is **explicitly** retained as long as the lifetime.
    Bindings are **NOT** compile-time safe, since the `Action` is allowed to deinitialize ahead of the end of lifetime due to various reasons. But it (wishfully) forces users to consider the need of uniquing.

mdiep:
I think we should stick with the view that `Action`s must be explicitly retained. I don't think they really make sense otherwise.

If you have an `Action` property on an object, you'll always want to return the same object so that it serializes correctly.

andersio:
Yeah, either way they should be retained by convention. It is just a matter of static-time correctness. 

If `Action` wraps a lifetime that outlives it, I would expect the bindings towards it to be aligned with the wrapped lifetime, despite the fact that generally `Action` would be explicitly retained by the owner of the wrapped lifetime.

mdiep:
I just don't see standalone `Action`s being useful.

Yes, it protects against user error. But I don't think that's worth the added complexity. In the `magnitude * reach` test, I think it scores low areas. It's unlikely to have a large effect and unlikely to affect many people.

At the very least, I think we should leave things as they are now. We can reconsider for ReactiveSwift 2. But I'd like to focus for now on shipping the 1.0 release.

andersio:
I guess we can leave `dismissAnimated` as is - it wouldn't be a source breaking change even if we decide to swap in `Action` later.

mdiep:
I think we could still make `Action` conform to `BindingTargetProtocol` and use it in RAC for `dismissAnimated`.

mdiep:
Awesome. Thanks!

PR #39, opened 2016-09-25
Allow targeting work on a specific queue

liscio:
When modernizing the QueueScheduler API, I inadvertently left out the rare-but-useful case where work could be targeted to a specific destination queue.

This was driven by some late-breaking discussion in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2295.

This change is rather minimal in that it pushes the target queue into an optional parameter with a default nil value, so callers should not be forced to change.

I also intended for the API to be clear about the fact that we are targeting an existing queue, so callers (hopefully) won't end up "falling back" into the old pattern of targeting a global queue of their preferred priority.

### DIFF
--- a/Sources/Scheduler.swift
+++ b/Sources/Scheduler.swift
@@ -157,8 +157,8 @@ public final class QueueScheduler: DateSchedulerProtocol {
 	///         the `QueueScheduler` will be serial with respect to each other.
 	///
 	/// - warning: Obsoleted in OS X 10.11
-	@available(OSX, deprecated:10.10, obsoleted:10.11, message:"Use init(qos:, name:) instead")
-	@available(iOS, deprecated:8.0, obsoleted:9.0, message:"Use init(qos:, name:) instead.")
+	@available(OSX, deprecated:10.10, obsoleted:10.11, message:"Use init(qos:name:targeting:) instead")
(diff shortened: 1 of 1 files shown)


Sources/Scheduler.swift
```diff
@@ -169,15 +169,14 @@ public final class QueueScheduler: DateSchedulerProtocol {
...
 	@available(OSX 10.10, *)
-	public convenience init(
-		qos: DispatchQoS = .default,
-		name: String = "org.reactivecocoa.ReactiveSwift.QueueScheduler"
-	) {
-		self.init(internalQueue: DispatchQueue(
-			label: name,
-			qos: qos
-		))
+	public convenience init(qos: DispatchQoS = .default, name: String = "org.reactivecocoa.ReactiveSwift.QueueScheduler", targeting targetQueue: DispatchQueue? = nil) {
```
NachoSoto: Can you keep the formatting like it was for readability?

liscio: It seemed so…out of place. :)

I can restore it though, since that seems to be the preference.

liscio:
This latest change addresses the formatting, as well as passing the target along in the `DispatchQueue` initializer.

PR #40, opened 2016-09-25
Remove the `BindingTarget.consume` requirement.

andersio:
After bringing up #30, #38 and https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3162, IMO it is safe to draw the conclusion of`BindingTarget.consume` being pointless.

It is born to serve the default implementation of `<~(Self, Signal)`, which assumes the binding has the same lifetime as `self`. However, in practice, this is probably only valid for types that own their own lifetime e.g. `MutableProperty` and `Action`.

Let's say in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3162: `DynamicProperty` passes through its underlying object's `rac_lifetime` instead of having its own. Moreover, it also overrides the `<~` operator so that bindings act directly on the `NSObject` via KVC. This renders `BindingTarget.consume` useless, and in fact for any conforming types that overrides `<~` it is useless.

Therefore, this PR proposes remove both:

1. the `consume(_:)` requirement; and
2. the default implementation of `<~(Self, Signal)`.

It also restores the use of concrete types of signals and producers in the `BindingTargetProtocol` extensions and requirements, since the protocols of these are just used for constraining the extensions. They can be removed once concrete same-type requirements are allowed in Swift (ref: [the Completing Generic manifesto](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160229/011666.html)

~~P.S. `BindingTarget._type` is a ugly workaround, but I can't find a way to make the type inferrer infer from the `<~` conformance.~~


andersio:
I am not sure if what you all would prefer:

1. Easy conformance by implementing `consume(_:)`. But for those overriding the `<~` operator, `consume(_:)` is useless as it is only used by the default implementation of `<~` being overrides.

2. Forcing all conforming types to implement their own `<~`.

Sources/Property.swift
```diff
@@ -25,18 +25,13 @@ public protocol PropertyProtocol: class {
 }
 
 /// Represents an observable property that can be mutated directly.
-public protocol MutablePropertyProtocol: PropertyProtocol, BindingTarget {
+public protocol MutablePropertyProtocol: BindingTarget, PropertyProtocol {
+	associatedtype Value
```
andersio: this is the only working combination in Swift 3.0 GM I've found so far.

liscio:
> Easy conformance: just consume(_:). But for those overriding <~(Self, Signal), consume(_:) is useless but still needs to be implemented. (Currently in master)

I like the idea of easy conformance, however with the introduction of `BindingTarget` I think that this point is moot. We should drive API users to expose a `BindingTarget` class instance rather than e.g. declaring `BindingTargetProtocol` conformance directly on their types, so the # of additions of conforming types will be small once all this work is complete.

> Forcing all conforming types to implement their own <~(Self, Signal). (This PR)

The downside here is that we may lose the centralized documentation for the `<~` operator in the IDE. (I hope I'm not the only person that command-clicks on custom operators in Xcode?)

But this is a very small downside. The indexer *barely works* most of the time for me, anyway…

andersio:
The latest commit restores the use of concrete types of signals and producers in the `BindingTargetProtocol` extensions and requirements, since the protocols of these are just used for constraining the extensions. They can be removed once concrete same-type requirements are allowed in Swift (ref: [the Completing Generic manifesto](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160229/011666.html)

mdiep:
I don't understand the issue is here. 😕 

Can you try to explain again why `consume` is useless? It seems useful in #38.

> Let's say in ReactiveCocoa/ReactiveCocoa#3162: DynamicProperty passes through its underlying object's rac_lifetime instead of having its own. Moreover, it also overrides the <~ operator so that bindings act directly on the NSObject via KVC. This renders BindingTarget.consume useless, and in fact for any conforming types that overrides <~ it is useless.

`lifetime` is part of `BindingTargetProtocol`. Why isn't that enough to do this?

andersio:
@mdiep The default implementation of `<~` holds a _weak_ reference to `self`, and calls `consume` for every new value. In the case of `DynamicProperty` of https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3162, the `DynamicProperty` is no longer retained by the underlying object, while passing through the lifetime of its underlying object. This means `self` could deinitialise before the binding ends, thus stopping the values from propagating. Therefore, `DynamicProperty` overrides `<~`, and establishes bindings that acts directly on the object.

liscio:
> The latest commit restores the use of concrete types of signals and producers in the BindingTargetProtocol extensions and requirements, since the protocols of these are just used for constraining the extensions.

Yes! I was trying to achieve this myself and could not get it to work out because I was going about it all wrong. I was trying to declare an associated type that represented the incoming signal, but then that limited my ability to have >1 type of incoming source (i.e. both signals *and* producers.)

Sources/Property.swift
```diff
@@ -640,4 +635,42 @@ public final class MutableProperty<Value>: MutablePropertyProtocol {
...
+	///
+	/// - parameters:
+	///   - target: A target to be bond to.
+	///   - signal: A signal to bind.
+	///
+	/// - returns: A disposable that can be used to terminate binding before the
+	///            deinitialization of the target or the signal's `completed`
+	///            event.
+	@discardableResult
+	public static func <~ (target: MutableProperty, signal: Signal<Value, NoError>) -> Disposable? {
```
liscio: Why would we implement this on `MutableProperty` and not as an extension to `MutablePropertyProtocol`? Wouldn't that give us the `<~` implementation on `DynamicProperty` "for free"?

andersio: Assuming https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3162 get a pass, `DynamicProperty` would have its own `<~` implementation. This as the default implementation is only useful if we keep the current easy-to-leak-because-of-the-stealth-retain `DynamicProperty`.

liscio: Ah yes, makes sense. Thanks!

mdiep:
That description helped a lot. Thank you! Now I understand what you're trying to do with the lifetimes of the various objects.

I'll try to find some time this weekend to think about what the semantics should be so we can move forward.

mdiep:
I've commented in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3162 and #38, but I don't think the _throw-away proxy_ semantics are something we should adopt—at least now for ReactiveSwift 1.0. So I think it makes sense to keep `BindingTarget.consume`.

andersio:
The "throw-away" semantic is used by `BindingTarget` (the class) in the sense that the stateless binding target needn't be retained & uniqued.

mdiep:
I think those semantics are fine for `BindingTarget` since it was created specifically for those purpose. But I don't think they're the right semantics for the protocol.

PR #43, opened 2016-09-26
Make it buildable on Linux

ikesyo:


Sources/Scheduler.swift
```diff
@@ -6,8 +6,13 @@
 //  Copyright (c) 2014 GitHub. All rights reserved.
 //
 
+import Dispatch
 import Foundation
 
+#if os(Linux)
+	import CDispatch
```
ikesyo: This is required for `NSEC_PER_SEC` (https://github.com/Quick/Nimble/pull/343).

Sources/Scheduler.swift
```diff
@@ -85,7 +90,11 @@ public final class UIScheduler: SchedulerProtocol {
 			                               value: dispatchSpecificValue)
 	}()
 
+	#if os(Linux)
+	private var queueLength: Atomic<Int32> = Atomic(0)
```
ikesyo: I'm not sure if this is a right way. Please advise me if there is a better way to implement the functionality. 🙏

andersio: It is fine, though I wonder if you can access `atomic_*` from C++ atomics?

Sources/Scheduler.swift
NachoSoto: Can you put this as a comment on the code?

Sources/Scheduler.swift
```diff
@@ -113,10 +122,21 @@ public final class UIScheduler: SchedulerProtocol {
 				action()
 			}
 
-			OSAtomicDecrement32(&self.queueLength)
+			#if os(Linux)
```
NachoSoto: I think these implementation details shouldn't be here. Maybe we can make an `internal class IntAtomic: AtomicProtocol`, so that this class works evenly regardless of the actual type being used?

ikesyo: I'm not so sure how does it look like... 🤔

PR #44, opened 2016-09-26
Make links work again

milch:
None of the links that referenced types in the Documentation worked anymore, so I updated them to the current version.


mdiep:
Thanks! Missed these in the conversion.

Issue #33, opened 2016-09-23
Replace TimeInterval in APIs with DispatchTimeInterval


NachoSoto:
## Reasoning:

The current `TimeInterval` APIs are fragile and dangerous for many reasons:
- We always end up converting to milliseconds, but if the input value is too high, it will overflow (#31 fixed this to some extent by reducing the likelihood, but it's still possible).
- The API is not strongly typed: there is no guarantee or way to enforce users sending the right magnitude (seconds versus milliseconds, etc), other than a weak comment. 

But with `Swift 3` we can do better: using `DispatchTimeInterval` solves all these problems :)
## Process:

So I'm proposing deprecating current APIs that use `TimeInterval` to point users to use new `DispatchTimeInterval` versions.

ikesyo:
We probably need an extension like this:

```swift
extension Date {
    internal func addingDispatchTimeInterval(_ interval: DispatchTimeInterval) -> Date {
        switch interval {
        case let .seconds(s): ...
        case let .milliseconds(ms): ...
        ....
        }
    }
}
```

liscio:
> We probably need an extension like this:

Should we not instead remove the use of `Date` entirely in favour of `DispatchTime`? I never did like using `{NS,}Date` to represent a point in time…

The usage gets much nicer, too. Compare using `Date().addingDispatchTimeInterval(.milliseconds(300))` vs. `.now() + .milliseconds(300)`, for example.

NachoSoto:
Yup 👍

liscio:
I've gone ahead and done much of the work already here for a whole-hog replacement of `Date` and `TimeInterval`: https://github.com/ReactiveCocoa/ReactiveSwift/compare/master...liscio:cl-dispatchwalltime

Unfortunately, it appears that `DispatchWallTime` (and `DispatchTime`, but we don't want to base our scheduler on that) is busted in the Swift standard library. I filed a bug [SR-2807](https://bugs.swift.org/browse/SR-2807) to track the issues I encountered. Specifically, the crash when comparing two `DispatchWallTime`s is going to block progress.

liscio:
I've *also* gone ahead and opened #48 with the approach that was originally proposed, since it's not practical to expect the above issue to get corrected in time for 1.0.

ashfurrow:
It looks like this was fixed by #48 (good job!). Is the 1.0 milestone up to date? This is the only open issue left in it at the moment.

mdiep:
[...]

PR #48, opened 2016-09-30
Use DispatchTimeInterval to offset Date values

liscio:
This work addresses #33.

There are still things to do before I'll pull the WIP status…
- [x] Needs testing for the internal extensions on `Date` / `DispatchTimeInterval`
  - [x] Test the interval scaling using the `*` operator
  - [x] Test all the variants of `DispatchTimeInterval` conversions
- [x] Internal extensions should be relocated

### DIFF
--- a/Sources/Scheduler.swift
+++ b/Sources/Scheduler.swift
@@ -6,7 +6,12 @@
 //  Copyright (c) 2014 GitHub. All rights reserved.
 //
 
+import Dispatch
 import Foundation
+
+#if os(Linux)
+	import CDispatch
+#endif
 
 /// Represents a serial queue of work items.
 public protocol SchedulerProtocol {
@@ -85,7 +90,11 @@ public final class UIScheduler: SchedulerProtocol {
 			                               value: dispatchSpecificValue)
 	}()
 
+	#if os(Linux)
+	private var queueLength: Atomic<Int32> = Atomic(0)
+	#else
 	private var queueLength: Int32 = 0
+	#endif
 
 	/// Initializes `UIScheduler`
 	public init() {
@@ -113,10 +122,21 @@ public final class UIScheduler: SchedulerProtocol {
 				action()
 			}
 
-			OSAtomicDecrement32(&self.queueLength)
+			#if os(Linux)
+				self.queueLength.modify { $0 -= 1 }
+			#else
+				OSAtomicDecrement32(&self.queueLength)
+			#endif
 		}
 
-		let queued = OSAtomicIncrement32(&queueLength)
+		#if os(Linux)
+			let queued = self.queueLength.modify { value -> Int32 in
+				value += 1
+				return value
+			}
+		#else
+			let queued = OSAtomicIncrement32(&queueLength)
+		#endif
 
 		// If we're already running on the main queue, and there isn't work
 		// already enqueued, we can skip scheduling and just execute directly.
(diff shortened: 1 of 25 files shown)


Sources/Scheduler.swift
```diff
@@ -13,6 +13,48 @@ import Foundation
...
 #endif
 
+extension Date {
+	internal func addingTimeInterval(_ interval: DispatchTimeInterval) -> Date {
+		return addingTimeInterval(interval.asTimeInterval())
+	}
+}
+
+extension DispatchTimeInterval {
+	internal func asTimeInterval() -> TimeInterval {
```
NachoSoto: This would be better as a derived property instead of a function.

Sources/Scheduler.swift
```diff
@@ -13,6 +13,48 @@ import Foundation
...
+			return .nanoseconds(-ns)
+		}
+	}
+
+	/// Scales a time interval by the given scalar specified in `rhs`.
+	///
+	/// - returns: Scaled interval in microseconds
+	internal static func *(lhs: DispatchTimeInterval, rhs: Double) -> DispatchTimeInterval {
+		let seconds = lhs.asTimeInterval() * rhs
+		return .microseconds(Int(seconds * 1000 * 1000))
```
NachoSoto: This is falling into one of the points I wanted to fix with this approach: potential overflows.

Sources/Scheduler.swift
```diff
@@ -57,7 +99,7 @@ public protocol DateSchedulerProtocol: SchedulerProtocol {
 	/// - returns: Optional `Disposable` that can be used to cancel the work
 	///            before it begins.
 	@discardableResult
-	func schedule(after date: Date, interval: TimeInterval, leeway: TimeInterval, action: @escaping () -> Void) -> Disposable?
```
NachoSoto: Awesome. Can you add the deprecations for all these?

Sources/Scheduler.swift
liscio: @NachoSoto would you keep the `asTimeInterval` name? I've not seen precedent for this pattern yet outside of Apple's use of `cgColor` (i.e. using the type name itself) except it'd be strange to call a `DispatchTimeInterval` method called `timeInterval`. 😛

Sources/Scheduler.swift
liscio: @NachoSoto So this is *only* used internally to "scale down" a time interval. Specifically it's used only to scale intervals to 10% of their original value for the default leeway parameter.

But you are correct. If seconds is over ~200,000, 10% is 2000, and hence we end up with ~2billion which will overflow a signed integer *on 32-bit platforms* (of which there aren't really any more for much longer, anyway…)

But I'm pretty sure that even 200k seconds is going to be a rarely (if ever) specified interval for our APIs.

One of the tests I plan to add is to find this tipping point.

liscio:
@NachoSoto I'm pretty sure I addressed everything you pointed out. Let me know if I missed anything (or created any new issues.) Removing [WIP] tag…

Sources/Scheduler.swift
liscio: I added them all in the "deprecations bucket file."

liscio:
@NachoSoto Everything look OK to you for this to go in? 

Any others want to take a look? @mdiep, @andersio? The related issue (#33) is the last item on the 1.0 milestone.

mdiep:
I'll try to take a look in the next few days unless someone beats me to it.

I did put #22 in the 1.0 milestone just now because I think that needs to be addressed.

mdiep:
Brilliant. Thanks @liscio!

PR #49, opened 2016-09-30
Use Schedulers in `BindingTarget`.

andersio:
When making progress for the Rex PR, I hit a ~~weird~~ case which:

1. `Thread.isMainThread` is true;
2. `DispatchQueue.main.getSpecific(key: specificKey)` gives me the right queue id (for the main queue);
3. BUT `DispatchQueue.getSpecific(key: specificKey)` gives `nil`.

~~It seems it might be because of this:~~

> ~~https://developer.apple.com/reference/dispatch/1453125-dispatch_get_specific~~
>
> ~~This function is intended to be called from a block executing in a dispatch queue. You use it to obtain context data associated with the queue. Calling this method from code not running in a dispatch queue returns NULL because there is no queue to provide context.~~

The issue is in fact caused by the implicit nesting of `DispatchQueue.sync`. Somewhere in the middle of the call stack was a `sync` to another queue, causing the current context being bound to the private queue despite still being on the main thread. This PR adds the ability to elide the `sync` call for this particular case.

### DIFF
--- a/Sources/UnidirectionalBinding.swift
+++ b/Sources/UnidirectionalBinding.swift
@@ -1,3 +1,4 @@
+import Foundation
 import Dispatch
 import enum Result.NoError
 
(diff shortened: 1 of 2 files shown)


Sources/UnidirectionalBinding.swift
```diff
@@ -304,14 +305,17 @@ public final class BindingTarget<Value>: BindingTargetProtocol {
...
+		let usesMainQueue = queue == DispatchQueue.main
 
 		/// Ensures the queue has been setup property.
 		if nil == queue.getSpecific(key: specificKey) {
 			queue.setSpecific(key: specificKey, value: queueId)
 		}
 
 		let setter: (Value) -> Void = { value in
-			if queueId == DispatchQueue.getSpecific(key: specificKey) {
+			if usesMainQueue && Thread.isMainThread {
```
liscio: What if we instead did this, and killed the `Bool` variable?

```
if queue == DispatchQueue.main && Thread.isMainThread {
```

That, or `s/usesMainQueue/onMainQueue` so it communicates the intent a little more clearly. I know I'm splitting hairs here, but nobody's really *using* the main queue.

andersio: `queue` != current queue.

andersio: We just need to evaluate it once for all invocations.

mdiep: We shouldn't be using the main thread: https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2912.

It seems like this API shouldn't be using queues at all—it should use schedulers. I don't know if that will have any affect on the issue at hand.

andersio: `Scheduler` is asynchronous. Using it would make all Cocoa binding targets to accept values asynchronously.

andersio: Well it would still fix it, since the block would be dispatched asynchronously (not on the main queue). Perhaps for correctness' sake we should not jump the queue this way too.

Tests/ReactiveSwiftTests/UnidirectionalBindingSpec.swift
```diff
@@ -50,6 +50,22 @@ class UnidirectionalBindingSpec: QuickSpec {
...
+				                       lifetime: lifetime,
+				                       setter: { value = $0 })
+
+				let property = MutableProperty(1)
+
+				queue.sync {
+					_ = target <~ property
+				}
+
+				expect(value).toEventually(equal(1))
```
andersio: This was `expect(value) == 1` as the binding target was guaranteed to be synchronous before.

Sources/UnidirectionalBinding.swift
sharplet: It wouldn't be "by default" though, would it? If you use the designated initialiser you can still create a synchronous binding, but if you use the `onScheduler:` form it will most likely be async (since `UIScheduler` is meant to skip asynchronous dispatch if possible, right?).

andersio: Yeah, you are right. I just meant for this initializer async is the default (with limited reentrancy from `UIScheduler`).

PR #51, opened 2016-10-01
Introduce `Reactive` to host reactive extensions.

andersio:


andersio:
Hmm, my second thought strikes me that the reactivity is a bad choice (the intensity of it being reactive, which is irrelevant here). Better keep it as `Reactive`.

Sources/Reactive.swift
```diff
@@ -0,0 +1,26 @@
+/// Describes a class which has been extended with reactive elements.
+///
+/// - note: `ExtendedForReactiveness` is only intended for extensions to types
+///         that are not owned by the module. Non-conforming types may carry
+///         first-party reactive elements.
+public protocol ExtendedForReactiveness: class {}
```
mdiep: How about `ReactiveType` for this?

I know it's a bit outside of the Swift norm, but I think it could work here.

1. `Reactive` is the obvious choice, but the `struct` is `Reactive`. (Which I absolutely think is the right name.)

2. `ReactiveProtocol` might be useful to apply to `Reactive` itself

3. `ReactiveType` is meant to suggest that the type itself is reactive.

Sources/Reactive.swift
```diff
@@ -0,0 +1,26 @@
...
+
+extension ExtendedForReactiveness {
+	/// A proxy which exposes the reactivity of `self`.
+	public var rac: Reactive<Self> {
+		return Reactive(self)
+	}
+}
+
+// A proxy which exposes the reactivity of the Reactant type.
+public struct Reactive<Reactant: ExtendedForReactiveness> {
```
mdiep: I think `Reactant` should be called `Base` instead. RxSwift did that, and I think it's the right call because it matches what the Swift standard library does for similar types. (c.f. `LazySequence`)

Also, I don't think `Reactant` needs to conform to `ExtendedForReactiveness`.

Sources/Reactive.swift
```diff
@@ -0,0 +1,26 @@
+/// Describes a class which has been extended with reactive elements.
+///
+/// - note: `ExtendedForReactiveness` is only intended for extensions to types
+///         that are not owned by the module. Non-conforming types may carry
+///         first-party reactive elements.
+public protocol ExtendedForReactiveness: class {}
+
+extension ExtendedForReactiveness {
+	/// A proxy which exposes the reactivity of `self`.
+	public var rac: Reactive<Self> {
```
mdiep: I feel a little torn about using `rac` in ReactiveSwift, but (1) I can't think of anything better and (2) it does seem nice to use the same thing everywhere since they'll lead to the same type anyway.

Sources/Reactive.swift
andersio: It was `Reactive` and `Reactivity` before, but I changed it to `ExtendedForReactiveness` because `Reactive` somehow sounds like implying non-conforming types of `Reactive` are not reactive.

Sources/Reactive.swift
andersio: Changed.

Sources/Reactive.swift
mdiep: `ReactiveExtensible`? `ReactiveEnabled`?

> Reactive* somehow sounds like implying non-conforming types of Reactive* are not reactive.

If you think of it as implying they're not `Reactive`, I think it works. 😕

andersio: `ReactiveExtensionProviding`, heh?

Sources/Reactive.swift
```diff
@@ -0,0 +1,26 @@
...
+		return Reactive(self)
+	}
+}
+
+// A `Reactive` proxy hosts reactive extensions to `Base`.
+public struct Reactive<Base> {
```
NachoSoto: This type is too general, and it also has no guarantees of having value semantics.

Sources/Reactive.swift
NachoSoto: What's the reason behind adding this extra level of indirection?

mdiep: To prevent name collisions in a slightly nicer way than using the `rac_` prefix. (https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3049)

Sources/Reactive.swift
mdiep: How so? This seems akin to Swift's `LazySequence`.

Sources/Reactive.swift
andersio: Properties do not have overloading at this moment. Even if they would do (planned), we would probably want to avoid return type ambiguity as much as possible, wouldn't we?

Sources/Reactive.swift
andersio: The proxy itself does not need an identity, and we care about just the reference it holds.

RuiAAPeres:
This is really interesting. Thanks @andersio. ❤️

Sources/Reactive.swift
liscio: Count me in for a strong 👎 on any strange/uncommon words like `Reactant`, `Reactivity`, `Reactiveness` because I feel they are forced adjectives. Reactive itself is an adjective, so there's no need to push it further.

I'll throw another few ideas out: `ReactiveExtensionsHosting`, `ReactiveExtensionsProtocol`, `ExtendedForReactiveSwift`, `ExposesReactiveBindings`, `ExposesReactiveExtensions`. I can go on all day… 😄

Sources/Reactive.swift
liscio: Why not `reactive`? Consider usage:

```
// "Give me rac notifications for…"
NSNotificationCenter.default.rac.notifications(forName:

// "Give me rac data with…"
someURLSession.rac.data(with:

// "Give me reactive notifications for…"
NSNotificationCenter.default.reactive.notifications(forName:

// "Give me reactive data with…"
someURLSession.reactive.data(with:
```

The latter feels more readable/sensical to me.

Sources/Reactive.swift
sharplet: How about `ReactiveExtensionProvider`? I think that describes the role well.

Sources/Reactive.swift
andersio: If we weight clarify above brevity - which is what the Swift API guideline prefers - then yes, `reactive` is a sensible choice.

Tests/ReactiveSwiftTests/FoundationExtensionsSpec.swift
```diff
@@ -20,11 +20,11 @@ extension Notification.Name {
 
 class FoundationExtensionsSpec: QuickSpec {
 	override func spec() {
-		describe("NSNotificationCenter.rac_notifications") {
+		describe("Reactivity<NotificationCenter>.notifications") {
```
liscio: `s/Reactivity/Reactive/` to match the rest of the changes.

Sources/Reactive.swift
```diff
@@ -0,0 +1,26 @@
+/// Describes a provider of reactive extensions.
+///
+/// - note: `ReactiveExtensionProvider` does not indicate whether a type is
+///         reactive. It is intended for extensions to types that do not own by
+///         the module so as to avoid name collision and return type ambiguity.
+public protocol ReactiveExtensionProvider: class {}
```
liscio: Just a little nit: Should it be `ReactiveExtensionsProvider` (i.e. the plural) instead, since there are often more than one extension being provided?

andersio: I think we'd rather treat all extension methods and properties as a single bloc under `reactive`.

liscio: @andersio Sorry, I just meant that the collection of e.g. `stringValue`, `isEnabled`, etc. forces the plural. Not that we'd have a "sibling" for `reactive`.

Sources/Reactive.swift
```diff
@@ -0,0 +1,26 @@
+/// Describes a provider of reactive extensions.
+///
+/// - note: `ReactiveExtensionsProvider` does not indicate whether a type is
+///         reactive. It is intended for extensions to types that do not own by
+///         the module so as to avoid name collision and return type ambiguity.
```
mdiep: The wording here is a little off. Should be something like this:

> It is intended for extensions to types that are not owned by the module in order to prevent name collisions and return type ambiguity.

Sources/Reactive.swift
mdiep: I still don't love this name. But I guess it's good enough for now. 😕

Sources/Reactive.swift
andersio: Ouch, passive voice.

mdiep:
@NachoSoto Did you want to take another look at this before we merge?

Sources/FoundationExtensions.swift
```diff
@@ -24,29 +26,33 @@ extension NotificationCenter {
 	///         will terminate immediately with an `interrupted` event.
 	///         Otherwise, the producer will not terminate naturally, so it must
 	///         be explicitly disposed to avoid leaks.
-	public func rac_notifications(forName name: Notification.Name?, object: AnyObject? = nil) -> SignalProducer<Notification, NoError> {
+	public func notifications(forName name: Notification.Name?, object: AnyObject? = nil) -> SignalProducer<Notification, NoError> {
 		// We're weakly capturing an optional reference here, which makes destructuring awkward.
 		let objectWasNil = (object == nil)
-		return SignalProducer { [weak object] observer, disposable in
+		return SignalProducer { [base, weak object] observer, disposable in
```
NachoSoto: Can you use `self.base` here to be explicit please? It took me a while to figure out where this value was coming from, because it's not in the current context.

Sources/FoundationExtensions.swift
```diff
@@ -60,9 +66,9 @@ extension URLSession {
 	/// - note: This method will not send an error event in the case of a server
 	///         side error (i.e. when a response with status code other than
 	///         200...299 is received).
-	public func rac_data(with request: URLRequest) -> SignalProducer<(Data, URLResponse), NSError> {
-		return SignalProducer { observer, disposable in
-			let task = self.dataTask(with: request) { data, response, error in
+	public func data(with request: URLRequest) -> SignalProducer<(Data, URLResponse), NSError> {
+		return SignalProducer { [base] observer, disposable in
```
NachoSoto: Ditto

Sources/FoundationExtensions.swift
andersio: Done in https://github.com/ReactiveCocoa/ReactiveSwift/pull/51/commits/046c65192655576c1cbd0ecb95d66ffe776f5d20.

Sources/FoundationExtensions.swift
```diff
@@ -29,18 +29,18 @@ extension Reactive where Base: NotificationCenter {
...
 		let objectWasNil = (object == nil)
-		return SignalProducer { [base, weak object] observer, disposable in
+		return SignalProducer { [weak object] observer, disposable in
 			guard object != nil || objectWasNil else {
 				observer.sendInterrupted()
 				return
 			}
 
-			let notificationObserver = base.addObserver(forName: name, object: object, queue: nil) { notification in
+			let notificationObserver = self.base.addObserver(forName: name, object: object, queue: nil) { notification in
```
andersio: Changed.

mdiep:
🎉

lukaskubanek:
@andersio This is such a great idea!

PR #52, opened 2016-10-01
MutablePropertyFacade

andersio:
Before digging into this, please have a look at https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3229.

-

Examples: [NSTextField](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/a904aca0cb8b2a4b1b57dd5dc50889960c1b436e/ReactiveCocoa/AppKit/NSTextField.swift), [UIControl](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/a904aca0cb8b2a4b1b57dd5dc50889960c1b436e/ReactiveCocoa/UIKit/UIControl.swift#L49), https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3210/commits/a904aca0cb8b2a4b1b57dd5dc50889960c1b436e

> A new type MutablePropertyFacade is crafted for UIControl/NSControl. It is like a blend of BindingTarget and MutableProperty.

> Rationale:
>
>1. The current MutableProperty-hijacking approach abuses the fact that manually setting the state would not send an action message.

>    In other words, for each user-triggered action message, the property would manually set the state _again_ with the fresh value. While it probably did no harm in practice, this doesn't seem ideal IMO.
>
> 2. Vulnerable to accidental retainment which delays the termination of bindings.

(*I am bad at naming though. Feel free to suggest a better one*)

### DIFF
--- a/Sources/Property.swift
+++ b/Sources/Property.swift
@@ -641,3 +641,72 @@ public final class MutableProperty<Value>: MutablePropertyProtocol {
 		observer.sendCompleted()
 	}
 }
+
+/// A facade that provides `MutableProperty` like semantics.
+public final class MutablePropertyFacade<Value>: MutablePropertyProtocol {
+	private let getter: () -> Value
+	private let setter: (Value) -> Void
+	private let scheduler: SchedulerProtocol?
+
+	/// The lifetime of the property.
+	public let lifetime: Lifetime
+
+	/// A producer which emits the current value, and then forwards all
+	/// subsequent changes of the property.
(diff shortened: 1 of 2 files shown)


liscio:
I'm not loving this idea _at all_, as I think we're heading down the wrong path if we continue trying to force properties onto controls. More on this here: https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3229#issuecomment-251395000

PR #57, opened 2016-10-05
Revert `Observer` back to be a `struct`.

andersio:
This seems to be an artefact from https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3120. 

`Observer` being `class` was solely to serve the internal implementation of `replayLazily`.


andersio:
While closures themselves do not have a concept of identity in Swift, every subscription to the `Signal` is unique to each other in the context of `Signal`, even if the underlying action is the same. This should be enough to justify the change.

NachoSoto:
Closures DO have reference semantics.

PR #58, opened 2016-10-06
Add throttle(while:on:)

sharplet:
There are times where you'd like to throttle values whenever some condition is met, rather than based on a time interval.

This is inspired by ReactiveViewModel's `-throttleSignalWhileInactive:`.

Basic usage might involve a simple `MutableProperty` to toggle throttled values:

``` swift
let isActive = MutableProperty(false)

label.reactive.text <~ somethingChanged
  .throttle(while: isActive.map(!), on: QueueScheduler())
  .flatMap(.latest, transform: doSomethingExpensive)
  .observeOn(UIScheduler())

// later

isActive.value = true
```

At some point it would be great to be able to throttle events while the view has disappeared:

``` swift
somethingChanged
  .throttle(while: reactive.viewHasDisappeared, on: UIScheduler())
  .startWithValues {
    // expensive layout stuff
  }
```

I'd love some feedback on this, especially potential thread safety issues! For example, I'm a little concerned that `shouldThrottle` could technically have changed between reading `shouldThrottle.value` and starting `shouldThrottle.producer.uniqueValues { $0 }`.

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -1565,6 +1565,10 @@ extension SignalProtocol {
 	///         that value will be discarded and the returned signal will
 	///         terminate immediately.
 	///
+	/// - note: If `shouldThrottle` completes before the receiver, and its last
+	///         value is `true`, the returned signal will remain in the throttled
+	///         state, emitting no further values until it terminates.
+	///
 	/// - parameters:
 	///   - shouldThrottle: A boolean property that controls whether values
 	///                     should be throttled.
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -1586,19 +1586,20 @@ extension SignalProtocol {
 			disposable += shouldThrottle.producer
 				.skipRepeats()
 				.startWithValues { shouldThrottle in
-					var valueToSend: Value?
-
-					state.modify { state -> ThrottleWhileState<Value> in
-						guard !state.isTerminated else { return state }
+					let valueToSend = state.modify { state -> Value? in
+						guard !state.isTerminated else { return nil }
 
 						if shouldThrottle {
-							return .throttled(nil)
+							state = .throttled(nil)
 						} else {
+							defer { state = .resumed }
+
 							if case let .throttled(value?) = state {
-								valueToSend = value
+								return value
 							}
-							return .resumed
 						}
+
+						return nil
 					}
 
 					if let value = valueToSend {
@@ -1609,24 +1610,22 @@ extension SignalProtocol {
 				}
 
 			disposable += self.observe { event in
-				var eventToSend: Event<Value, Error>?
-
-				state.modify { state -> ThrottleWhileState<Value> in
+				let eventToSend = state.modify { state -> Event<Value, Error>? in
 					switch event {
 					case let .value(value):
 						switch state {
 						case .throttled:
-							return .throttled(value)
+							state = .throttled(value)
+							return nil
 						case .resumed:
-							eventToSend = event
-							return state
+							return event
 						case .terminated:
-							return state
+							return nil
 						}
 
 					case .completed, .interrupted, .failed:
-						eventToSend = event
-						return .terminated
+						state = .terminated
+						return event
 					}
 				}
 
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -1575,7 +1575,7 @@ extension SignalProtocol {
 		where P.Value == Bool
 	{
 		return Signal { observer in
-			let initial: ThrottleWhileState<Value> = shouldThrottle.value ? .throttled(nil) : .resumed
+			let initial: ThrottleWhileState<Value> = .resumed
 			let state = Atomic(initial)
 			let schedulerDisposable = SerialDisposable()
 
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -1581,7 +1581,6 @@ extension SignalProtocol {
 
 			let disposable = CompositeDisposable()
 			disposable += schedulerDisposable
-			disposable += { _ = shouldThrottle }
 
 			disposable += shouldThrottle.producer
 				.skipRepeats()
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -1556,10 +1556,9 @@ extension SignalProtocol {
 	/// `shouldThrottle` is true, forwarding values on the given scheduler.
 	///
 	/// - note: While `shouldThrottle` remains false, values are forwarded on the
-	///         given scheduler.
-	///
-	/// - note: If multiple values are received while `shouldThrottle` is true,
-	///         the latest value is the one that will be passed on.
+	///         given scheduler. If multiple values are received while
+	///         `shouldThrottle` is true, the latest value is the one that will
+	///         be passed on.
 	///
 	/// - note: If the input signal terminates while a value is being throttled,
 	///         that value will be discarded and the returned signal will
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -1586,6 +1586,8 @@ extension SignalProtocol {
 			disposable += shouldThrottle.producer
 				.skipRepeats()
 				.startWithValues { shouldThrottle in
+					var valueToSend: Value?
+
 					state.modify { state -> ThrottleWhileState<Value> in
 						guard !state.isTerminated else { return state }
 
@@ -1593,16 +1595,22 @@ extension SignalProtocol {
 							return .throttled(nil)
 						} else {
 							if case let .throttled(value?) = state {
-								schedulerDisposable.innerDisposable = scheduler.schedule {
-									observer.send(value: value)
-								}
+								valueToSend = value
 							}
 							return .resumed
+						}
+					}
+
+					if let value = valueToSend {
+						schedulerDisposable.innerDisposable = scheduler.schedule {
+							observer.send(value: value)
 						}
 					}
 				}
 
 			disposable += self.observe { event in
+				var eventToSend: Event<Value, Error>?
+
 				state.modify { state -> ThrottleWhileState<Value> in
 					switch event {
 					case let .value(value):
@@ -1610,19 +1618,21 @@ extension SignalProtocol {
 						case .throttled:
 							return .throttled(value)
 						case .resumed:
-							schedulerDisposable.innerDisposable = scheduler.schedule {
-								observer.action(event)
-							}
+							eventToSend = event
 							return state
 						case .terminated:
 							return state
 						}
 
 					case .completed, .interrupted, .failed:
-						schedulerDisposable.innerDisposable = scheduler.schedule {
-							observer.action(event)
-						}
+						eventToSend = event
 						return .terminated
+					}
+				}
+
+				if let event = eventToSend {
+					schedulerDisposable.innerDisposable = scheduler.schedule {
+						observer.action(event)
 					}
 				}
 			}
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -1552,6 +1552,85 @@ extension SignalProtocol {
 		}
 	}
 
+	/// Conditionally throttles values sent on the receiver whenever
+	/// `shouldThrottle` is true, forwarding values on the given scheduler.
+	///
+	/// - note: While `shouldThrottle` remains false, values are forwarded on the
+	///         given scheduler.
+	///
+	/// - note: If multiple values are received while `shouldThrottle` is true,
+	///         the latest value is the one that will be passed on.
+	///
+	/// - note: If the input signal terminates while a value is being throttled,
+	///         that value will be discarded and the returned signal will
+	///         terminate immediately.
+	///
+	/// - parameters:
+	///   - shouldThrottle: A boolean property that controls whether values
+	///                     should be throttled.
+	///   - scheduler: A scheduler to deliver events on.
+	///
+	/// - returns: A signal that sends values only while `shouldThrottle` is false.
+	public func throttle<P: PropertyProtocol>(while shouldThrottle: P, on scheduler: SchedulerProtocol) -> Signal<Value, Error>
+		where P.Value == Bool
+	{
+		return Signal { observer in
+			let initial: ThrottleWhileState<Value> = shouldThrottle.value ? .throttled(nil) : .resumed
+			let state = Atomic(initial)
+			let schedulerDisposable = SerialDisposable()
+
+			let disposable = CompositeDisposable()
+			disposable += schedulerDisposable
+			disposable += { _ = shouldThrottle }
+
+			disposable += shouldThrottle.producer
+				.skipRepeats()
+				.startWithValues { shouldThrottle in
+					state.modify { state -> ThrottleWhileState<Value> in
+						guard !state.isTerminated else { return state }
+
+						if shouldThrottle {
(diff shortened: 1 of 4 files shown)


sharplet:
There are also no tests yet, which I'm planning to rectify. I should be able to cover the main semantics, although I don't trust I can right tests to verify this code is thread-safe.

Sources/Signal.swift
```diff
@@ -1492,6 +1492,83 @@ extension SignalProtocol {
...
+	///         that value will be discarded and the returned signal will
+	///         terminate immediately.
+	///
+	/// - parameters:
+	///   - shouldThrottle: A boolean property that controls whether values
+	///                     should be throttled.
+	///   - scheduler: A scheduler to deliver events on.
+	///
+	/// - returns: A signal that sends values only while `shouldThrottle` is false.
+	func throttle<P: PropertyProtocol>(while shouldThrottle: P, on scheduler: SchedulerProtocol) -> Signal<Value, Error>
```
mdiep: This seems like it should take a Signal, not a Property. 😕 

It should be easy to go from the property to the signal, but it's not easy to go the other way.

Sources/Signal.swift
```diff
@@ -1492,6 +1492,83 @@ extension SignalProtocol {
...
+		return Signal { observer in
+			let initial: ThrottleWhileState<Value> = shouldThrottle.value ? .throttled(nil) : .resumed
+			let state = Atomic(initial)
+			let schedulerDisposable = SerialDisposable()
+
+			let disposable = CompositeDisposable()
+			disposable += schedulerDisposable
+			disposable += { _ = shouldThrottle }
+
+			disposable += shouldThrottle.producer.uniqueValues { $0 }.startWithValues { shouldThrottle in
```
mdiep: You probably meant `skipRepeats` here, not `uniqueValues`?

Also, this should be formatted according to the RAC style.

Sources/Signal.swift
```diff
@@ -1492,6 +1492,83 @@ extension SignalProtocol {
...
+				state.modify { state -> ThrottleWhileState<Value> in
+					guard !state.isTerminated else { return state }
+
+					if shouldThrottle {
+						return .throttled(nil)
+					} else {
+						if case .throttled(let value?) = state {
+							schedulerDisposable.innerDisposable = scheduler.schedule {
+								observer.send(value: value)
```
mdiep: I think this value should get sent outside of the `Atomic`'s lock to prevent deadlock.

Even though it's in a `schedule` block, that can execute synchronously if, e.g., you're on the main queue and you're using the `UIScheduler`.

Same goes for the send below.

Sources/Signal.swift
```diff
@@ -1578,6 +1655,20 @@ private struct ThrottleState<Value> {
...
+	case throttled(Value?)
+	case terminated
+
+	var isTerminated: Bool {
+		if case .terminated = self {
+			return true
+		} else {
+			return false
+		}
+	}
```
mdiep: This could just be `return self == .terminated`.

ikesyo: That requires explicit `Equatable` conformance of `ThrottleWhileState` since `case throttled` has an associated value.

sharplet: Stylistically, this could also become a switch statement if you prefer, which would add exhaustively checking.

Sources/Signal.swift
sharplet: Fixed.

Sources/Signal.swift
sharplet: Made this change in e2dd0901a461bab7f95baf34ab6d42b0b59701de, let me know what you think.

Sources/Signal.swift
sharplet: Hmm, I think I kinda see it the other way. If this was implemented in terms of `Signal`, we'd have to pick a default starting state for the throttle, probably `.resumed` so that events are passed through. Using `Property` means we don't have to make that assumption.

From a slightly different perspective, `Property` is more powerful than `Signal` in that you could implement a signal-based overload of this method by composing it with an initial value and calling through to the property-based implementation.

I pushed 4dcd2bfc92dad2e0f022606a1dc79b86e58d3a96 to demonstrate what I mean. If you think those overloads would be useful I'll add some docs and merge those changes in here.

Sources/Signal.swift
```diff
@@ -1492,6 +1492,95 @@ extension SignalProtocol {
...
+
+						if shouldThrottle {
+							return .throttled(nil)
+						} else {
+							if case .throttled(let value?) = state {
+								valueToSend = value
+							}
+							return .resumed
+						}
+					}
```
ikesyo: The returned value from `Atomic.modify` block is not a new value for the `Atomic` instance since https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2984. So this should be written as follows:

```swift
let valueToSend = state.modify { state -> Value? in
	guard !state.isTerminated else { return nil }

	if shouldThrottle {
		state = .throttled(nil)
		return nil
	} else {
		state = .resumed
		if case let .throttled(value) = state {
			return value
		} else {
			return nil
		}
	}
}
```

Sources/Signal.swift
```diff
@@ -1492,6 +1492,95 @@ extension SignalProtocol {
...
+							return state
+						case .terminated:
+							return state
+						}
+
+					case .completed, .interrupted, .failed:
+						eventToSend = event
+						return .terminated
+					}
+				}
```
ikesyo: Same as https://github.com/ReactiveCocoa/ReactiveSwift/pull/58/files#r82397528:

```swift
let eventToSend = state.modify { state -> Event<Value, Error>? in
	switch event {
	case .value(let value):
		switch state {
		case .throttled:
			state = .throttled(value)
			return nil
		case .resumed:
			return .value(value)
		case .terminated:
			return nil
		}

	case .completed, .interrupted, .failed:
		state = .terminated
		return event
	}
}
```

Sources/Signal.swift
sharplet: Good call! I've pushed 58e5bed. I ended up having a single `return nil` after the conditional block, but if everyone prefers this style (a return at every branch) I can change it.

Sources/Signal.swift
```diff
@@ -1492,6 +1492,94 @@ extension SignalProtocol {
...
+				.startWithValues { shouldThrottle in
+					let valueToSend = state.modify { state -> Value? in
+						guard !state.isTerminated else { return nil }
+
+						if shouldThrottle {
+							state = .throttled(nil)
+						} else {
+							state = .resumed
+
+							if case .throttled(let value?) = state {
```
ikesyo: You don't need to unwrap the bound value with `?` since the return type is `Value?`: `case .throttled(let value)`.

ikesyo: We should consistently place `let` right after `case`: `if case let .throttled(value)`.

Sources/Signal.swift
```diff
@@ -1492,6 +1492,94 @@ extension SignalProtocol {
...
+						schedulerDisposable.innerDisposable = scheduler.schedule {
+							observer.send(value: value)
+						}
+					}
+				}
+
+			disposable += self.observe { event in
+				let eventToSend = state.modify { state -> Event<Value, Error>? in
+					switch event {
+					case .value(let value):
```
ikesyo: Same as https://github.com/ReactiveCocoa/ReactiveSwift/pull/58/files#r82402862.

Sources/Signal.swift
```diff
@@ -1492,6 +1492,94 @@ extension SignalProtocol {
...
+						switch state {
+						case .throttled:
+							state = .throttled(value)
+							return nil
+						case .resumed:
+							return .value(value)
```
ikesyo: Could return the original `event` here?

Sources/Signal.swift
NachoSoto: > which would add exhaustively checking.

And fail to compile if one adds a new `case`, which is a good thing

sharplet:
Ok, I think I've addressed everyone's feedback, thank you! I've also pushed some tests.

@mdiep Do you think it's worth adding the `Signal` and `SignalProducer` overloads like I mentioned in https://github.com/ReactiveCocoa/ReactiveSwift/pull/58#discussion_r82393681?

mdiep:
> Do you think it's worth adding the Signal and SignalProducer overloads like I mentioned in #58

I think we can go without them for now.

Mind merging `master` in? That should fix up the SwiftPM tests.

sharplet:
And we're green 👍

Sources/Signal.swift
```diff
@@ -1492,6 +1492,94 @@ extension SignalProtocol {
...
+	///   - shouldThrottle: A boolean property that controls whether values
+	///                     should be throttled.
+	///   - scheduler: A scheduler to deliver events on.
+	///
+	/// - returns: A signal that sends values only while `shouldThrottle` is false.
+	public func throttle<P: PropertyProtocol>(while shouldThrottle: P, on scheduler: SchedulerProtocol) -> Signal<Value, Error>
+		where P.Value == Bool
+	{
+		return Signal { observer in
+			let initial: ThrottleWhileState<Value> = shouldThrottle.value ? .throttled(nil) : .resumed
```
andersio: This initial check seems unnecessary, since `shouldThrottle.producer` below would emit the current value.

sharplet:
@andersio Is this kind of thing what you had in mind? I'm trying to work out how to safely get the initial value, and also not miss any changes to `shouldThrottle`...

```diff
diff --git a/Sources/Signal.swift b/Sources/Signal.swift
index cb3adbf..941bf79 100644
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -1515,7 +1515,8 @@ extension SignalProtocol {
 		where P.Value == Bool
 	{
 		return Signal { observer in
-			let initial: ThrottleWhileState<Value> = shouldThrottle.value ? .throttled(nil) : .resumed
+			let shouldThrottleProducer = shouldThrottle.producer.replayLazily(upTo: Int.max)
+			let initial: ThrottleWhileState<Value> = shouldThrottleProducer.first()!.value! ? .throttled(nil) : .resumed
 			let state = Atomic(initial)
 			let schedulerDisposable = SerialDisposable()
 
@@ -1523,7 +1524,8 @@ extension SignalProtocol {
 			disposable += schedulerDisposable
 			disposable += { _ = shouldThrottle }
 
-			disposable += shouldThrottle.producer
+			disposable += shouldThrottleProducer
+				.skip(first: 1)
 				.skipRepeats()
 				.startWithValues { shouldThrottle in
 					let valueToSend = state.modify { state -> Value? in
```

andersio:
I mean the reverse of this. `initial` can be simply `.resumed`, because `shouldThrottleProducer` would always emit that very first value synchronously.

sharplet:
Oh, that totally clears things up. Fixed in https://github.com/ReactiveCocoa/ReactiveSwift/pull/58/commits/f757e53e1492b55cd9f0100a7b03ac324c2fa40d (and rebased).

Sources/Signal.swift
```diff
@@ -1492,6 +1492,94 @@ extension SignalProtocol {
...
+		where P.Value == Bool
+	{
+		return Signal { observer in
+			let initial: ThrottleWhileState<Value> = .resumed
+			let state = Atomic(initial)
+			let schedulerDisposable = SerialDisposable()
+
+			let disposable = CompositeDisposable()
+			disposable += schedulerDisposable
+			disposable += { _ = shouldThrottle }
```
andersio: Hmm, may I ask why would you need to retain it?

andersio: @sharplet Hmm, may I ask what's the specific reason to retain it? I don't think the resulting signal should own `shouldThrottle` in any way. What if it terminates the resulting signal when `shouldThrottle` completes, gaining the `take(during:)` semantic?

andersio: There is no signal & producer operator, as far as I can remember, which establishes themselves as an owner of their sources.

> Regardless of the current behaviour, are you saying that if shouldThrottle completes, the returned signal should or should not also complete?
If it is not retained by the resulting signal, the signal should complete when the property is gone.

sharplet: It seems to me like the options are:

1. When `shouldThrottle` completes, the signal may get stuck in the throttled state (i.e., whichever state was last)
2. When `shouldThrottle` completes, the state gets switched back to `.resumed`
3. When `shouldThrottle` completes, the signal completes

(2) feels plain wrong to me, which makes it a toss-up between (1) and (3). So I feel like it comes down to which option is more surprising, and which option is easier to see/debug. For that reason I'm leaning towards tying the lifetime of the returned signal to the lifetime of `shouldThrottle`.

@mdiep What do you think?

sharplet: Or anyone else who's been involved in the thread of course! 😅

Sources/SignalProducer.swift
```diff
@@ -1092,6 +1092,31 @@ extension SignalProducerProtocol {
...
+	/// - parameters:
+	///   - shouldThrottle: A boolean property that controls whether values
+	///                     should be throttled.
+	///   - scheduler: A scheduler to deliver events on.
+	///
+	/// - returns: A producer that sends values only while `shouldThrottle` is false.
+	public func throttle<P: PropertyProtocol>(while shouldThrottle: P, on scheduler: SchedulerProtocol) -> SignalProducer<Value, Error>
+		where P.Value == Bool
+	{
+		return lift { $0.throttle(while: shouldThrottle, on: scheduler) }
```
sharplet: This keeps a strong reference to `shouldThrottle` for `SignalProducer`s. @andersio does this need to change as well?

sharplet: The first thing I tried was to wrap `shouldThrottle` in an existential:

```swift
let shouldThrottle = Property(shouldThrottle)
return lift { $0.throttle(while: shouldThrottle, on: scheduler) }
```

But that also increases the lifetime of the underlying property. Is there any way to capture a reference to this property that won't extend its lifetime? It looks like composing properties fundamentally alters the lifetime of the source property, extending it to be the union of all transitive property lifetimes. Is that intentional?

andersio: That is intentional. I assume in this case you might need an internal `throttle(while:on:)` that takes an "unsafe" property producer, and have the two public operators depending on it.

andersio: #117 is merged. Now this is the last stop before RC1. :tada:

Sources/Signal.swift
```diff
@@ -1552,6 +1552,97 @@ extension SignalProtocol {
...
+	/// - note: If multiple values are received while `shouldThrottle` is true,
+	///         the latest value is the one that will be passed on.
+	///
+	/// - note: If the input signal terminates while a value is being throttled,
+	///         that value will be discarded and the returned signal will
+	///         terminate immediately.
+	///
+	/// - note: If `shouldThrottle` completes before the receiver, and its last
+	///         value is `true`, the returned signal will remain in the throttled
+	///         state, emitting no further values until it terminates.
```
sharplet: Is 4 notes excessive? 😅

sharplet:
That should be everything, please take a look at the latest 2 commits!

sharplet:
Streamlined to 3 notes in the doc comments.

sharplet:
Green 👍

andersio:
:shipit:

sharplet:
:tada:

PR #61, opened 2016-10-11
Convenience Lifetime methods.

andersio:
```
// Use the infix `&` and `|`?
a.lifetime.and(b.lifetime)
a.lifetime.or(b.lifetime)

// `Disposable`-ish interface.
a.lifetime += emptyingTrashCansDisposable
a.lifetime += { [weak self] in self?.emptyTrashCans() }
(lifetime += {})?.dispose()
```

**TODO**
- [ ] Use infix operators instead?
- [ ] Test Cases

### DIFF
--- a/Sources/Lifetime.swift
+++ b/Sources/Lifetime.swift
@@ -39,6 +39,58 @@ public final class Lifetime {
 		self.init(ended: token.ended)
 	}
 
+	/// Create a lifetime that is an OR of `self` and `other`.
+	///
+	/// - parameters:
+	///   - other: The lifetime to be composed with.
+	///
+	/// - returns:
+	///   An OR lifetime.
+	public func or(_ other: Lifetime) -> Lifetime {
+		return Lifetime(ended: ended.zip(with: other.ended).map { _ in })
+	}
+
+	/// Create a lifetime that is an AND of `self` and `other`.
+	///
+	/// - parameters:
+	///   - other: The lifetime to be composed with.
+	///
+	/// - returns:
+	///   An AND lifetime.
+	public func and(_ other: Lifetime) -> Lifetime {
+		return Lifetime(ended: ended.combineLatest(with: other.ended).map { _ in })
+	}
+
+	/// Attach a disposable to `lifetime`.
+	///
+	/// - parameters:
+	///   - lifetime: The lifetime being attached.
+	///   - disposable: The disposable to be attached.
+	///
+	/// - returns:
+	///   A cancel disposable to detach the disposable, or `nil` if the lifetime
+	///   has terminated.
+	@discardableResult
+	public static func += (lifetime: Lifetime, disposable: Disposable?) -> Disposable? {
+		return disposable.flatMap { disposable in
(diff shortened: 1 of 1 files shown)


liscio:
Is there somewhere in ReactiveSwift / ReactiveCocoa that would benefit from these convenience methods? 

I think it'd be helpful if the PR also included some improved internal implementations that take advantage of these, if that's the case.

Sources/Lifetime.swift
```diff
@@ -39,6 +39,58 @@ public final class Lifetime {
...
+	///
+	/// - parameters:
+	///   - lifetime: The lifetime being attached.
+	///   - disposable: The disposable to be attached.
+	///
+	/// - returns:
+	///   A cancel disposable to detach the disposable, or `nil` if the lifetime
+	///   has terminated.
+	@discardableResult
+	public static func += (lifetime: Lifetime, disposable: Disposable?) -> Disposable? {
```
mdiep: `+=` doesn't makes sense to me since this doesn't alter `lifetime`.

Sources/Lifetime.swift
```diff
@@ -39,6 +39,58 @@ public final class Lifetime {
...
+	/// - parameters:
+	///   - other: The lifetime to be composed with.
+	///
+	/// - returns:
+	///   An OR lifetime.
+	public func or(_ other: Lifetime) -> Lifetime {
+		return Lifetime(ended: ended.zip(with: other.ended).map { _ in })
+	}
+
+	/// Create a lifetime that is an AND of `self` and `other`.
```
NachoSoto: This docstring doesn't add anything that's not already in the method signature 😅
I fail to see a possible use case for these, could you give an example?

andersio: Observing a notification from `NotificationCenter` posted by another object.

```
NotificationCenter.default.reactive.notifications(forName: .Name, object: object)
    .take(during: object.reactive.lifetime.or(self.reactive.lifetime))
    // ...
```

andersio:
@liscio I just roll this out to see the initial opinion.

For composition, I am not aware of any within the code base. For the disposable & the completion action attachment though, it could trim down some boilerplates. e.g.

```
disposable += object.reactive.lifetime.ended.observeCompleted(disposable.dispose)
```

The lifetime ended signal is known to emit only a `completed` event. So IMO it is appropriate to give a concise shorthand. These can even be moved to `NSObject.reactive`.

```
disposable += object.reactive.observeEnded(disposable.dispose)

// Overload that takes a `Disposable`.
disposable += object.reactive.observeEnded(disposable)
```

Sources/Lifetime.swift
NachoSoto: > But of course these can be eliminated if notifications(forName:object:) would terminate the returned signal if object has deinitialized.

That is actually the case: https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2747.

andersio: It wasn't. It interrupts when the object is freed *at the time the producer starts*. But it never stops once it is started. It is quite obvious as there is no reference to `lifetime` in `notifications`.

PR #65, opened 2016-10-15
Post-callout side effect injection.

andersio:
All events have options of pre- and post-callout injection, except for `value`. This PR adds a `valueSent` injection point to `Signal.on` and `SignalProducer.on`. It also reordered the arguments in `SignalProducer.on` to match `Signal.on`.

**Rationale**
It wasn't possible to inject a side effect **_after_** every value is consumed by a `BindingTarget`.
```
textField.stringValue <~ viewModel.name
    .on(value: { _ in print("This happens before the consumption") })

// You can wrap a binding target in a binding target though.
didSet(textField.stringValue) {
    // My "didSet" side effect.
} <~ viewModel.name
```

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -812,16 +812,21 @@ extension SignalProtocol where Value: EventProtocol, Error == NoError {
 extension SignalProtocol {
 	/// Inject side effects to be performed upon the specified signal events.
 	///
+	/// - note: Except for `valueSent` and `disposed`, all side effects are
+	///         invoked before the call-out to the observers.
+	///
 	/// - parameters:
 	///   - event: A closure that accepts an event and is invoked on every
 	///            received event.
-	///   - value: A closure that accepts a value from `value` event.
 	///   - failed: A closure that accepts error object and is invoked for
 	///             failed event.
 	///   - completed: A closure that is invoked for `completed` event.
 	///   - interrupted: A closure that is invoked for `interrupted` event.
 	///   - terminated: A closure that is invoked for any terminating event.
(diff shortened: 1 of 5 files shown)


Sources/Signal.swift
```diff
@@ -844,6 +850,9 @@ extension SignalProtocol {
 				switch receivedEvent {
 				case let .value(v):
 					value?(v)
+					observer.send(value: v)
```
NachoSoto: I'm confused, why is this line added here too? Aren't we sending events twice now?

Sources/SignalProducer.swift
```diff
@@ -1242,32 +1242,37 @@ extension SignalProducerProtocol {
...
 		event: ((Event<Value, Error>) -> Void)? = nil,
-		value: ((Value) -> Void)? = nil,
 		failed: ((Error) -> Void)? = nil,
 		completed: (() -> Void)? = nil,
 		interrupted: (() -> Void)? = nil,
 		terminated: (() -> Void)? = nil,
-		disposed: (() -> Void)? = nil
+		disposed: (() -> Void)? = nil,
+		valueSent: ((Value) -> Void)? = nil,
+		value: ((Value) -> Void)? = nil
```
NachoSoto: The ordering was on purpose, can you move `value` up again?

Sources/Signal.swift
andersio: There is a `return` below as an early out.

andersio:
@NachoSoto True. Just I almost forgot that `BindingTarget` could be asynchronous.

PR #66, opened 2016-10-17
Created internal SignalProducer.startAndRetrieveSignal to simplify some implementations

NachoSoto:
I've noticed this pattern being used in many places (also used in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3210), so this finally simplifies it.
Note the docstring explaining why we don't want to actually expose this, it's very important as it most certainly would be misused.


ikesyo:
Some `Signal.merge()` tests are failing 😱

andersio:
The flatten operator should be applied before the retrieval, since values are sent in the start handler.

NachoSoto:
Huh, I thought I ran the tests locally, my bad.

PR #70, opened 2016-10-27
Allow reactive extensions to provide static members

sharplet:
It's useful to be able to attach reactive extensions to types, not just instances. This adds support for static members in Reactive extensions, adding a static `reactive` property to types conforming to `ReactiveExtensionsProvider`.

### DIFF
--- a/Sources/Reactive.swift
+++ b/Sources/Reactive.swift
@@ -11,6 +11,11 @@ extension ReactiveExtensionsProvider {
 	public var reactive: Reactive<Self> {
 		return Reactive(self)
 	}
+
+	/// A proxy which hosts static reactive extensions for the type of `self`.
+	public static var reactive: Reactive<Self>.Type {
+		return Reactive<Self>.self
+	}
 }
 
 // A `Reactive` proxy hosts reactive extensions to `Base`.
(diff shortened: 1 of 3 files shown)


Tests/ReactiveSwiftTests/ReactiveExtensionsSpec.swift
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	static var staticProperty: SignalProducer<String, NoError> {
+		return SignalProducer(value: Base.staticProperty)
+	}
+}
+
+final class ReactiveExtensionsSpec: QuickSpec {
+	override func spec() {
+		describe("ReactiveExtensions") {
```
NachoSoto: 👏

Sources/Reactive.swift
```diff
@@ -11,6 +11,11 @@ extension ReactiveExtensionsProvider {
 	public var reactive: Reactive<Self> {
 		return Reactive(self)
 	}
+
+	/// A proxy which hosts static reactive extensions for the type of `self`.
+	public static var reactive: Reactive<Self>.Type {
+		return Reactive<Self>.self
+	}
```
mdiep: Huh. I didn't think you could have a static and instance members with the same name. (Did that change in Swift at some point?)

Nicely done!

PR #71, opened 2016-10-28
Export Result as part of ReactiveSwift's public API

sharplet:
Makes working with Result-based APIs much simpler by avoiding the need to manually import Result when using ReactiveSwift.

Note: This only works on Apple platforms.

### DIFF
--- a/ReactiveSwift.playground/Pages/Sandbox.xcplaygroundpage/Contents.swift
+++ b/ReactiveSwift.playground/Pages/Sandbox.xcplaygroundpage/Contents.swift
@@ -12,9 +12,7 @@
  1. Choose `View > Show Debug Area`
  */
 
-import Result
 import ReactiveSwift
-import Foundation
 
 /*:
  ## Sandbox
--- a/ReactiveSwift.playground/Pages/Signal.xcplaygroundpage/Contents.swift
+++ b/ReactiveSwift.playground/Pages/Signal.xcplaygroundpage/Contents.swift
@@ -12,9 +12,7 @@
 1. Choose `View > Show Debug Area`
 */
 
-import Result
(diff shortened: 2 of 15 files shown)


andersio:
`@_exported` is a stdlib implementation detail so far, and was said to [break code completion](http://stackoverflow.com/questions/33558995/what-is-the-exported-attribute-in-swift).

Sources/Exports.swift
```diff
@@ -0,0 +1,10 @@
+//
+//  Result.swift
+//  ReactiveSwift
+//
+//  Created by Adam Sharp on 28/10/16.
+//  Copyright © 2016 GitHub. All rights reserved.
+//
+
+@_exported import Foundation
```
NachoSoto: This is a private API.

sharplet: Yeah I've seen other libraries use this and was experimenting. Reverted.

mdiep:
👎 on this since it only works on Apple platforms.

I don't want to make it easy for downstream libraries to accidentally break support for other platforms. This seems like it might be good for RAC though—it could export ReactiveSwift and Result this way.

sharplet:
I have a similar branch on RAC, but there's a big problem: these modular imports don't appear work for operator declarations. So if you only `import ReactiveCocoa` and then go to use `<~`, you'll still be confused as to why it won't compile, and you need to `import ReactiveSwift` anyway.

The only workaround I'm aware of for RAC would be to somehow include the source for the operator definitions, effectively duplicating it. Argo employed this strategy with Runes for a while, but recently reverted it due to issues with CocoaPods' support for submodules.

andersio:
We may explore the possibility of doing this in RAC's module map though.

PR #75, opened 2016-11-01
Added `--no-use-binaries` flag 

Iyuna:
Added `--no-use-binaries` flag to playground setup through Carthage documentation as without it Carthage installs Result binary instead of checking out the sources. In such case Result build scheme is not defined and cannot be built.


sharplet:
Thanks for contributing! I also just realised that the instructions mention running `script/bootstrap`, which doesn't actually exist anymore. I'll have to fix that up 😅

PR #76, opened 2016-11-02
Remove references to 'script/bootstrap' from playground documentation

sharplet:
Noticed this when reviewing #75.

This was removed in commit 923f07b1cd19bbf632f9fc9a9f078958168efe88.  It's referenced in the playground setup documentation and is useful, so restoring it.

### DIFF
--- a/ReactiveSwift.playground/Pages/Property.xcplaygroundpage/Contents.swift
+++ b/ReactiveSwift.playground/Pages/Property.xcplaygroundpage/Contents.swift
@@ -2,7 +2,7 @@
  > # IMPORTANT: To use `ReactiveSwift.playground`, please:
  
  1. Retrieve the project dependencies using one of the following terminal commands from the ReactiveSwift project root directory:
-    - `script/bootstrap`
+    - `git submodule update --init`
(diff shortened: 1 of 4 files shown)


mdiep:
Do you think this is better than instructing people to run `carthage bootstrap`? If so, why?

_(Looking to educate myself, not arguing either way)_

sharplet:
No I probably wouldn't say this is better or worse. I think if users have Carthage installed then that's great, but having an alternative for those who may not want to go install it makes sense to me.

ikesyo:
I prefer just using `git submodule update --init`, that does not require installing Carthage.

sharplet:
You're right, that might be simpler. Perhaps it would make more sense to update the instructions to recommend that.

dcaunt:
Yes, it would make sense to update the playground instructions to match the README, which changed in 745d6d498f8bda9a3ccb0dd896bd40f99f74407d

mdiep:
> Perhaps it would make more sense to update the instructions to recommend that.

Let's go that route. 👍 As a user, I generally prefer that to running an opaque script.

sharplet:
Removed all references to `script/bootstrap` in favour of `git submodule update --init`.

PR #68, opened 2016-10-21
Update Nimble to v5.1.0

ikesyo:
https://github.com/Quick/Nimble/releases/tag/v5.1.0

PR #80, opened 2016-11-04
Fixed a bug in `on` which caused the signal to be disposed of too early.

### LINKED ISSUE #77: Inner signal not "interrupted" on outer signal disposal
(Ref: https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2959)

Below gist (SignalProducerSpec's "disposal") passed the test in `ReactiveCocoa v4.2.2` but not in current `ReactiveSwift 1.0.0-alpha.3`.

https://gist.github.com/inamiy/925644779adbfca621ea716770d3865e

It seems `innerSignal.on().observe()` is disposed first before `on()` receives `.interrupted`.
This could be a NEW correct behavior as discussed in https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3075, but if I want to know the interruption of `innerSignal`, what should I do?

andersio:
Fixes #77.

#### The root cause

```
 		return SignalProducer { observer, compositeDisposable in
  			// ...
  			self.startWithSignal { signal, disposable in
  				compositeDisposable += disposable
 				compositeDisposable += signal.on(...).observe(observer)
```
The outer input observer was detached by the outer interruption, and this causes the started signal to dispose of itself (no observer + unreachable) under the new lifetime semantics. 

Moreover, since `CompositeDisposable` disposes its inner disposables in the reversed insertion order, the inner producer is interrupted after the said detachment and self-disposal happened.

So in the end, both together caused the injected side effect via `on` capturing no interrupted event. 

#### The fix

Generally speaking, the disposable of the outer observer to the produced inner signal should not be added to the outer producer's disposable. The termination is already propagated via `interrupted`. 

#### Note 
We might need to review the rest of the operators.

### DIFF
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -1291,7 +1291,7 @@ extension SignalProducerProtocol {
 
 			self.startWithSignal { signal, disposable in
 				compositeDisposable += disposable
-				compositeDisposable += signal
+				signal
 					.on(
 						event: event,
 						failed: failed,
--- a/Tests/ReactiveSwiftTests/SignalProducerSpec.swift
+++ b/Tests/ReactiveSwiftTests/SignalProducerSpec.swift
@@ -1419,6 +1419,8 @@ class SignalProducerSpec: QuickSpec {
 				var execute: ((FlattenStrategy) -> Void)!
 
 				var innerDisposable = SimpleDisposable()
+				var isInnerInterrupted = false
+				var isInnerDisposed = false
 				var interrupted = false
 
 				beforeEach {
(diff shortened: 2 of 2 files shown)


inamiy:
Thanks for catch!
It seems reversing `compositeDisposable += ...` also works well (so that `self` producer's `signal` is safely interrupted before `signal.on(...).observe(observer)`'s disposable is being disposed.

```swift
 		return SignalProducer { observer, compositeDisposable in
  			// ...
  			self.startWithSignal { signal, disposable in
 				compositeDisposable += signal.on(...).observe(observer)
  				compositeDisposable += disposable
```

andersio:
@inamiy It is redundant to detach from a terminated signal.

andersio:
@ikesyo Would you mind to have a look at why Linux PM test fails? It says `toEventually` is missing, but it seems the package config of Nimble is fine.

andersio:
Linux test error should be fixed by #81.

inamiy:
@andersio I see. Thanks for explanation. LGTM!

ikesyo:
@andersio `Carthage/Checkouts/Nimble` submodule would not be related to SwiftPM integration unfortunately. I will take a look at the failure.

andersio:
@ikesyo I found that 5.0.0 has wrapped the async matchers in `#if _runtime(ObjC)`, while 5.1.1 doesn't.

ikesyo:
@andersio That is the reason I checked out a specific commit in `.travis.yml`: https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/.travis.yml#L85. See #68 too (I'm waiting a new release of Quick).

PR #93, opened 2016-11-12
Method lifting.

andersio:
Related: https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3287

Introduce a Swift version of `rac_liftSelector` that can be used with any Swift or ObjC objects, provided that the class conforms to `LifetimeProvider`.

```
// ReactiveObjC
[self rac_liftSelector:@selector(updateWithData:) withSignal:dataSignal];

// ReactiveSwift
dataSignal.take(during: reactive.lifetime).observeValues { [weak self] in self?.update(with: $0) }

// Method Lifting
self.lift { $0.update }(dataSignal)
self.lift { $0.update(with:) }(dataSignal)

// Method Lifting with SR-0068*
self.lift(Self.update)(dataSignal)
self.lift(Self.update(with:))(dataSignal)
```

[SR-0068: Expanding Swift Self to class members and value types](https://github.com/apple/swift-evolution/blob/master/proposals/0068-universal-self.md)


NachoSoto:
This public API is lacking documentation. I'm still unclear about what this accomplishes exactly.

andersio:
It turns a method reference into a weakly reference one, as simple as that. Say you might wanna do `observeValues(self.update(_:_:))` but with `self` being weakly captured. In this case, you have to fallback to call it in full signature instead of partial application.

andersio:
Perhaps a direct counterpart to `rac_liftSelector` is a better take in expressivity, hmm.
@NachoSoto

mdiep:
I don't think this makes a lot of sense to include. But TBH I always had mixed feelings about the Objective-C version.

sharplet:
I made a similar proposal with https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2880. I think the majority of use cases I had for the feature are covered by simple binding targets.

This could still be a really useful feature, but I don't think it's necessary for 1.0.

PR #101, opened 2016-11-20
Feature/property documentation

### LINKED ISSUE #6: Add Property's Page @ Playground

iv-mexx:
This resolves #6 

I tried to stick to stick to the existing playground documentation for Signal and SignalProducer.

I've also updated the Properties part in `FrameworkOverview.md`. 
This document is duplicated in [ReactiveCocoa](https://github.com/ReactiveCocoa/ReactiveCocoa), either it should be updated there as well, or the documentation should be de-duplicated between ReactiveCocoa and ReactiveSwift?

### DIFF
--- a/Documentation/FrameworkOverview.md
+++ b/Documentation/FrameworkOverview.md
@@ -136,15 +136,15 @@ For interaction with `NSControl` or `UIControl`, RAC provides the
 
 ## Properties
 
-A **property**, represented by the [`PropertyType`][Property] protocol,
+A **property**, represented by the [`PropertyProtocol`][Property],
 stores a value and notifies observers about future changes to that value.
 
 The current value of a property can be obtained from the `value` getter. The
 `producer` getter returns a [signal producer](#signal-producers) that will send
-the property’s current value, followed by all changes over time.
+the property’s current value, followed by all changes over time. The `signal` getter returns a [signal](#signals) that will send all changes over time, but not the initial value.
 
 The `<~` operator can be used to bind properties in different ways. Note that in
-all cases, the target has to be a [`MutablePropertyType`][Property].
+all cases, the target has to be a [`BindingTarget`][BindingTarget], [`MutableProperty`][MutableProperty] is the only property that implements this.
 
 * `property <~ signal` binds a [signal](#signals) to the property, updating the
   property’s value to the latest value sent by the signal.
@@ -157,8 +157,13 @@ The [`DynamicProperty`][Property] type can be used to bridge to Objective-C APIs
 that require Key-Value Coding (KVC) or Key-Value Observing (KVO), like
 `NSOperation`. Note that most AppKit and UIKit properties do _not_ support KVO,
 so their changes should be observed through other mechanisms.
+
+[ReactiveCocoa][ReactiveCocoa] implements a number of extensions on AppKit and UIKit to allow observation of and binding to properties via the `.reactive` structure.
+
 [`MutableProperty`][Property] should be preferred over dynamic properties
 whenever possible!
+
+Properties provide a number of transformations like `map`, `combineLatest` or `zip` for manipulation similar to [signal](#signals) and [signal producer](#signal-producers)
 
 ## Disposables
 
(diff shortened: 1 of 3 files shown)


andersio:
It seems natural to host them in ReactiveSwift. Perhaps we could turn them into links in ReactiveCocoa, hmm?

Documentation/FrameworkOverview.md
```diff
@@ -136,15 +136,15 @@ For interaction with `NSControl` or `UIControl`, RAC provides the
...
 stores a value and notifies observers about future changes to that value.
 
 The current value of a property can be obtained from the `value` getter. The
 `producer` getter returns a [signal producer](#signal-producers) that will send
-the property’s current value, followed by all changes over time.
+the property’s current value, followed by all changes over time. The `signal` getter returns a [signal](#signals) that will send all changes over time, but not the initial value.
 
 The `<~` operator can be used to bind properties in different ways. Note that in
-all cases, the target has to be a [`MutablePropertyType`][Property].
+all cases, the target has to be a [`BindingTarget`][BindingTarget], [`MutableProperty`][MutableProperty] is the only property that implements this.
```
andersio: `BindingTarget` is a concrete type. Maybe 

> binding target, represented by `BindingTargetProtocol`

instead?

Moreover, `MutableProperty` is not the only property that can be a target, since you can have any arbitrary number of mutable property types. I'd prefer to say:

> All mutable property types are inherently a binding target.

iv-mexx: 👍

iv-mexx:
> It seems natural to host them in ReactiveSwift. Perhaps we could turn them into links in ReactiveCocoa, hmm?

I agree that the documentation for things implemented in ReactiveSwift should be hosted in ReactiveSwift. 
On that note, I've just noticed that DynamicProperty is not in ReactiveSwift but in ReactiveCocoa, so I removed the mention from the FrameworkOverview.

Still not sure if the Documentation in ReactiveCocoa should just link back to ReactiveSwift or be complete, maybe I'll just try both and see how it feels...

Documentation/FrameworkOverview.md
```diff
@@ -153,12 +153,9 @@ all cases, the target has to be a [`MutablePropertyType`][Property].
 * `property <~ otherProperty` binds one property to another, so that the destination
   property’s value is updated whenever the source property is updated.
 
-The [`DynamicProperty`][Property] type can be used to bridge to Objective-C APIs
-that require Key-Value Coding (KVC) or Key-Value Observing (KVO), like
-`NSOperation`. Note that most AppKit and UIKit properties do _not_ support KVO,
-so their changes should be observed through other mechanisms.
-[`MutableProperty`][Property] should be preferred over dynamic properties
-whenever possible!
+[ReactiveCocoa][ReactiveCocoa] implements a number of extensions on AppKit and UIKit to allow observation of and binding to properties via the `.reactive` structure.
```
mdiep: I think we can remove this mention of ReactiveCocoa.

ReactiveSwift.playground/Pages/Property.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,290 @@
...
+    }
+    
+    print("Setting new value for property: 1")
+    property.value = 1
+    
+    print("Setting new value for property: 2")
+    property.value = 2
+}
+
+scopedExample("`skipRepeats`") {
```
mdiep: This appears to be a duplication of the more-complete example below.

mdiep:
Thanks for the PR!

I think we can focus on ReactiveSwift here, without mentions of ReactiveCocoa. ReactiveCocoa doesn't need to duplicate the information here, but could link to it. Adding a page to RAC that explains `DynamicProperty` would probably be a good idea.

ReactiveSwift.playground/Pages/Property.xcplaygroundpage/Contents.swift
iv-mexx: Oh right, I created the more complex example for `uniqueValues` first and then decided to use the same sequence for `skipRepeats` as well to better illustrate the difference. 
I'll clean up the simple example.

Documentation/FrameworkOverview.md
iv-mexx: 👍

iv-mexx:
Alright, @mdiep I've addressed your comments

mdiep:
Thanks @iv-mexx! This is great. ✨

iv-mexx:
Thanks! I'll probably look into the documentation some more soon

PR #104, opened 2016-11-22
SignalProducer.on: restore `value` as the last parameter

dmcrodrigues:
Conceptually, `value` should be defined before `failed` since that goes in line with the natural order of events, however [this reordering](https://github.com/ReactiveCocoa/ReactiveCocoa/commit/85e2359d931c978b751940e066bcd95f0b215dd6) can introduce some confusion/inconvenience while using trailing closures specially during the migration from ReactiveCocoa 4.x where `value`, formerly known as `next`, was the last parameter and now represents `disposed`.

This can silently introduce some problems after migration if a project is using `.on` with a trailing closure to inject side-effects on values without being interested in the value itself.

```swift
producer.on { _ in 
    // Inject side-effect, e.g. invalidate context
}
```

## ⚠️ Note ⚠️
If we do not go forward with this we need to fix an inconsistency between `SignalProducer` and `Signal` where the former was not updated accordingly.

### DIFF
--- a/Sources/EventLogger.swift
+++ b/Sources/EventLogger.swift
@@ -102,12 +102,12 @@ extension SignalProducerProtocol {
 		return self.on(
 			starting: log(.starting),
 			started: log(.started),
-			value: log(.value),
 			failed: log(.failed),
 			completed: log(.completed),
 			interrupted: log(.interrupted),
 			terminated: log(.terminated),
-			disposed: log(.disposed)
+			disposed: log(.disposed),
+			value: log(.value)
 		)
 	}
 }
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -815,13 +815,13 @@ extension SignalProtocol {
 	/// - parameters:
 	///   - event: A closure that accepts an event and is invoked on every
 	///            received event.
-	///   - value: A closure that accepts a value from `value` event.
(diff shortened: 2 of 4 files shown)


dmcrodrigues:
@NachoSoto this was discussed on Slack today and I opened a PR to discuss what should we do.

andersio:
If we prefer the natural order, adding `_: Void = ()` as the last parameter would prevent the use of trailing closure.

dmcrodrigues:
I thought about that and it's definitely another option but with some quirks.

>Closure parameter prior to parameters with default arguments will not be treated as a trailing closure

A warning is raised as expected but can we ignore it?

screen shot 2016-11-23 at 08 48 41

Despite of having a default value, Xcode will ask us a value for our (unnamed) last parameter which I think is far from desirable and can introduce some confusion about the why and what it really means.

andersio:
Another alternative is being in line with `observe` and `start`, so that `event` would be the trailing closure. But it won't solve the migration issue anyway.

andersio:
I don't feel it is frequently used enough to justify shorthands like how `start` or `observe` does. So leaving it as it was in 4.x seems fine. Although it does bugs me a bit since `start` and `observe` take an event action, instead of a value action.

PR #106, opened 2016-11-23
Refactored & Bugfix: `startWithSignal` and the hot-to-cold `lift`.

andersio:
### `startWithSignal `
One less layer of wrapping = more joules saved.

#### ~~Breaking Change~~ Bugfix
If the signal created & passed to the `setUp` closure is neither retained nor observed, the producer disposable would be disposed of.

Detailed Example: https://github.com/ReactiveCocoa/ReactiveSwift/pull/106#issuecomment-263079737

Due to the new signal lifetime semantics in 1.0, the signal would have been disposed of upon exiting the scope of the closure. But the producer disposable does not follow the disposal of the produced signal - it is only disposed of until the upstream sends a termination event, or the producer is interrupted by the user.

In other words, it doesn't match the expected behaviour of `SignalProducer` having downstream disposals propagated back to the upstream.

This fix should IMO hardly affect any existing use case though, since the closure based `start`s are way more popular.

### `lift` that returns `(Signal) -> SignalProducer`
The signal is lifted as a producer (so it becomes interruptible) before applying the transform.

### DIFF
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -176,11 +176,7 @@ public struct SignalProducer<Value, Error: Swift.Error> {
 		// upstream producers.
 		let producerDisposable = CompositeDisposable()
 
-		var observer: Signal<Value, Error>.Observer!
-		let signal = Signal<Value, Error> { innerObserver in
-			observer = innerObserver
-			return producerDisposable
-		}
+		let (signal, observer) = Signal<Value, Error>.pipe(disposable: producerDisposable)
 
 		// Directly disposed of when `start()` or `startWithSignal()` is
 		// disposed.
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -172,18 +172,19 @@ public struct SignalProducer<Value, Error: Swift.Error> {
 	/// - parameters:
 	///   - setUp: A closure that accepts a `signal` and `interrupter`.
 	public func startWithSignal(_ setup: (_ signal: Signal<Value, Error>, _ interrupter: Disposable) -> Void) {
-		let (signal, observer) = Signal<Value, Error>.pipe()
-
 		// Disposes of the work associated with the SignalProducer and any
 		// upstream producers.
 		let producerDisposable = CompositeDisposable()
 
+		var observer: Signal<Value, Error>.Observer!
+		let signal = Signal<Value, Error> { innerObserver in
+			observer = innerObserver
+			return producerDisposable
+		}
+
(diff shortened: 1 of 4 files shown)


NachoSoto:
Any chance you can add a test case for the actual hug you fixed? 🙏

andersio:
Added in https://github.com/ReactiveCocoa/ReactiveSwift/pull/106/commits/f6266580970e68837a40c342d09e5c59dc777ff1. Though you might have also seen a bunch of test cases that failed and needed a change because of the fix.

Sources/SignalProducer.swift
```diff
@@ -462,27 +448,9 @@ extension SignalProducerProtocol {
...
-					otherSignalObserver.sendInterrupted()
-				}
-				outerDisposable += otherSignal.observe(otherSignalObserver)
-
-				self.startWithSignal { signal, disposable in
-					outerDisposable += disposable
-					outerDisposable += transform(signal)(wrapperSignal).observe(observer)
-				}
-			}
+			return self.liftRight(transform)(SignalProducer(signal: otherSignal))
```
NachoSoto: Oh this is great

Tests/ReactiveSwiftTests/SignalProducerSpec.swift
```diff
@@ -1970,6 +1988,7 @@ class SignalProducerSpec: QuickSpec {
 				producer
 					.observe(on: TestScheduler())
 					.startWithSignal { signal, innerDisposable in
+						signal.observe { _ in }
```
NachoSoto: What's this for?

andersio: Under the new signal lifetime semantic, a signal is kept alive only if it is retained or has active observers. So the observer is added to keep the signal alive, or otherwise the signal would dispose of itself upon exiting the scope.

The bug fixed is kinda related too - a produced signal would dispose of itself when it is not retained and not observed. But then its associated producer disposable would not follow.

andersio:
```swift
// RAC 4.0: The signal is alive until a termination event or a user interruption.
// RAS 1.0: The signal deinitializes upon exiting the closure scope.
producer.startWithSignal { _ in }

// `master`: The produced signal deinitializes, but the associated composite
//           disposable does not.
// Bugfix: The signal is asked to dispose of also the composite disposable.
let d = SimpleDisposable()
SignalProducer { _, pd in pd += d }.startWithSignal { _ in }

// `Signal` is retained. Nothing has changed.
var signal: Signal!
producer.startWithSignal { signal = $0 }
```

NachoSoto:
Awesome

PR #107, opened 2016-11-23
Lock-free disposal.

andersio:
`Disposable` is crucial part of the signal resource management, especially in transformed signals and `SignalProducer`s.

Previous attempt: https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2463

~~This PR focuses on just `Disposable`. All disposables have a `DisposableState`, which exposes accessors built on top of OSAtomic primitives. `DisposableState` is made a `struct` so that the state is embedded into the containing class without extra indirection.~~

This PR introduces a lock-free primitive `UnsafeAtomicState`, and refactored all disposable types to use it. Note that the primitive is marked with `Unsafe` as it requires manual memory management.

Note that while `CompositeDisposable` and `SerialDisposable` still use `Atomic` for storing their children, they may still benefit from the fast atomic boolean checks.

The result is pretty impressive. For 1000000 iterations of `disposable.dispose()` (just a CAS) with whole module optimization, lock-free `SimpleDisposable` completed in just 0.011 seconds (12% SD), while `Atomic<Bool>` took 0.233 sec (4% SD).

Less joules. 😸 

--

~~This includes also a 1.5-2x faster `Bag.forEach` overload under `-Owholemodule` and `-Ounchecked`.~~

### DIFF
--- a/Sources/Atomic.swift
+++ b/Sources/Atomic.swift
@@ -24,7 +24,7 @@ internal protocol AtomicStateProtocol {
 	///
 	/// - returns:
 	///   `true` if the transition succeeds. `false` otherwise.
-	mutating func tryTransiting(from expected: State, to next: State) -> Bool
+	func tryTransiting(from expected: State, to next: State) -> Bool
 }
 
 /// A simple, generic lock-free finite state machine.
--- a/Sources/Atomic.swift
+++ b/Sources/Atomic.swift
@@ -7,6 +7,123 @@
 //
 
 import Foundation
+#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
+import MachO
+#endif
+
+/// Represents a finite state machine that can transit from one state to
+/// another.
+internal protocol AtomicStateProtocol {
+	associatedtype State: RawRepresentable
+
+	/// Try to transit from the expected current state to the specified next
+	/// state.
+	///
+	/// - parameters:
+	///   - expected: The expected state.
+	///
+	/// - returns:
+	///   `true` if the transition succeeds. `false` otherwise.
+	mutating func tryTransiting(from expected: State, to next: State) -> Bool
+}
+
+/// A simple, generic lock-free finite state machine.
+///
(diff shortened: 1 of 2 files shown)


andersio:
[DisposablePerfTests.swift](https://gist.github.com/andersio/8f3086a66b9ff5e3a84270218c677765)

| Item | -Onone | -Owholemodule |
| -----| ------- | --------- |
| Lock-free | 0.017 sec (4%) | 0.013 (5%) |
| `Atomic` | 0.267 sec (4%) | 0.234 (2%) |

Environment: 2.4 GHz Intel Core i5-4258U, macOS 10.12.1, Xcode 8.1

andersio:
I've factored out the lock-free logic, and turned it into a generic FSM-like container*. It should be at least reusable in `Signal` for handling interruption. :)

It is a little bit faster too, with the inlining annotation in place. With [the same test and configuration](https://github.com/ReactiveCocoa/ReactiveSwift/pull/107#issuecomment-262679244), it completes in 0.010 sec (10%).

_\* Transitions are represented as protocol extensions. Tried tuples and type parameter, but both have a nontrivial cost at runtime._

andersio:
Comparison across branches. The test case is sort of simulating a fast scrolling table view, which gets a bunch of properties rebinding agains the cells.

`master`: 1.164 sec (3%)
#124: 1.081 sec (2%)
This PR: 0.995 sec (2%)

```swift
	func testProducerStart() {
		let property = MutableProperty(1).map { $0 + 1 }.map { $0 + 1 }.map { $0 + 1 }
		let token = Lifetime.Token()
		let lifetime = Lifetime(token)

		var i = 0
		let target = BindingTarget(lifetime: lifetime, setter: { i = $0 })

		measure {
			for i in 0 ..< 10000 {
				let d = target <~ property
				d.dispose()
			}
		}
	}
```

andersio:
It seems nested `@inline(__always)` declarations across files would fail. Not a problem here though, unless we mark `dispose` as inlined.

NachoSoto:
Awesome work!!

PR #108, opened 2016-11-23
Fixed race conditions in the Signal interruption handling.

andersio:
~~One less thing to have its own lock, and~~

Let `Signal` use a faster send lock on Darwin platforms.

------

It seems there is a race condition in the signal interruption handling that would result in the interruption being indefinitely delayed until the next event arrives.

```
// Race condition if we perform the interruption check within the
// protected section, specifically:
//
// 1. The `interrupted` event sets the atomic flag to true, but after
//    the send lock owner has checked the atomic flag.
//
// 2. The `interrupted` event proceeds to try to lock the `sendLock`,
//    but failed since the lock has not yet been released.
//
// 3. The `sendLock` is released. But since the `interrupted` event
//    does no attempt to retry, the interruption would be indefinitely
//    delayed until the next event arrives.
```

The original design might also let events slip through after the `interrupted` event, which violates the Signal contract.

The PR addressed these by making the interruption status now a tri-state value, and use it to grant only one pending (post-interruption) event to pass.

### DIFF
--- a/Sources/Atomic.swift
+++ b/Sources/Atomic.swift
@@ -8,7 +8,7 @@
 
 import Foundation
 
-final class PosixThreadMutex: NSLocking {
+internal final class PosixThreadMutex: NSLocking {
 	private var mutex = pthread_mutex_t()
 
 	init() {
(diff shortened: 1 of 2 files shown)


andersio:
Hmm, the change delayed the interruption.

andersio:
_(Feels like I've jumped into a huge whirl...)_

andersio:
It seems there is a race condition in the signal interruption handling that would result in the interruption being indefinitely delayed until the next event arrives.

Issue #115, opened 2016-11-24
Odd behaviour on `on`


gkaimakas:
I have the following snippet of code:

````Swift
	SignalProducer<Int, NSError>(value: 15)
			.on(starting: {
				print("starting")
			})
			.on(started: {
				print("started")
			})
			.on(value: { _ in
				print("value")
			})
			.on(failed: { _ in
				print("failed")
			})
			.on(completed: {
				print("completed")
			})
			.start()
````

That prints the following results on the console

````
starting
value
completed
started
````

Shouldn't the output be 
````
starting
started
value
completed
````

instead?

The same thing happens when the `producer` fails.

mdiep:
This is because of how `SignalProducer.init(value:)` works. It sends the value _immediately_.

It goes something like this.

1. Call `starting`
2. Call `SignalProducer` to start
    1. `init(value:)` sends a value (`value`) and then completes (`completed`)
3. Call `started`

gkaimakas:
@mdiep This is quite confusing. I have observed the same behaviour on `MutableProperty`

andersio:
`started` is invoked after the starting side effect is done. So you would exhibit the same behaviour in any producer that sends events synchronously in the start hamdler.

mdiep:
I think it'd be more surprising for `started` to not always fire.

gkaimakas:
Shouldn't `started` fire after `starting`?

andersio:
No.

> `started` is invoked after the starting side effect is done.

Look at the signature of `SignalProducer.init(_:)`. It accepts a `startHandler` that is passed an observer. This means events can be sent as part of the starting side effect, and `started` indicates the side effect is completed, following these events.

mdiep:
Maybe this will help you make sense of it: the signal completed before it finished starting.

PR #116, opened 2016-11-24
Update Basic Operators for ReactiveSwift 1.0

iv-mexx:
All the examples have been update so that they work again (e.g. in the Playground)

Most changes were straightforward, like the renaming of methods and events.

However, the `.flatten` examples are a bit more complicated. Formerly, they relied on `SignalProducer.buffer` [which has been removed](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/CHANGELOG.md#signalproducer-buffer-has-been-removed). 

At first I wanted to preserve the same example behavior, first try was to use `MutableProperty` instead, but the examples need `completion` of the producer.
Next try was to use `SignalProducer.init(signal:).replyLazily(upto:)` but then it was necessary to additionally start these producers prior to the actual example (to achieve the same behavior as before).

Finally, I've simplified the examples by using `Signal`, but this does not show exactly the same behavior as before. Most notably, there are now values sent when
* `lettersSignal` completes in the `.concat` example
* `producerB`, `producerC` are sent on `producer` in the (old) `.latest` example.

### DIFF
--- a/Documentation/BasicOperators.md
+++ b/Documentation/BasicOperators.md
@@ -305,25 +305,17 @@ Note, how the values interleave and which values are even included in the result
 
 ### Merging
 
-The `.merge` strategy immediately forwards every value of the inner `SignalProducer`s to the outer `SignalProducer`. Any failure sent on the outer producer or any inner producer is immediately sent on the flattened producer and terminates it.
+The `.merge` strategy immediately forwards every value of the inner `Signal`s to the outer `Signal`. Any failure sent on the outer signal or any inner signal is immediately sent on the flattened signal and terminates it.
 
 ```Swift
 let (lettersSignal, lettersObserver) = Signal<String, NoError>.pipe()
 let (numbersSignal, numbersObserver) = Signal<String, NoError>.pipe()
+let (signal, observer) = Signal<Signal<String, NoError>, NoError>.pipe()
 
-let lettersProducer = SignalProducer(signal: lettersSignal).replayLazily(upTo: 3)
-let numbersProducer = SignalProducer(signal: numbersSignal).replayLazily(upTo: 3)
+signal.flatten(.merge).observeValues { print($0) }
 
-lettersProducer.start()
-numbersProducer.start()
-
-let (signal, observer) = Signal<SignalProducer<String, NoError>, NoError>.pipe()
-let producer = SignalProducer(signal: signal)
-
-producer.flatten(.merge).startWithValues { print($0) }
-
-observer.send(value: lettersProducer)
-observer.send(value: numbersProducer)
+observer.send(value: lettersSignal)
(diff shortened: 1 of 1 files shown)


andersio:
Should we group the producer specific operators together?

iv-mexx:
Hm, which operators do you have in mind?

Appart from that, I've just seen seen this line from the introduction again: 
> This document will use the term “event stream” when dealing with concepts that
apply to both `Signal` and `SignalProducer`.

It seems that this is not consistently followed in the document, I'll take another look at it.

iv-mexx:
> Should we group the producer specific operators together?

Now that "event stream" is used for every operator that applies to `Signal` as well as `SignalProducer`, only `retry` is left that is specific to `SignalProducer`.

andersio:
Oh, never mind. It seems only `retry` falls into this category.

Documentation/BasicOperators.md
```diff
@@ -49,68 +49,72 @@ types will be referred to by name.
...
-    }, interrupted: {
+    }, interrupted: { 
         print("Interrupted")
-    }, terminated: {
+    }, terminated: { 
         print("Terminated")
-    }, disposed: {
+    }, disposed: { 
         print("Disposed")
-    }, next: { value in
```
andersio: This is not true with `starting`. 😸

iv-mexx: You are talking about this line: 
> Note that nothing will be printed until `producer` is started (possibly somewhere else).

right?

Sorry, I'm not sure what you mean exactly. I've tried this example in a Playground
```
let signalProducer = SignalProducer<Int, NoError>(value: 1)

let producer = signalProducer
    .on(starting: {
        print("Starting")
    }, started: {
        print("Started")
    }, event: { event in
        print("Event: \(event)")
    }, value: { value in
        print("Value: \(value)")
    }, failed: { error in
        print("Failed: \(error)")
    }, completed: {
        print("Completed")
    }, interrupted: {
        print("Interrupted")
    }, terminated: {
        print("Terminated")
    }, disposed: {
        print("Disposed")
    })

print("---")
producer.start()
print("---")
```

And it behaved as expected and as the line says ("nothing is printed until the producer is started"):

```
---
Starting
Event: VALUE 1
Value: 1
Event: COMPLETED
Completed
Terminated
Disposed
Started
---
```

andersio: We can have outputs printed before the producer has started. But no output is ever printed until the producer is started, yeah. Hmm, it should be fine, just that it _might_ be a confusing point for beginners.

iv-mexx: Mhm, are you referring to #115 ?

andersio: Kind of.

PR #117, opened 2016-11-24
Remove Property source capturing.

andersio:
The PR proposes to have composed properties *not* retaining its sources. [This test case](https://github.com/ReactiveCocoa/ReactiveSwift/pull/117/files#diff-8719ae7a50d7fc426e03c8da6c209061R411) helps illustrate the new behaviour.

Recall that retaining `Property` would defer the termination of the event stream. Source capturing retains the sources, causing the sources' lifetime being a union with the lifetime of all their derivatives.

However, if we consider properties are just streams of values with a stronger guarantee than `Signal`s (having one latest value), operators - that are fundamentally an observation - should not have a side effect on the observed event stream. In this sense, source capturing should be dropped.

~~It also adds a new initializer `Property(reflecting:)`, which does the same thing as `Property(_:)` but without retaining the supplied property.~~

The default initializer `Property(_:)` is changed not to capture the supplied property, while the capturing "existential" variant is renamed to `Property(capturing:)`. The renaming makes a better migration path when we deprecate the initializer as [generalised existential](https://github.com/apple/swift/blob/master/docs/GenericsManifesto.md#generalized-existentials) lands in Swift.

P.S. Existential properties serves to work around the current language limits, so they should be left untouched.

#### Background

As brought up in https://github.com/ReactiveCocoa/ReactiveSwift/pull/58#discussion_r89527936, I have made a review on why we had composed properties capturing its source.

It was originated in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2788:

> Chain constructs like `let p1 = p0.map { $0.location }.map { $0.latitude }` are not safe with the proposed implementation, unless you explicitly capture `p0` for the whole `p1`'s lifecycle. 

It was true at that time, since the [property composition operators](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2922) wasn't a thing yet, and the composition is done through `AnyProperty` which at that time was based on `MutableProperty`. So had we not made the property captured, **chaining** of `map` would not behave as expected.

But now with the [property composition formally introduced](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2922), we propagate the values using [replayed producer](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3078), which works properly when chained, rendering the need of capturing moot.

### DIFF
--- a/Sources/Property.swift
+++ b/Sources/Property.swift
@@ -359,28 +359,22 @@ extension PropertyProtocol {
 	}
 }
 
-/// A read-only property that can be observed for its changes over time. There are
-/// three categories of read-only property:
+/// A read-only property that can be observed for its changes over time. There
+/// are two categories of read-only properties:
 ///
 /// # Constant property
 /// Created by `Property(value:)`, the producer and signal of a constant
 /// property would complete immediately when it is initialized.
 ///
-/// # Existential property
-/// Created by `Property(_:)`, an existential property passes through the
-/// behavior of the wrapped property.
-///
 /// # Composed property
-/// Created by either the compositional operators in `PropertyProtocol`, or
-/// `Property(initial:followingBy:)`, a composed property presents a
-/// composed view of its source, which can be a set of properties,
-/// a producer, or a signal.
+/// A composed property presents a composed view of its sources, which can be
+/// one or more properties, a producer, or a signal.
 ///
-/// A composed property respects the lifetime of its source rather than its own.
-/// In other words, its producer and signal can outlive the property itself, if
-/// its source outlives it too.
+/// It respects and have no effect on the lifetime of its root sources. In other
+/// words, the producer and signal of a composed property could complete before
+/// or outlive the composed property, depending on its sources and the
+/// composition.
 public final class Property<Value>: PropertyProtocol {
-	private let sources: [AnyObject]
 	private let disposable: Disposable?
 
 	private let _value: () -> Value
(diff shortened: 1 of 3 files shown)


Sources/Property.swift
```diff
@@ -419,16 +420,27 @@ public final class Property<Value>: PropertyProtocol {
...
 		_signal = { property.signal }
 	}
 
+	/// Initializes a composed property which reflects the given property.
+	///
+	/// - note: The resulting property does not retain the given property.
+	///
+	/// - parameters:
+	///   - property: A property to be wrapped.
+	public convenience init<P: PropertyProtocol>(reflecting property: P) where P.Value == Value {
```
sharplet: Just wanted to note that `reflecting` doesn't feel quite right to me, though I haven't been able to come up with a suitable alternative yet.

sharplet:
Ok, so the main use I have for the existential initialiser is this pattern:

```swift
private let _foo: MutableProperty<Foo>

var foo: Property<Foo> {
  return Property(_foo)
}
```

I kind of feel like this initialiser should _not_ retain the underlying property by default. Then, to opt into capturing semantics, you could have:

```swift
return Property(capturing: _foo)
```

What do you think?

andersio:
@sharplet Yeah, that does seem better. It gives a better default when generalised existentials eventually land in Swift, and we have the existential initializer deprecated.

andersio:
Should we mark `AnyProperty(_:)` renamed to `Property(capturing:)` instead? Or mark it as deprecated and tell the user there are two variants (no auto-fixit) instead?

@ReactiveCocoa/reactiveswift

sharplet:
I've seen some fixits  that provide multiple alternatives (unfortunately can't recall which, specifically) — maybe that's worth exploring?

andersio:
Hmm... the unavailable annotation of `AnyProperty` the type has a higher precedence than `Property`... So annotating `init` does not translate into anything.

andersio:
We probably can't do any better. So perhaps we should just let it be. 🙈

Issue #118, opened 2016-11-24
Rename `attempt*` and `flatMapError` back to `try*` and `catch`.


andersio:
With [SE-0071](https://github.com/apple/swift-evolution/blob/master/proposals/0071-member-keywords.md) in place, `catch` and `try` can be referenced as members without backticks.

Perhaps `times` can also be renamed as `repeat` too.

PR #119, opened 2016-11-24
Rename `times` to `repeat`.

andersio:
Related issue: #118

### DIFF
--- a/Sources/Deprecations+Removals.swift
+++ b/Sources/Deprecations+Removals.swift
@@ -1,5 +1,5 @@
 import Foundation
-import enum Result.NoError
+import Result
 
 // MARK: Removed Types and APIs in ReactiveCocoa 5.0.
 
@@ -279,6 +279,9 @@ extension SignalProducerProtocol {
 
 	@available(*, unavailable, message:"This SignalProducer may emit errors which must be handled explicitly, or observed using `startWithResult(_:)`.")
 	public func startWithNext(_ next: (Value) -> Void) -> Disposable { fatalError() }
+
+	@available(*, unavailable, renamed:"repeat(_:)")
+	public func times(_ count: Int) -> SignalProducer<Value, Error> { fatalError() }
 }
 
 extension SignalProducerProtocol where Value: OptionalProtocol {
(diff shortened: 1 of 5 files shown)


Sources/Deprecations+Removals.swift
```diff
@@ -279,6 +288,18 @@ extension SignalProducerProtocol {
 
 	@available(*, unavailable, message:"This SignalProducer may emit errors which must be handled explicitly, or observed using `startWithResult(_:)`.")
 	public func startWithNext(_ next: (Value) -> Void) -> Disposable { fatalError() }
+
+	@available(*, unavailable, renamed:"repeat(_:)")
```
NachoSoto: I agree `repeat` is better.

Sources/Deprecations+Removals.swift
```diff
@@ -207,6 +207,15 @@ extension SignalProtocol {
 
 	@available(*, unavailable, message: "This Signal may emit errors which must be handled explicitly, or observed using `observeResult(_:)`")
 	public func observeNext(_ next: (Value) -> Void) -> Disposable? { fatalError() }
+
+	@available(*, unavailable, renamed:"catch(_:)")
+	public func flatMapError<F>(_ handler: @escaping (Error) -> SignalProducer<Value, F>) -> Signal<Value, F> { fatalError() }
```
NachoSoto: `flatMapError` is far more descriptive. `catch` is a well known word but it doesn't carry the extra information that this does, which draws a parallel with `flatMap`.

Sources/Deprecations+Removals.swift
```diff
@@ -207,6 +207,15 @@ extension SignalProtocol {
...
+
+	@available(*, unavailable, renamed:"try(_:)")
+	public func attempt(_ operation: @escaping (Value) -> Result<(), Error>) -> Signal<Value, Error> { fatalError() }
+
+	@available(*, unavailable, renamed:"tryMap(_:)")
+	public func attemptMap<U>(_ operation: @escaping (Value) -> Result<U, Error>) -> Signal<U, Error> { fatalError() }
```
NachoSoto: These names are meant to match the `Result` counterparts.

Sources/Deprecations+Removals.swift
```diff
@@ -1,5 +1,5 @@
 import Foundation
-import enum Result.NoError
+import Result
```
ikesyo: Should this be reverted too?

PR #120, opened 2016-11-24
Drop the label of `SignalProducer` initializers that take a sequence.

andersio:
Since this is not a lossy conversion, and there is no conflicting signature, the `signal` label does not give any extra clarity at call site.

### DIFF
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -29,7 +29,7 @@ public struct SignalProducer<Value, Error: Swift.Error> {
 	///
 	/// - parameters:
 	///   - signal: A signal to observe after starting the producer.
-	public init<S: SignalProtocol>(signal: S) where S.Value == Value, S.Error == Error {
+	public init<S: SignalProtocol>(_ signal: S) where S.Value == Value, S.Error == Error {
 		self.init { observer, disposable in
 			disposable += signal.observe(observer)
 		}
@@ -100,9 +100,9 @@ public struct SignalProducer<Value, Error: Swift.Error> {
 	/// - parameters:
 	///   - values: A sequence of values that a `Signal` will send as separate
 	///             `value` events and then complete.
-	public init<S: Sequence>(values: S) where S.Iterator.Element == Value {
+	public init<S: Sequence>(_ sequence: S) where S.Iterator.Element == Value {
 		self.init { observer, disposable in
-			for value in values {
+			for value in sequence {
 				observer.send(value: value)
 
 				if disposable.isDisposed {
@@ -122,7 +122,7 @@ public struct SignalProducer<Value, Error: Swift.Error> {
 	///   - second: Second value for the `Signal` to send.
 	///   - tail: Rest of the values to be sent by the `Signal`.
 	public init(values first: Value, _ second: Value, _ tail: Value...) {
-		self.init(values: [ first, second ] + tail)
+		self.init([ first, second ] + tail)
 	}
 
 	/// A producer for a Signal that will immediately complete without sending
@@ -448,7 +448,7 @@ extension SignalProducerProtocol {
 	///            `SignalProducer`.
 	public func lift<U, F, V, G>(_ transform: @escaping (Signal<Value, Error>) -> (Signal<U, F>) -> Signal<V, G>) -> (Signal<U, F>) -> SignalProducer<V, G> {
 		return { otherSignal in
-			return self.liftRight(transform)(SignalProducer(signal: otherSignal))
+			return self.liftRight(transform)(SignalProducer(otherSignal))
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


andersio:
I'd argue that the variable names usually have conveyed that they are signals, e.g. ```SignalProducer(signal: usernameSignal)```, ```SignalProducer(signal: occupationValues)```, which makes the `signal` label look redundant. Labeled initializer is also used only when it comes to type narrowing, although in our case we have to prevent conflict with `values: ...` and friends.

sharplet:
Most of the time I use this initialiser to compose a signal with other producers, or to use producer-specific operators on it. This means that it's usually part of some kind of operator chain, where a mess of keywords, operators and other syntax can make things dense and hard to parse. I personally find that the `signal:` label doesn't add much other than redundancy, which for me _reduces_ clarity / increases visual noise.

sharplet:
But, it would be good to show some real examples of how this changes code at the point of use.

andersio:
TBH I haven't used it in my projects with all the stuff we have introduced since 4.2.0. The last time I've ever used it is probably in our codebase. As you might have seen in the diff, the label doesn't feel like adding clarity at all but verbosity. Though maybe if we rename the label to `forwarding` it would make a better sense.

```
 	public convenience init(initial: Value, then signal: Signal<Value, NoError>) {
 -		self.init(unsafeProducer: SignalProducer(signal: signal).prefix(value: initial),
 +		self.init(unsafeProducer: SignalProducer(signal).prefix(value: initial),
  		          capturing: [])
  	}
```

mdiep:
I'm pretty ambivalent about this. It's nice on the one hand—I always prefer unnamed arguments when they're unambiguous—but this initializer doesn't seem widely used, so it might be a little confusing.

andersio:
In a short discussion with @sharplet, we came up with a more generalised idea of dropping the label for all initializers that take a sequence.

The argument was:
> Do we need to distinguish it here _(the variants of initializer)_ though? Both are conceptually _(taking)_ a sequence, and results in a producer that emits a sequence of values.
>
> It has to be distinguished, IMO, only if it would result in different behavior, or has loss of information e.g. type narrowing. Moreover, at the call site, one _should_ already have known what it is when it comes to type conversion or initialization. Otherwise, garbage in garbage out (?).

Summing up with @sharplet's words:
> Yeah I'd agree with that. You could say that these convert a time-based sequence of events to a `SignalProducer`, where the `Sequence` version is immediate and the `Signal` version is streamed.

sharplet:
> Not sure if we should count the vararg initializer as one though.

I'd recommend avoiding that ambiguity if we can.

Sources/SignalProducer.swift
```diff
@@ -29,7 +29,7 @@ public struct SignalProducer<Value, Error: Swift.Error> {
 	///
 	/// - parameters:
 	///   - signal: A signal to observe after starting the producer.
-	public init<S: SignalProtocol>(signal: S) where S.Value == Value, S.Error == Error {
+	public init<S: SignalProtocol>(_ signal: S) where S.Value == Value, S.Error == Error {
```
sharplet: Likewise, what do you think about naming this `values` internally?

Sources/SignalProducer.swift
```diff
@@ -122,7 +122,7 @@ public struct SignalProducer<Value, Error: Swift.Error> {
 	///   - second: Second value for the `Signal` to send.
 	///   - tail: Rest of the values to be sent by the `Signal`.
 	public init(values first: Value, _ second: Value, _ tail: Value...) {
```
sharplet: It makes sense to me to keep the `values:` label here.

sharplet:
I'd like to add that the discussion was informed specifically by [this section of the API design guidelines](https://swift.org/documentation/api-design-guidelines/#argument-labels).

Sources/SignalProducer.swift
andersio: It'd better stay `signal`, since it has more than just values but also termination events.

Sources/SignalProducer.swift
```diff
@@ -100,7 +100,7 @@ public struct SignalProducer<Value, Error: Swift.Error> {
 	/// - parameters:
 	///   - values: A sequence of values that a `Signal` will send as separate
```
NachoSoto: I'm still not a fan of any of this because users are left to the expense of the type inference to determine which initializer to use, rather than being able to explicitly pick one.
If Swift's type inference and diagnostics were great I might not mind, but considering how poor they are, it's my opinion that this will bring more problems than benefit (in fact I can't think of any benefit other than conciseness, which is not necessarily an advantage).

andersio: It doesn't feel like the type checker could ever cause any chaos here though.

`SignalProducer` is a discrete time sequence, while `Sequence` is immediate. This means we would likely never conform `SignalProducer` to `Sequence`, or in other words the two signatures would never collide in any sense.

mdiep:
I wish we could have unanimity here, but we do have consensus and—more importantly—the Swift API guidelines seem pretty clear on this point. So I'm going to go ahead and merge.

Issue #122, opened 2016-11-25
Synthetic benchmark on a few pending PRs combined.


andersio:
[Benchmark Source](https://gist.github.com/andersio/5cc8fd27a8adccf70d53036505069db3)
#106 + #112 + #107 + #113 

#### Seven scenarios

1. Send 10000 events to 1 observer of a plain signal.
2. Send 10000 events to 8 observers of a plain signal.
3. Send 10000 events to 1 observer of a produced signal from a 5-times-transformed producer.
4. Send 10000 events to 2 observer of a produced signal from a 5-times-transformed producer.
5. Start, send one event to and dispose a 5-times-transformed producer for 10000 times.
6. Start a 5-times-transformed producer for 10000 times.
7. Start and dispose a 5-times-transformed producer for 10000 times.

#### TL;DR
1. A non-trivial regression in sending events. (as a result of the interruption handling fix in #112). Transformed producer would amplify the effect.
2. But if there is enough work done for each event (e.g. invoking eight empty observers), the regression would be outweighed.
3. Single-digit-percentage improvements in the `SignalProducer` starting & disposing performance.
1. When it comes to a common pattern for a scrolling collection views (rebinding reusable cells to different producers, represented by scenario 5), there is a minor net improvement.

#### Further optimization

1. Synchronise `Signal.observers` using read-copy-update. This removes a lock from the critical path of a `Signal`, at a cost of more expensive observer insertion.

```
// -Owholemodule
// `signal-state-patch` + `disposable-lockfree` + `lift-fix` + `minor-signal-refactor`
One Observer:    0.026 12%
Eight Observers: 0.042 8%
Producer:        0.326 3%
Producer 2 obv:  0.632 2%
Producer Mixed:  2.038 2%
Start Producer:  1.582 2%
Start & Dispose: 1.581 2%

// `master`
One Observer:    0.018 10%
Eight Observers: 0.053 5%
Producer:        0.223 3%
Producer 2 obv:  0.422 2%
Producer Mixed:  2.090 2%
Start Producer:  1.663 2%
Start & Dispose: 1.683 2%

// Speedup
One Observer:    0.692x ***
Eight Observers: 1.262x
Producer:        0.684x ***
Producer 2 obv:  0.668x ***
Producer Mixed:  1.025x
Start Producer:  1.051x
Start & Dispose: 1.065x
```

PR #123, opened 2016-11-25
Mitigate race conditions in `Signal` interrupt handling.

andersio:
Superseded #112.

#### Mitigated race conditions
1. [Contention window after the lock on `interrupted` is released, and before `sendLock` is released.](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/Signal.swift#L92) It is racing against the [interrupted event handling routine](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/Signal.swift#L71), if the `interrupted` event is sent on a different thread from the current sender.

2. [The retrieval of observers](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/Signal.swift#L80) is racing against the interrupt event sending ([1](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/Signal.swift#L71), [2](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/Signal.swift#L74)). It is possible to have events delivered after `interrupted` event is delivered.

Sure. The measured regression is to that in #122. Note that the PR includes no optimisation. There are a few optimizations I can think of that can always be done later.

> Why don't you replace `sendLock` and `terminated` with `Atomic<Bool>`?

The fun fact is that even under `-Owholemodule`, I still managed to measure a hefty overhead (~20%) in using closures-based, non-copying `Atomic.modify`. Maybe the `Atomic` optimization patch would help though.

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -815,13 +815,13 @@ extension SignalProtocol {
 	/// - parameters:
 	///   - event: A closure that accepts an event and is invoked on every
 	///            received event.
-	///   - value: A closure that accepts a value from `value` event.
 	///   - failed: A closure that accepts error object and is invoked for
 	///             failed event.
 	///   - completed: A closure that is invoked for `completed` event.
 	///   - interrupted: A closure that is invoked for `interrupted` event.
 	///   - terminated: A closure that is invoked for any terminating event.
 	///   - disposed: A closure added as disposable when signal completes.
+	///   - value: A closure that accepts a value from `value` event.
 	///
 	/// - returns: A signal with attached side-effects for given event cases.
 	public func on(
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -57,17 +57,14 @@ public final class Signal<Value, Error: Swift.Error> {
 			}
 
 			if case .interrupted = event {
-				// Normally we disallow recursive events, but `interrupted` is
-				// kind of a special snowflake, since it can inadvertently be
-				// sent by downstream consumers.
+				// Recursive events are generally disallowed. But `interrupted` is kind
+				// of a special snowflake, since it can inadvertently be sent by
+				// downstream consumers.
 				//
-				// So we'll flag Interrupted events specially, and if it
-				// happened to occur while we're sending something else,  we'll
-				// wait to deliver it.
-
-
-				// Clear the signal state to prevent new observers, but retain a copy of
-				// it.
+				// So we would treat `interrupted` events specially. If it happens
+				// to occur while the `sendLock` is acquired, the observer call-out and
+				// the disposal would be delegated to the current sender, or
+				// occasionally one of the senders waiting on `sendLock`.
 				if let state = signal.state.swap(nil) {
 					interruptedState.value = state
 
@@ -95,6 +92,8 @@ public final class Signal<Value, Error: Swift.Error> {
 							observer.action(event)
 						}
 
+						// Check if a downstream consumer or a concurrent sender has
+						// interrupted the signal.
 						if !isTerminating, let state = interruptedState.swap(nil) {
 							for observer in state.observers {
 								observer.sendInterrupted()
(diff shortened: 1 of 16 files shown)


Tests/ReactiveSwiftTests/SignalSpec.swift
```diff
@@ -213,6 +214,97 @@ class SignalSpec: QuickSpec {
...
+				if #available(macOS 10.10, *) {
+					queue = DispatchQueue.global(qos: .userInitiated)
+				} else {
+					queue = DispatchQueue.global(priority: .high)
+				}
+
+				let (signal, observer) = Signal<Int, NoError>.pipe()
+
+				var hasSlept = false
+				var events = [Event<Int, NoError>]()
```
NachoSoto: The format we normally use is separating type and value: `var events: [Event<Int, NoError>] = []`.

Tests/ReactiveSwiftTests/SignalSpec.swift
```diff
@@ -213,6 +214,97 @@ class SignalSpec: QuickSpec {
...
+
+				let sema = DispatchSemaphore(value: 0)
```
NachoSoto: No need to shorten this name (especially when `sema` has other meanings), just call if semaphore?

Tests/ReactiveSwiftTests/SignalSpec.swift
```diff
@@ -213,6 +214,97 @@ class SignalSpec: QuickSpec {
...
+					if index == 0 {
+						sema.wait()
+						observer.sendInterrupted()
+					}
+				}
+
+				group.wait()
+
+				expect(events.count) == 2
+				expect(events.count >= 2 ? events[1].isTerminating : false) == true
```
NachoSoto: This failure would be impossible to read. Can you separate the 2 checks with an `if`? That way it's possible to see what was checked and what the expectation was.

Tests/ReactiveSwiftTests/SignalSpec.swift
```diff
@@ -213,6 +214,97 @@ class SignalSpec: QuickSpec {
...
+				let group = DispatchGroup()
+
+				queue.async(group: group) {
+					DispatchQueue.concurrentPerform(iterations: iterations) { _ in
+						let (signal, observer) = Signal<(), NoError>.pipe()
+
+						var isInterrupted = false
+						signal.observeInterrupted { counter.modify { $0 += 1 } }
+
+						let sema = DispatchSemaphore(value: 0)
```
NachoSoto: Ditto

Sources/Signal.swift
```diff
@@ -40,58 +40,109 @@ public final class Signal<Value, Error: Swift.Error> {
...
 
-					if event.isTerminating || shouldInterrupt {
+					// Based on the implicit memory order, any updates to the
+					// `interruptedState` should always be visible after `sendLock` is
+					// released. So we check it again and handle the interruption if
+					// it has not been taken over.
+					if !shouldDispose && !terminated && !isTerminating, let state = interruptedState {
+						sendLock.lock()
+
+						if !terminated {
```
NachoSoto: This is already checked in the condition.

andersio: It could be a false negative. `terminated` is flipped only within the protected section of `sendLock`, while here we are checking it outside the lock.

Tests/ReactiveSwiftTests/SignalSpec.swift
andersio: I guess we'd need a set of coding guidelines then. Just did a search and found quite a few of `[U]()` 🙈

Sources/Signal.swift
NachoSoto: That could use a comment then to make sure a future reader doesn't accidentally removed it.

andersio: https://github.com/ReactiveCocoa/ReactiveSwift/pull/123/commits/7a73f824e4d4a8830e5095bd388b21c631c60c40

Added before merging. :tada:

PR #124, opened 2016-11-25
`Atomic`: The Unexpected Journey.

andersio:
#### The benchmark
[Source](https://gist.github.com/andersio/9b3f64754e4f35437673fe78a693209f)
It simulates the critical path of `Signal`, which is grabbing the signal state + querying the `interrupted`.

Compiler Settings: Both the framework and the test are compiled with `-Owholemodule`. Checked builds.
Test Environment: i5 4258U, macOS 10.12.1, Xcode 8.1, Swift 3.0.1

#### Final Speedup
4.641x 😦

The optimised `Atomic` now takes only 0.029 sec (5%) in the synthetic test of the [lock-free disposal PR](https://github.com/ReactiveCocoa/ReactiveSwift/pull/107), significantly faster than it was before.

#### Baseline Performance
`master`
0.673 sec (2%)

#### The Journey
Mark `AtomicProtocol.value` and `AtomicProtocol.swap` as `@inline(__always)`.
0.434 sec (2%)

Mark `Atomic.modify` and `Atomic.swap` as `@inline(__always)`.
0.296 sec (2%)

Mark `PosixThreadMutex.lock` and `PosixThreadMutex.unlock` as `@inline(__always)`.
0.223 sec (2%)

Drop `defer` statements in `Atomic.modify` and `Atomic.swap`.
0.212 sec (3%)

Drop `precondition` in favour of `if-fatalError`.
0.188 sec (2%)

Embed `PosixThreadMutex` into `Atomic`.
0.179 sec (4%)

Implement `value` and `swap` in `Atomic`.
0.149 sec (4%)

Use the non-copying `modify` instead of `withValue` in `value.get`.
0.145 sec (2%)

#### Misc 
Inline `modify` into `swap` manually.
0.145 sec (2%)

Closure based `_PosixThreadMutex.lock`.
0.179 sec (3%)

`-Onone`
0.479 sec (3%)

### DIFF
--- a/Sources/Atomic.swift
+++ b/Sources/Atomic.swift
@@ -125,6 +125,12 @@ internal struct UnsafeAtomicState<State: RawRepresentable>: AtomicStateProtocol 
 #endif
 }
 
+/// A pthread mutex wrapper. Manual memory management by calling
+/// `deinitialize` is required.
+///
+/// The wrapper is a struct so that it can be inlined when it is contained as
+/// part of a reference type, and dodges an extra level of calling indirection
+/// that would otherwise happen if the wrapper itself is an reference type.
 internal struct PosixThreadMutex {
 	private let mutex: UnsafeMutablePointer<pthread_mutex_t>
 
--- a/Sources/Atomic.swift
+++ b/Sources/Atomic.swift
@@ -379,7 +379,7 @@ extension AtomicProtocol {
 	/// - returns: The old value.
 	@discardableResult
 	@inline(__always)
-	func swap(_ newValue: Value) -> Value {
+	public func swap(_ newValue: Value) -> Value {
 		return modify { (value: inout Value) in
 			let oldValue = value
 			value = newValue
--- a/Sources/Atomic.swift
+++ b/Sources/Atomic.swift
@@ -210,6 +210,8 @@ public final class Atomic<Value>: AtomicProtocol {
 
 	/// Atomically modifies the variable.
 	///
+	/// - note: Throwing in `action` does not revert any changes made.
+	///
 	/// - parameters:
 	///   - action: A closure that takes the current value.
 	///
@@ -222,7 +224,7 @@ public final class Atomic<Value>: AtomicProtocol {
 			let value = try action(&_value)
 			lock.unlock()
 			return value
-		} catch let error {
+		} catch {
 			lock.unlock()
 			throw error
 		}
@@ -243,7 +245,7 @@ public final class Atomic<Value>: AtomicProtocol {
 			let value = try action(_value)
 			lock.unlock()
 			return value
-		} catch let error {
+		} catch {
 			lock.unlock()
 			throw error
 		}
@@ -258,7 +260,7 @@ public final class Atomic<Value>: AtomicProtocol {
 	@discardableResult
 	@inline(__always)
 	public func swap(_ newValue: Value) -> Value {
-		return modify { (value: inout Value) in
+		return modify { value in
 			let oldValue = value
 			value = newValue
 			return oldValue
@@ -302,6 +304,9 @@ internal final class RecursiveAtomic<Value>: AtomicProtocol {
 
 	/// Atomically modifies the variable.
 	///
+	/// - note: Throwing in `action` does not revert any changes made, and the
+	///         `didSetObserver` would be called regardless.
+	///
 	/// - parameters:
 	///   - action: A closure that takes the current value.
 	///
@@ -315,7 +320,8 @@ internal final class RecursiveAtomic<Value>: AtomicProtocol {
 			didSetObserver?(_value)
 			lock.unlock()
 			return returnValue
-		} catch let error {
+		} catch {
+			didSetObserver?(_value)
 			lock.unlock()
 			throw error
 		}
@@ -336,7 +342,7 @@ internal final class RecursiveAtomic<Value>: AtomicProtocol {
 			let returnValue = try action(_value)
 			lock.unlock()
 			return returnValue
-		} catch let error {
+		} catch {
 			lock.unlock()
 			throw error
 		}
@@ -377,7 +383,7 @@ extension AtomicProtocol {
 	@discardableResult
 	@inline(__always)
 	func swap(_ newValue: Value) -> Value {
-		return modify { (value: inout Value) in
+		return modify { value in
 			let oldValue = value
 			value = newValue
 			return oldValue
--- a/Sources/Atomic.swift
+++ b/Sources/Atomic.swift
@@ -214,9 +214,14 @@ public final class Atomic<Value>: AtomicProtocol {
 	@inline(__always)
 	public func modify<Result>(_ action: (inout Value) throws -> Result) rethrows -> Result {
 		lock.lock()
-		let value = try action(&_value)
-		lock.unlock()
-		return value
+		do {
+			let value = try action(&_value)
+			lock.unlock()
+			return value
+		} catch let error {
+			lock.unlock()
+			throw error
+		}
 	}
 	
 	/// Atomically perform an arbitrary action using the current value of the
@@ -230,9 +235,14 @@ public final class Atomic<Value>: AtomicProtocol {
 	@inline(__always)
 	public func withValue<Result>(_ action: (Value) throws -> Result) rethrows -> Result {
 		lock.lock()
-		let value = try action(_value)
-		lock.unlock()
-		return value
+		do {
+			let value = try action(_value)
+			lock.unlock()
+			return value
+		} catch let error {
+			lock.unlock()
+			throw error
+		}
 	}
 
 	/// Atomically replace the contents of the variable.
@@ -296,10 +306,15 @@ internal final class RecursiveAtomic<Value>: AtomicProtocol {
 	@inline(__always)
 	func modify<Result>(_ action: (inout Value) throws -> Result) rethrows -> Result {
 		lock.lock()
-		let returnValue = try action(&_value)
-		didSetObserver?(_value)
-		lock.unlock()
-		return returnValue
+		do {
+			let returnValue = try action(&_value)
+			didSetObserver?(_value)
+			lock.unlock()
+			return returnValue
+		} catch let error {
+			lock.unlock()
+			throw error
+		}
 	}
 	
 	/// Atomically perform an arbitrary action using the current value of the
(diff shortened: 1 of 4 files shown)


Sources/Atomic.swift
```diff
@@ -104,14 +176,13 @@ internal final class RecursiveAtomic<Value>: AtomicProtocol {
...
 		lock.lock()
-		defer {
-			didSetObserver?(_value)
-			lock.unlock()
-		}
-
-		return try action(&_value)
+		let returnValue = try action(&_value)
+		didSetObserver?(_value)
+		lock.unlock()
```
sharplet: Without `defer`, if `try action(&_value)` throws, the lock will never be unlocked.

Sources/Atomic.swift
```diff
@@ -122,34 +193,12 @@ internal final class RecursiveAtomic<Value>: AtomicProtocol {
...
-	public var value: Value {
-		get {
-			return withValue { $0 }
-		}
-	
-		set(newValue) {
-			swap(newValue)
-		}
+		let returnValue = try action(_value)
+		lock.unlock()
```
sharplet: I think we should bring back `defer` here too.

andersio: Hmm, this is an oversight. Thanks for catching this.

andersio:
The hole @sharplet caught is fixed. `defer` still isn't used though, since it bumps the time up by ~20% due to its closure based implementation.

P.S. It seems in `master`, `didSetObserver` is called when an error is thrown.

Sources/Atomic.swift
```diff
@@ -9,37 +9,44 @@
...
 		precondition(result == 0, "Failed to initialize mutex with error \(result).")
 	}
 
-	mutating func deinitialize() {
-		let result = pthread_mutex_destroy(&mutex)
+	func deinitialize() {
+		let result = pthread_mutex_destroy(mutex)
 		precondition(result == 0, "Failed to destroy mutex with error \(result).")
+
+		UnsafeMutableRawPointer(mutex).deallocate(bytes: MemoryLayout<pthread_mutex_t>.size,
```
sharplet: I think of using a raw pointer here it might make sense to use `mutex.deinitialize(1)` and `mutex.deallocate(1)`? I'm not super familiar with the new pointer APIs though.

Sources/Atomic.swift
```diff
@@ -97,9 +97,14 @@ public final class Atomic<Value>: AtomicProtocol {
...
 	public func modify<Result>(_ action: (inout Value) throws -> Result) rethrows -> Result {
 		lock.lock()
-		let value = try action(&_value)
-		lock.unlock()
-		return value
+		do {
+			let value = try action(&_value)
+			lock.unlock()
+			return value
+		} catch let error {
```
sharplet: You can omit `let error` here, and in the other `catch` blocks.

Sources/Atomic.swift
```diff
@@ -68,11 +122,33 @@ public final class Atomic<Value>: AtomicProtocol {
...
+	/// Atomically replace the contents of the variable.
+	///
+	/// - parameters:
+	///   - newValue: A new value for the variable.
+	///
+	/// - returns: The old value.
+	@discardableResult
+	@inline(__always)
+	public func swap(_ newValue: Value) -> Value {
+		return modify { (value: inout Value) in
```
sharplet: Is the closure type annotation necessary here?

andersio:
Hmm. It won't restore automatically. Now I see why it makes sense to notify regardless, and the restoration should be left to the users.

Sources/Atomic.swift
andersio: TIL

Sources/Atomic.swift
andersio: Seems `UnsafeMutablePointer` is enough.

> Instances must be aligned to MemoryLayout<Pointee>.alignment, i.e. (UnsafePointer<Int8>(self) - nil) % MemoryLayout<Pointee>.alignment == 0

Sources/Atomic.swift
andersio: Removed.

Sources/Atomic.swift
```diff
@@ -8,56 +8,107 @@
...
+		let result = pthread_mutex_init(mutex, nil)
 		precondition(result == 0, "Failed to initialize mutex with error \(result).")
 	}
 
-	deinit {
-		let result = pthread_mutex_destroy(&mutex)
+	func deinitialize() {
+		let result = pthread_mutex_destroy(mutex)
 		precondition(result == 0, "Failed to destroy mutex with error \(result).")
+		mutex.deallocate(capacity: 1)
```
sharplet: https://developer.apple.com/reference/swift/unsafemutablepointer/2295090-deallocate:

> Precondition: The memory is not initialized.

This is what made me think that perhaps we should be calling `mutex.deinitialize()` first.

andersio: According to the memory model, it has to be first initialised too. Addressed in https://github.com/ReactiveCocoa/ReactiveSwift/pull/124/commits/974663c308911d2dc156dd07f38d15f7d3cf0ef7.

Sources/Atomic.swift
```diff
@@ -8,56 +8,109 @@
...
+	@inline(__always)
+	func unlock() {
+		let result = pthread_mutex_unlock(mutex)
+		if result != 0 {
+			fatalError("Failed to unlock \(self) with error \(result).")
+		}
+	}
+}
+
+internal final class PosixThreadMutex: NSLocking {
```
mdiep: It doesn't look like this is used anywhere? Let's remove it and then remove the underscore from the version we _are_ using.

Sources/Atomic.swift
```diff
@@ -122,33 +218,16 @@ internal final class RecursiveAtomic<Value>: AtomicProtocol {
...
-	@discardableResult
-	func withValue<Result>(_ action: (Value) throws -> Result) rethrows -> Result
-
-	@discardableResult
-	func modify<Result>(_ action: (inout Value) throws -> Result) rethrows -> Result
-}
-
-extension AtomicProtocol {	
-	/// Atomically get or set the value of the variable.
-	public var value: Value {
```
mdiep: Should we still keep this and `swap` as part of `AtomicProtocol`?

andersio: The performance is much worse due to the need to query the protocol witness table, even with the `@inline(__always)` annotation. I have rerun my benchmarks and it is still the case.

I am actually thinking of removing `AtomicProtocol`, hmm...

> Embed PosixThreadMutex into Atomic.
> 0.179 sec (4%)
>
> Implement value and swap in Atomic.
> 0.149 sec (4%)

Sources/Atomic.swift
```diff
@@ -8,27 +8,37 @@
 
 import Foundation
 
-final class PosixThreadMutex: NSLocking {
-	private var mutex = pthread_mutex_t()
+internal struct PosixThreadMutex {
```
mdiep: Why is this a `struct` an `UnsafeMutablePointer` instead of a `class` with a `deinit`?

andersio: It is usually used with an apparent owner. So defining it as a struct gives the flexibility to have it embedded in the containing type, should it be a stored property, and one less pointer to chase.

> Drop precondition in favour of if-fatalError.
> 0.188 sec (2%)
>
> Embed PosixThreadMutex into Atomic.
> 0.179 sec (4%)

andersio:
Related: https://lists.swift.org/pipermail/swift-users/Week-of-Mon-20161205/004147.html

andersio:
Related: https://lists.swift.org/pipermail/swift-users/Week-of-Mon-20161205/004147.html

> _Joe Groff jgroff at apple.com_
> If you apply them to memory you allocated manually with malloc/free on UnsafeMutablePointer's allocation methods, then yeah, they should work as they do in C. That's the safest way to use these functions today. Passing a Swift `var` inout to one of these functions does not guarantee that accesses to that var will maintain atomicity, since there may be bridging or reabstracting conversions happening under the hood.

Sources/Atomic.swift
```diff
@@ -163,18 +186,30 @@ public final class Atomic<Value>: AtomicProtocol {
...
-
-		return try action(&_value)
+		do {
+			let value = try action(&_value)
+			lock.unlock()
+			return value
+		} catch {
+			lock.unlock()
+			throw error
+		}
```
mdiep: What do you think about keeping the `defer`s?

Removing them seems to have a very minor effect on the speed but makes mistakes more likely.

Sources/Atomic.swift
```diff
@@ -277,7 +378,8 @@ extension AtomicProtocol {
 	///
 	/// - returns: The old value.
 	@discardableResult
-	public func swap(_ newValue: Value) -> Value {
+	@inline(__always)
+	func swap(_ newValue: Value) -> Value {
```
mdiep: Is this still public?

Sources/Atomic.swift
```diff
@@ -125,27 +125,37 @@ internal struct UnsafeAtomicState<State: RawRepresentable>: AtomicStateProtocol
 #endif
 }
 
-final class PosixThreadMutex: NSLocking {
-	private var mutex = pthread_mutex_t()
+internal struct PosixThreadMutex {
+	private let mutex: UnsafeMutablePointer<pthread_mutex_t>
```
mdiep: Can you add a comment that explains this?

Any time we choose the non-obvious route to get a speed boost, we should document it.

Sources/Atomic.swift
andersio: How likely are we to change this though? It are pieces that perhaps would never be altered otherwise.

Sources/Atomic.swift
andersio: `Atomic.swap` is. `RecursiveAtomic.swap` isn't.

Sources/Atomic.swift
```diff
@@ -131,7 +131,7 @@ internal struct UnsafeAtomicState<State: RawRepresentable>: AtomicStateProtocol
 /// The wrapper is a struct so that it can be inlined when it is contained as
 /// part of a reference type, and dodges an extra level of calling indirection
 /// that would otherwise happen if the wrapper itself is an reference type.
-internal struct PosixThreadMutex {
+internal struct PthreadMutex {
```
andersio: Well, it is generally called pthread anyway... So let's make it a bit shorter...

Sources/Atomic.swift
mdiep: There's no way to know. 🙃

mdiep: There's no way to know. 🙃

PR #128, opened 2016-11-28
Add withLatest(from:) (Rx.withLatestFrom)

inamiy:
This PR is RAC port of `Rx.withLatestFrom` (namely, `sample(from:)`) with `samplee`'s error type constrained to `NoError` only.

Diagram: http://rxmarbles.com/#withLatestFrom

This was originally discussed in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2082 (not accepted), but I found this operator very useful because it is easy to fetch current state (`MutableProperty`) via `sourceSignal`:

```swift
let state: MutableProperty<String> = ...

sourceSignal
    .sample(from: state.producer)
    .observeValues { sourceValue, currentState in ... }
```

Above code is much easier to read compared to:

```swift
sourceSignal
    .map { [unowned state] sourceValue in (sourceValue, state.value) }
    .observeValues { sourceValue, currentState in ... }
```

or:

```swift
state.producer
    .sample(with: sourceSignal)
    .observeValues { currentState, sourceValue in ... }
// receiver is flipped, and hard to understand which signal is the source of this whole chaining
```

Please note that `sample(from:)` is NOT actually a flipped version of `sample(with:)` since it discards `samplee`'s any terminal events.
For `.completed` handling, this behavior is same as `Rx.withLatestFrom` (and is often easy to use).

---

The code is originally from: https://github.com/inamiy/ReactiveAutomaton/blob/28027a656b923008f34089c739e51cdbe13b65cb/Sources/ReactiveCocoa%2BSampleFrom.swift

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -1030,7 +1030,7 @@ extension SignalProtocol {
 	///         nothing happens.
 	///
 	/// - parameters:
-	///   - samplee: A signal that its latest value is sampled by `self`.
+	///   - samplee: A signal whose latest value is sampled by `self`.
 	///
 	/// - returns: A signal that will send values from `self` and `samplee`,
 	///            sampled (possibly multiple times) by `self`, then terminate
@@ -1073,7 +1073,7 @@ extension SignalProtocol {
 	///         nothing happens.
 	///
 	/// - parameters:
-	///   - samplee: A producer that its latest value is sampled by `self`.
+	///   - samplee: A producer whose latest value is sampled by `self`.
 	///
 	/// - returns: A signal that will send values from `self` and `samplee`,
 	///            sampled (possibly multiple times) by `self`, then terminate
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -1036,7 +1036,7 @@ extension SignalProtocol {
 	///            sampled (possibly multiple times) by `self`, then terminate
 	///            once `self` has terminated. **`samplee`'s terminated events
 	///            are ignored**.
-	public func withLatest<U>(_ samplee: Signal<U, NoError>) -> Signal<(Value, U), Error> {
+	public func withLatest<U>(from samplee: Signal<U, NoError>) -> Signal<(Value, U), Error> {
 		return Signal { observer in
 			let state = Atomic<U?>(nil)
 			let disposable = CompositeDisposable()
@@ -1079,12 +1079,12 @@ extension SignalProtocol {
 	///            sampled (possibly multiple times) by `self`, then terminate
 	///            once `self` has terminated. **`samplee`'s terminated events
 	///            are ignored**.
-	public func withLatest<U>(_ samplee: SignalProducer<U, NoError>) -> Signal<(Value, U), Error> {
+	public func withLatest<U>(from samplee: SignalProducer<U, NoError>) -> Signal<(Value, U), Error> {
 		return Signal { observer in
 			let d = CompositeDisposable()
 			samplee.startWithSignal { signal, disposable in
 				d += disposable
-				d += self.withLatest(signal).observe(observer)
+				d += self.withLatest(from: signal).observe(observer)
 			}
 			return d
 		}
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -1021,6 +1021,75 @@ extension SignalProtocol {
 			.map { $0.0 }
 	}
 
+	/// Forward the latest value from `samplee` with the value from `self` as a
+	/// tuple, only when `self` sends a `value` event.
+	/// This is like a flipped version of `sample(with:)` and Rx's `withLatestFrom`.
+	///
+	/// - note: If `self` fires before a value has been observed on `samplee`,
+	///         nothing happens.
+	///
+	/// - parameters:
+	///   - samplee: A signal that its latest value is sampled by `self`.
+	///
+	/// - returns: A signal that will send values from `self` and `samplee`,
+	///            sampled (possibly multiple times) by `self`, then terminate
+	///            once `self` has terminated. **`samplee`'s terminated events
+	///            are ignored**.
+	public func sample<U>(from samplee: Signal<U, NoError>) -> Signal<(Value, U), Error> {
+		return Signal { observer in
+			let state = Atomic<U?>(nil)
+			let disposable = CompositeDisposable()
+
+			disposable += samplee.observeValues { value in
+				state.value = value
+			}
+
+			disposable += self.observe { event in
+				switch event {
+				case let .value(value):
+					if let value2 = state.value {
+						observer.send(value: (value, value2))
+					}
+				case .completed:
+					observer.sendCompleted()
+				case let .failed(error):
+					observer.send(error: error)
+				case .interrupted:
+					observer.sendInterrupted()
+				}
+			}
+
+			return disposable
+		}
+	}
+
+	/// Forward the latest value from `samplee` with the value from `self` as a
+	/// tuple, only when `self` sends a `value` event.
+	/// This is like a flipped version of `sample(with:)` and Rx's `withLatestFrom`.
+	///
+	/// - note: If `self` fires before a value has been observed on `samplee`,
+	///         nothing happens.
+	///
+	/// - parameters:
+	///   - samplee: A producer that its latest value is sampled by `self`.
+	///
+	/// - returns: A signal that will send values from `self` and `samplee`,
+	///            sampled (possibly multiple times) by `self`, then terminate
+	///            once `self` has terminated. **`samplee`'s terminated events
+	///            are ignored**.
+	public func sample<U>(from samplee: SignalProducer<U, NoError>) -> Signal<(Value, U), Error> {
+		return Signal { observer in
+			let d = CompositeDisposable()
+			samplee.startWithSignal { signal, disposable in
+				d += disposable
+				d += self.sample(from: signal).observe(observer)
+			}
+			return d
+		}
+	}
+}
+
+extension SignalProtocol {
 	/// Forwards events from `self` until `lifetime` ends, at which point the
 	/// returned signal will complete.
 	///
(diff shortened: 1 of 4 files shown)


sharplet:
Do you have an example with more meaningful names than `sourceSignal` and `state`? I think that would help to demonstrate the readability benefits.

inamiy:
@sharplet 
Button with "2 enabled flags" can be a good example:

```swift
let enabled1 = MutableProperty<Bool>(true) 
let enabled2 = MutableProperty<Bool>(true) 
let buttonTap: Signal<(), NoError> = ...

buttonTap
    .sample(from: enabled1.producer)
    .sample(from: enabled2.producer)  // this signal is `Signal<((Void, Bool), Bool), NoError>`
    .filter { $0.1 && $1 }
    .observe { _ in print("hello") }  // prints only when both `enabled1` & `enabled2` are true
```

mdiep:
That example seems like it'd be better with `sample(on:)`.

```swift
let enabled1 = MutableProperty<Bool>(true) 
let enabled2 = MutableProperty<Bool>(true) 
let buttonTap: Signal<(), NoError> = ...

SignalProducer
    .combineLatest([ enabled1.producer, enabled2.producer ])
    .sample(on: buttonTap)
    .filter { $0 && $1 }
    .observe { _ in print("hello") }  // prints only when both `enabled1` & `enabled2` are true
```

inamiy:
@mdiep 
Oh, your code works very nice 😅
How about asynchronously fetching 2 states, i.e. `sample(with:).flatMap().sample(with:)`?

```swift
/// Stores API response items.
let items = MutableProperty<[Item]>([])

items <~ buttonTap
    .sample(from: urlTextField.reactive.text)
    .map { APIRequest($1) }
    .flatMap(.merge) { sendRequest($0) }
    .sample(from: items.producer)
    .map { $1 + [$0.item] }

items.signal
    .observeValues { print("items updated: \($0)") }
```

inamiy:
Anyway, the dominant signal of above cases is always `buttonTap` which is what I mean by `sourceSignal`. Placing it at topmost of the method chaining (receiver) is always more readable.

mdiep:
I definitely think the `sample(from:)` name is wrong. Similarities exist with the `sample(on:)` and `sample(with:)`, but I think the defining characteristic of `sample` is that it's `source.sample(_: trigger)`. The name totally trips me up.

`withLatest` seems like a much clearer name.

But your example still doesn't seem ideal. In particular, I wouldn't use `items` in its own definition.

```
/// Stores API response items.
let items = MutableProperty<[Item]>([])

items <~ urlTextField.reactive.text
    .sample(on: buttonTap)
    .map { APIRequest($1) }
    .flatMap(.merge) { sendRequest($0) }
    .reduce([]) { $1 + $0 }

items.signal
    .observeValues { print("items updated: \($0)") }
```

sharplet:
And with the recent improvements to `Action`, you can take a different approach without any mutable state:

```swift
let url = Property(initial: nil, then: urlTextField.reactive.text)

// Only enabled when input url is non-nil
let submit: Action<(), [Item], Error> = Action(input: url) { url in
  // url is the current value when the button is tapped
  let request = APIRequest(url)
  return sendRequest(request)
}

// bind button to submit action:
//   - button is disabled while executing
//   - button is disabled if input is invalid
button.reactive.pressed = CocoaAction(submit)

let items = Property(
  initial: [],
  then: submit.values.scan([]) { $1 + $0 }
)

items.signal
  .observeValues { print("items updated: \($0)") }

// you can also handle errors gracefully by observing `submit.errors`
```

inamiy:
@mdiep 
Renaming to `withLatest(_:)` is good to me too.
It's a cousin of `sample(with:)`, but not just a flipped version anymore.

By the way, your code has difficulty when `items.value` is updated elsewhere other than `buttonTap`.
`reduce` causes unnecessary extra state inside.

@sharplet
Same here, `scan` is unnecessary internal state. 
(and `Action` is overkill for just achieving `map { ($0, stateProperty.value) }`)

sharplet:
> Placing it at topmost of the data flow (receiver) is more readable.

Now that you phrase it in terms of data flow, I see your point. This flows pretty well for me at the point of use:

```swift
items <~ buttonTap
  .withLatest(from: textField.reactive.continuousTextValues)
  .flatMap(.latest) { api.getItems($1) }
  // etc.
```

Sources/Signal.swift
```diff
@@ -1021,6 +1021,75 @@ extension SignalProtocol {
...
+	///         nothing happens.
+	///
+	/// - parameters:
+	///   - samplee: A signal that its latest value is sampled by `self`.
+	///
+	/// - returns: A signal that will send values from `self` and `samplee`,
+	///            sampled (possibly multiple times) by `self`, then terminate
+	///            once `self` has terminated. **`samplee`'s terminated events
+	///            are ignored**.
+	public func sample<U>(from samplee: Signal<U, NoError>) -> Signal<(Value, U), Error> {
```
mdiep: I think `withLatest(_ samplee: Signal<U, NoError>)` would be clearer.

Sources/Signal.swift
```diff
@@ -1021,6 +1021,75 @@ extension SignalProtocol {
...
+	///         nothing happens.
+	///
+	/// - parameters:
+	///   - samplee: A producer that its latest value is sampled by `self`.
+	///
+	/// - returns: A signal that will send values from `self` and `samplee`,
+	///            sampled (possibly multiple times) by `self`, then terminate
+	///            once `self` has terminated. **`samplee`'s terminated events
+	///            are ignored**.
+	public func sample<U>(from samplee: SignalProducer<U, NoError>) -> Signal<(Value, U), Error> {
```
mdiep: This should be removed. We don't expose mixed operators like this. If someone wants to use a `Signal` with a `SignalProducer`, they need to convert the `Signal` to a `SignalProducer`.

Sources/SignalProducer.swift
```diff
@@ -812,6 +812,42 @@ extension SignalProducerProtocol {
...
+	///         nothing happens.
+	///
+	/// - parameters:
+	///   - samplee: A signal that its latest value is sampled by `self`.
+	///
+	/// - returns: A signal that will send values from `self` and `samplee`,
+	///            sampled (possibly multiple times) by `self`, then terminate
+	///            once `self` has terminated. **`samplee`'s terminated events
+	///            are ignored**.
+	public func sample<U>(from samplee: Signal<U, NoError>) -> SignalProducer<(Value, U), Error> {
```
mdiep: This should also be removed.

Sources/Signal.swift
```diff
@@ -1021,6 +1021,75 @@ extension SignalProtocol {
...
+					observer.sendCompleted()
+				case let .failed(error):
+					observer.send(error: error)
+				case .interrupted:
+					observer.sendInterrupted()
+				}
+			}
+
+			return disposable
+		}
```
mdiep: Can this be implemented with `sample(with:)`?

```
return samplee.sample(with: self).map { ($1, $0) }
```

Sources/Signal.swift
andersio: Alternative: `combineLatest(from:)`.

Sources/SignalProducer.swift
andersio: This is for producer though. We have a couple of cold + hot -> cold producer operators.

Sources/Signal.swift
andersio: We do not have `hot + cold -> hot` except for flattening operators though.

inamiy: Is it illegal to have different type for the argument? 
I think `hot -> T -> hot` or `cold -> T -> cold` keeps the same structure, and it's good enough.

Sources/Signal.swift
inamiy: I prefer `withLatest(_:)` because "with" sounds like a small addition to the main stream, and receiver remains as the first class that also _implies_ the rank direction. 
Maybe `withLatest(from:)` can be a compromise, but I think we don't need this "from".

Sources/Signal.swift
andersio: It is not illegal. Just that `hot -> cold -> hot` doesn't seem something very commonly used, except for flattening operators (`Signal<SignalProducer, E>`).

Sources/Signal.swift
andersio: `withLatest(_:)` should be fine then.
```swift
withLatest(textField.reactive.continuousTextValues)
```
reads like "with latest text values" to me.

andersio:
(Nit)

```swift
// Button taps with latest text field continuous text values.
buttonTaps.withLatest(textField.reactive.continuousTextValues)

// Button taps with (the) latest (one) from text field continuous text values
buttonTaps.withLatest(from: textField.reactive.continuousTextValues)
```

sharplet:
I agree, and think `withLatest(from:)` clarifies what "latest" refers to: the latest value from the signal.

inamiy:
OK, it seems 2:2 now, but I can go for `withLatest(from:)` too.
If no other votes, I will rename it again later.
(BTW, is travis dead? Test is not working...)

andersio:
Travis CI had a huge backlog in Mac jobs.

mdiep:
I'm 👍 with `withLatest(from:)`.

inamiy:
Great! Renaming is done in 95bd754, and previous test has finally passed.

Sources/Signal.swift
```diff
@@ -1021,6 +1021,77 @@ extension SignalProtocol {
...
+	/// Forward the latest value from `samplee` with the value from `self` as a
+	/// tuple, only when `self` sends a `value` event.
+	/// This is like a flipped version of `sample(with:)`, but `samplee`'s
+	/// terminal events are completely ignored.
+	///
+	/// - note: If `self` fires before a value has been observed on `samplee`,
+	///         nothing happens.
+	///
+	/// - parameters:
+	///   - samplee: A signal that its latest value is sampled by `self`.
```
sharplet: I think this would read better as "A signal whose latest value..."

inamiy: Fixed in 9580123.

PR #129, opened 2016-11-28
`Signal` lock elision.

andersio:
While we advocate for a safe-by-default API like how our beloved Swift does, there are places we may elide the serialization:

1. **Synchronously `Signal` operators. e.g. `map`, `filter`, `scan`, `skipNil`.**

    These can rely on the serialization at the ultimate upstream, e.g. signals created by `Signal.init` or `Signal.pipe` with default settings.

    An illustration of how the call stack would look like with `signal.map(t).map(t).map(t).map(t)...`:
    ```
    sendLock
    observers ---> [sendLock]
                   observers    ---> [sendLock]
                                     observers    ---> [sendLock]
                                                       observers    ---> ...
                                                       [unlock]
                                     [unlock] <---/
                   [unlock] <---/
    unlock    <---/
    ```

2. **Asynchronous `Signal` operators which deliver all kinds of events on the same `Scheduler`.**

    For example, `observe(on:)` forwards **all** the events to the supplied `Scheduler`. Since the scheduler serialises all the enqueued actions, there is no need for the created signal to serialise the events again.

2. **`Signal`s that are owned by synchronized entities.**

    For entities that are already synchronised, `Signal` can safely assume events not to be emitted concurrently. For example, `MutableProperty` always sends event in its critical section.

#### Implementation

The PR introduces an optional parameter `attributes` to `Signal.init` and `Signal.pipe` that allows configurability of some of the `Signal` behavior.

For the purpose of the PR, a `inheritSerialization` option was introduced. `Signal`s created with this flag would not attempt to serialise the inputs it received.

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -106,16 +106,16 @@ public final class Signal<Value, Error: Swift.Error> {
 		private let updateLock: Lock
 
 		/// Used to ensure that events are serialized during delivery to observers.
-		private let sendLock: Lock
+		private var sendLock: SendLock
 
 		/// Used to indicate if the `Signal` has deinitialized.
 		private var hasDeinitialized: Bool
 
-		fileprivate init(_ generator: (Observer) -> Disposable?) {
+		fileprivate init(_ serialization: Serialization, _ generator: (Observer) -> Disposable?) {
 			state = .alive(AliveState())
 
 			updateLock = Lock.make()
-			sendLock = Lock.make()
+			sendLock = SendLock(nonatomic: serialization == .nonatomic)
 
 			hasDeinitialized = false
 			disposable = SerialDisposable()
@@ -292,7 +292,7 @@ public final class Signal<Value, Error: Swift.Error> {
 		///               `tryTerminate` would attempt to acquire the `sendLock`.
 		///
 		/// - returns: `.shouldDispose` if the attempt succeeds. `.none` otherwise.
-		private func tryToCommitTermination(acquired sendLock: Lock? = nil) -> OperationResult {
+		private func tryToCommitTermination(acquired sendLock: SendLock? = nil) -> OperationResult {
 			assert(sendLock == nil || sendLock === self.sendLock,
 			       "`tryTerminate` receives a lock that is not the `sendLock` of the signal.")
 			func commit() -> OperationResult {
@@ -400,7 +400,11 @@ public final class Signal<Value, Error: Swift.Error> {
 	///   - generator: A closure that accepts an implicitly created observer
 	///                that will act as an event emitter for the signal.
 	public init(_ generator: (Observer) -> Disposable?) {
-		core = Core(generator)
+		core = Core(.default, generator)
+	}
+
+	internal init(_ attributes: Serialization, _ generator: (Observer) -> Disposable?) {
+		core = Core(attributes, generator)
 	}
 
 	/// Observe the Signal by sending any future events to the given observer.
@@ -420,6 +424,19 @@ public final class Signal<Value, Error: Swift.Error> {
 
 	deinit {
 		core.signalDidDeinitialize()
+	}
+
+	/// Describes how a `Signal` should serialize.
+	internal enum Serialization {
+		/// The `Signal` may assume that all senders to the input sink are serialized.
+		/// Generally speaking, `nonatomic` is **only** applicable when an operator:
+		///
+		/// 1. is sourced from only one upstream, and is not asynchronous in any sense; or
+		/// 2. forwards **all** events on the same `Scheduler`.
+		case nonatomic
+
+		/// The `Signal` should always serialize the events received from the input sink.
+		case `default`
 	}
 
 	/// The state of a `Signal`.
@@ -481,6 +498,47 @@ public final class Signal<Value, Error: Swift.Error> {
 	}
 }
 
+private struct SendLock {
+	fileprivate let _lock: Lock?
+	private var isSendingNonatomic: Bool
+
+	init(nonatomic: Bool) {
+		_lock = !nonatomic ? Lock.make() : nil
+		isSendingNonatomic = false
+	}
+
+	mutating func lock() {
+		if let lock = _lock {
+			return lock.lock()
+		}
+
+		isSendingNonatomic = true
+	}
+
+	mutating func unlock() {
+		if let lock = _lock {
+			return lock.unlock()
+		}
+
+		isSendingNonatomic = false
+	}
+
+	mutating func `try`() -> Bool {
+		if let lock = _lock {
+			return lock.try()
+		}
+
+		return !isSendingNonatomic
+	}
+}
+
+private func ===(left: SendLock?, right: SendLock?) -> Bool {
+	if let left = left, let right = right, left._lock === right._lock {
+		return true
+	}
+	return false
+}
+
 extension Signal {
 	/// A Signal that never sends any events to its observers.
 	public static var never: Signal {
@@ -509,8 +567,12 @@ extension Signal {
 	/// - returns: A tuple of `output: Signal`, the output end of the pipe,
 	///            and `input: Observer`, the input end of the pipe.
 	public static func pipe(disposable: Disposable? = nil) -> (output: Signal, input: Observer) {
+		return pipe(attributes: .default, disposable: disposable)
+	}
+
+	internal static func pipe(attributes: Serialization, disposable: Disposable? = nil) -> (output: Signal, input: Observer) {
 		var observer: Observer!
-		let signal = self.init { innerObserver in
+		let signal = self.init(attributes) { innerObserver in
 			observer = innerObserver
 			return disposable
 		}
@@ -639,7 +701,7 @@ extension Signal {
 	///
 	/// - returns: A signal that will send new values.
 	public func map<U>(_ transform: @escaping (Value) -> U) -> Signal<U, Error> {
-		return Signal<U, Error> { observer in
+		return Signal<U, Error>(.nonatomic) { observer in
 			return self.observe { event in
 				observer.action(event.map(transform))
 			}
@@ -654,7 +716,7 @@ extension Signal {
 	///
 	/// - returns: A signal that will send new type of errors.
 	public func mapError<F>(_ transform: @escaping (Error) -> F) -> Signal<Value, F> {
-		return Signal<Value, F> { observer in
+		return Signal<Value, F>(.nonatomic) { observer in
 			return self.observe { event in
 				observer.action(event.mapError(transform))
 			}
@@ -691,7 +753,7 @@ extension Signal {
 	///
 	/// - returns: A signal that forwards the values passing the given closure.
 	public func filter(_ isIncluded: @escaping (Value) -> Bool) -> Signal<Value, Error> {
-		return Signal { observer in
+		return Signal(.nonatomic) { observer in
 			return self.observe { (event: Event) -> Void in
 				guard let value = event.value else {
 					observer.action(event)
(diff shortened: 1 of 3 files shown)


Sources/Signal.swift
```diff
@@ -49,15 +51,45 @@ public final class Signal<Value, Error: Swift.Error> {
...
+					lock.lock()
+				}
+			}
+
+			@inline(__always)
+			func sendLockTryLock() -> Bool {
+				if let lock = sendLock {
+					return lock.try()
+				}
+				return true
```
sharplet: This could be shortened to `return lock?.try() ?? true`

Sources/Signal.swift
```diff
@@ -49,15 +51,45 @@ public final class Signal<Value, Error: Swift.Error> {
...
+			}
+
+			@inline(__always)
+			func sendLockUnlock() {
+				if let lock = sendLock {
+					lock.unlock()
+				}
```
sharplet: This could be shortened to `lock?.unlock()`

Sources/Signal.swift
```diff
@@ -49,15 +51,45 @@ public final class Signal<Value, Error: Swift.Error> {
...
+			sendLock!.name = "org.reactivecocoa.ReactiveSwift.Signal"
+		}
 
 		let observer = Observer { [weak self] event in
 			guard let signal = self else {
 				return
 			}
 
 			@inline(__always)
+			func sendLockLock() {
```
sharplet: What do you think about just naming this `lock()`?

Sources/Signal.swift
```diff
@@ -49,15 +51,45 @@ public final class Signal<Value, Error: Swift.Error> {
...
+				if let lock = sendLock {
+					lock.lock()
+				}
+			}
+
+			@inline(__always)
+			func sendLockTryLock() -> Bool {
```
sharplet: What do you think about just naming this `tryLock()`?

Sources/Signal.swift
```diff
@@ -49,15 +51,45 @@ public final class Signal<Value, Error: Swift.Error> {
...
+				if let lock = sendLock {
+					return lock.try()
+				}
+				return true
+			}
+
+			@inline(__always)
+			func sendLockUnlock() {
```
sharplet: What do you think about just naming this `unlock()`?

Sources/Signal.swift
```diff
@@ -187,13 +219,16 @@ public final class Signal<Value, Error: Swift.Error> {
...
 	/// - parameters:
+	///   - contained: Indicates if the event emitter passed to `generator` would
+	///                be contained in a serialized routine. By default, it is
+	///                `false`, and should be used with cautions.
 	///   - disposable: An optional disposable to associate with the signal, and
 	///                 to be disposed of when the signal terminates.
 	///
 	/// - returns: A tuple made of signal and observer.
-	public static func pipe(disposable: Disposable? = nil) -> (Signal, Observer) {
+	public static func pipe(_ concurrency: SignalConcurrency = .serialized, disposable: Disposable? = nil) -> (Signal, Observer) {
```
sharplet: I think this should be a named parameter.

Sources/Signal.swift
```diff
@@ -35,9 +35,11 @@ public final class Signal<Value, Error: Swift.Error> {
 	///         Signal itself will remain alive until the observer is released.
 	///
 	/// - parameters:
+	///   - concurrency: The concurrency mode the signal should assume.
+	///                  By default, it is `serialized`.
 	///   - generator: A closure that accepts an implicitly created observer
 	///                that will act as an event emitter for the signal.
-	public init(_ generator: (Observer) -> Disposable?) {
+	public init(_ concurrency: SignalConcurrency = .serialized, _ generator: (Observer) -> Disposable?) {
```
sharplet: I think this should be a named parameter.

Sources/Signal.swift
```diff
@@ -237,6 +272,17 @@ public final class Signal<Value, Error: Swift.Error> {
...
 
+/// The concurrency mode of a `Signal`.
+public enum SignalConcurrency {
+	/// The `Signal` should serialize all the received events.
+	case serialized
+
+	/// The `Signal` may assume the events emitted to it have already been 
+	/// serialized. In other words, the event emitter of the `Signal` is assumed
+	/// to only be called in synchronized routines.
+	case contained
```
sharplet: I'm not sure about the naming — how about this?

```swift
public enum SignalSerialization {
  case mutex
  case inherit
}

Signal(serialization: .mutex) { observer in
  // ...
}
Signal.pipe(serialization: .inherit)
```

Sources/Signal.swift
```diff
@@ -187,13 +219,16 @@ public final class Signal<Value, Error: Swift.Error> {
 	///         retained.
 	///
 	/// - parameters:
+	///   - contained: Indicates if the event emitter passed to `generator` would
```
sharplet: This comment will need to be updated when the name is finalised.

Sources/Signal.swift
```diff
@@ -35,9 +35,11 @@ public final class Signal<Value, Error: Swift.Error> {
 	///         Signal itself will remain alive until the observer is released.
 	///
 	/// - parameters:
+	///   - concurrency: The concurrency mode the signal should assume.
```
sharplet: This comment will need to be updated too.

andersio:
> Are there any ramifications for existing code that might start signals as a side effect in on() or some other operator?

If you mean starting producers, I don't see it would cause harm. If you mean observing any `Signal` along the chain, the observer bag is CoW on insertion or removal.

> What would happen to recursive event sending that currently results in a deadlock — anything?

Obviously, the unlocked mode would grant a hole for recursive event sending. But our stance should remain unchanged. One way we could do this is allowing lock elision only in release builds.

As for recursive `interrupted` handling, since we are just dealing with `Signal`s here, the only source of `interrupted` events would be the upstream, i.e. no difference from other events.

andersio:
@NachoSoto Has to be part of public, or otherwise RAC can't use them.

andersio:
Renamed a few stuff according to @sharplet's suggestions, and had four operators joined their lock-elided friends in https://github.com/ReactiveCocoa/ReactiveSwift/pull/129/commits/1dd6d5f44df51e295f2a9124d623bac411f5b236.

andersio:
A note on the recursive `interrupted` handling:

AFAIU, it is made for `SignalProducer` specifically. However, as we are not addressing `SignalProducer` yet, it is simply assumed to never happen with inherited serialization.

```
 lock
 observers ---> elidedLock
                observers    ---> lock
                elidedUnlock      observers    ---> elidedLock
                                  unlock            observers    ---> ...
                                                    elidedUnlock
 unlock
-------------------------------------------------------------------------
   PRODUCER   |   LIFTED MAP    |    PRODUCER     |   LIFTED MAP    | ...
-------------------------------------------------------------------------
```

Sources/Signal.swift
```diff
@@ -237,6 +260,16 @@ public final class Signal<Value, Error: Swift.Error> {
...
 }
 
+/// Describes how `Signal` should serialize the events it received.
+public enum SignalSerialization {
+	/// The `Signal` should serialize all the received events.
+	case `default`
+
+	/// The `Signal` assumes to have inherited the serialization from its owner,
+	/// and its event emitter would not be called concurrently.
+	case inherited
```
andersio: Followed `FlattenStrategy` to use imperative verbs.

Sources/FoundationExtensions.swift
```diff
@@ -29,7 +29,7 @@ extension Reactive where Base: NotificationCenter {
 	/// - note: The signal does not terminate naturally. Observers must be
 	///         explicitly disposed to avoid leaks.
 	public func notifications(forName name: Notification.Name?, object: AnyObject? = nil) -> Signal<Notification, NoError> {
-		return Signal { [base = self.base] observer in
+		return Signal(strategy: .inherit) { [base = self.base] observer in
```
sharplet: I think we've lost too much meaning now at the point of use. It's no longer clear what's being inherited.

sharplet: I'd say `serializationStrategy`.

However, I'm still partial to `serialization: .default` and `serialization: .inherit`. What do you think?

sharplet: ...with declaration: `init(serialization strategy: SerializationStrategy, ...)`.

andersio: Sometimes I do wonder if `FlattenStrategy` should be `FlatteningStrategy` instead...

andersio:
@NachoSoto Maybe I should phrase it better: It can for sure be private to ReactiveSwift.

I just want to figure out a middle ground for RAC if it makes sense. If "internal through obscurity" is still considered exposing the implementation, there is probably no other way to expose this to RAC based on my understanding.

andersio:
Blocked by #137.

andersio:
@NachoSoto Hmm, I actually thought of a counter argument to keeping it private: while most users would not care, users - including depending libraries like ReactiveCocoa - who want custom operators or have all possible optimization should be given the tools.

If what you worry is having it as part of the normal overload, as mentioned we could export it in an separate enum namespace.

andersio:
[Benchmark Source](https://gist.github.com/andersio/8b055cccf26ad9d2c0db58c1bf11c40c)
Environment: macOS Sierra 10.12.1, i5 4258U, Xcode 8.1, `-Owholemodule`

**TableView Scrolling Simulation**
_Producer heavy, Produced signals send only 1 value._
+2% (7%)

**Send 10000 values to 1 observer.**
+9% (10%)

**Send 10000 values to 8 observers.**
+12% (5%)

**Send 10000 values to 1 produced signal.**
+4% (8%)

**Send 10000 values to 8 produced signals.**
+6% (3%)

andersio:
Blocked by #137.

andersio:
The `SignalSerialization` enum is replaced with an option set, since I expect #163 would need a similar point of configuration. The APIs are also promoted back to be `public`, primarily for depending frameworks.

Note that as a default attribute set is provided, it shows up in autocomplete as two overloads.

andersio:
Rolled back to be an internal API, and rebased on `master`. The benefit of opening it up is trivial, since its primary concern is deep signal composition.

Sources/Signal.swift
```diff
@@ -106,16 +106,21 @@ public final class Signal<Value, Error: Swift.Error> {
...
-		private let sendLock: Lock
+		private var sendLock: SendLock
 
 		/// Used to indicate if the `Signal` has deinitialized.
 		private var hasDeinitialized: Bool
 
-		fileprivate init(_ generator: (Observer) -> Disposable?) {
+		fileprivate init(_ attributes: SignalAttributes, _ generator: (Observer) -> Disposable?) {
+			var attributes = attributes
+			if (Thread.current.threadDictionary["reactiveswift_disable_inherit_serialization"] as! Bool?) ?? false {
```
mdiep: What's this for?

Sources/Signal.swift
```diff
@@ -481,6 +500,63 @@ public final class Signal<Value, Error: Swift.Error> {
...
+
+private func ===(left: SendLock?, right: SendLock?) -> Bool {
+	if let left = left, let right = right, left._lock === right._lock {
+		return true
+	}
+	return false
+}
+
+/// Describes how `Signal` should behave.
+internal struct SignalAttributes: OptionSet {
```
mdiep: Why an `OptionSet` instead of an `enum`?

Sources/Signal.swift
andersio: Ah, I forgot to revert this hack for benchmarking purpose. 🤦‍♂️

andersio:
> The thread safety is already hard to reason about without adding an additional level of complexity.

The annotation in this PR is built on top of the assumption of `Signal` contract, saying everything must be serialized. If a synchronous operator derives a `Signal` from one single upstream, the derived `Signal` is implicitly serialized. The keywords are _synchronous_ and _single upstream_, which are fairly explicit IMO.

andersio:
Closed since this becomes trivial with an upcoming higher-level `SignalProducer` optimisation branch.

PR #130, opened 2016-11-29
Only build pushes to master and PRs on CI

sharplet:
Every push to a PR currently triggers two builds on CI, which is pretty wasteful (not to mention slow). Hoping this cuts down on the duplication. *fingers crossed*


sharplet:
\o/

PR #131, opened 2016-11-29
BindingSourceProtocol

mdiep:
The idea here is to eliminate redundancy in things like https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3289.

By adding a `BindingSourceProtocol` type, both ends of `<~` become generic. That makes it easier to add conformance on either end.

This could also be called `Observable`—if we want to make it some more generic.

I'm curious what people think of this approach.

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -210,7 +210,7 @@ public final class Signal<Value, Error: Swift.Error> {
 	///   - observer: An observer to forward the events to.
 	///
 	/// - returns: A `Disposable` which can be used to disconnect the observer,
-	///            or `nil` if the signal has already completed.
+	///            or `nil` if the signal has already terminated.
 	@discardableResult
 	public func observe(_ observer: Observer) -> Disposable? {
 		var token: RemovalToken?
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -209,8 +209,8 @@ public final class Signal<Value, Error: Swift.Error> {
 	/// - parameters:
 	///   - observer: An observer to forward the events to.
 	///
-	/// - returns: An optional `Disposable` which can be used to disconnect the
-	///            observer.
+	/// - returns: A `Disposable` which can be used to disconnect the observer,
+	///            or `nil` if the signal has already completed.
 	@discardableResult
 	public func observe(_ observer: Observer) -> Disposable? {
 		var token: RemovalToken?
--- a/Sources/UnidirectionalBinding.swift
+++ b/Sources/UnidirectionalBinding.swift
@@ -16,7 +16,7 @@ public protocol BindingSourceProtocol {
 	associatedtype Value
 	associatedtype Error: Swift.Error
 	
-	/// Observe the binding source by sending any evenst to the given observer.
+	/// Observe the binding source by sending any events to the given observer.
 	@discardableResult
 	func observe(_ observer: Observer<Value, Error>) -> Disposable?
 }
@@ -49,11 +49,11 @@ public protocol BindingTargetProtocol: class {
 	func consume(_ value: Value)
 }
 
-/// Binds a signal to a target, updating the target's value to the latest
-/// value sent by the signal.
+/// Binds a source to a target, updating the target's value to the latest
+/// value sent by the source.
 ///
 /// - note: The binding will automatically terminate when the target is
-///         deinitialized, or when the signal sends a `completed` event.
+///         deinitialized, or when the source sends a `completed` event.
 ///
 /// ````
 /// let property = MutableProperty(0)
@@ -73,10 +73,10 @@ public protocol BindingTargetProtocol: class {
 ///
 /// - parameters:
 ///   - target: A target to be bond to.
-///   - signal: A signal to bind.
+///   - source: A source to bind.
 ///
 /// - returns: A disposable that can be used to terminate binding before the
-///            deinitialization of the target or the signal's `completed`
+///            deinitialization of the target or the source's `completed`
 ///            event.
 @discardableResult
 public func <~
@@ -92,10 +92,10 @@ public func <~
 }
 
 /// Binds a source to a target, updating the target's value to the latest
-/// value sent by the signal.
+/// value sent by the source.
 ///
 /// - note: The binding will automatically terminate when the target is
-///         deinitialized, or when the signal sends a `completed` event.
+///         deinitialized, or when the source sends a `completed` event.
 ///
 /// ````
 /// let property = MutableProperty(0)
@@ -115,10 +115,10 @@ public func <~
 ///
 /// - parameters:
 ///   - target: A target to be bond to.
-///   - signal: A signal to bind.
+///   - source: A source to bind.
 ///
 /// - returns: A disposable that can be used to terminate binding before the
-///            deinitialization of the target or the signal's `completed`
+///            deinitialization of the target or the source's `completed`
 ///            event.
 @discardableResult
 public func <~
--- a/Sources/UnidirectionalBinding.swift
+++ b/Sources/UnidirectionalBinding.swift
@@ -16,6 +16,7 @@ public protocol BindingSourceProtocol {
 	associatedtype Value
 	associatedtype Error: Swift.Error
 	
+	/// Observe the binding source by sending any evenst to the given observer.
 	@discardableResult
 	func observe(_ observer: Observer<Value, Error>) -> Disposable?
 }
--- a/Sources/UnidirectionalBinding.swift
+++ b/Sources/UnidirectionalBinding.swift
@@ -11,7 +11,45 @@ precedencegroup BindingPrecedence {
 
 infix operator <~ : BindingPrecedence
 
-/// Describes a target to which can be bound.
+/// Describes a source which can be bound.
+public protocol BindingSourceProtocol {
+	associatedtype Value
+	associatedtype Error: Swift.Error
+	
+	@discardableResult
+	func observe(_ observer: Observer<Value, Error>, lifetime: Lifetime) -> Disposable?
+}
+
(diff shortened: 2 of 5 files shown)


Sources/UnidirectionalBinding.swift
```diff
@@ -11,167 +11,61 @@ precedencegroup BindingPrecedence {
...
-	/// ````
-	///
-	/// - parameters:
-	///   - target: A target to be bond to.
-	///   - signal: A signal to bind.
-	///
-	/// - returns: A disposable that can be used to terminate binding before the
-	///            deinitialization of the target or the signal's `completed`
-	///            event.
+	associatedtype Error: Swift.Error
```
mdiep: This would need an error type until Swift 4 since we can't add conditional conformance when `Error == NoError`.

Sources/UnidirectionalBinding.swift
```diff
@@ -11,167 +11,61 @@ precedencegroup BindingPrecedence {
...
+	/// The lifetime of `self`. The binding operators use this to determine when
+	/// the binding should be teared down.
+	var lifetime: Lifetime { get }
+
+	/// Consume a value from the binding.
+	func consume(_ value: Value)
+}
+
+@discardableResult
+public func <~ <Target: BindingTargetProtocol, Source: BindingSourceProtocol>(target: Target, source: Source) -> Disposable? where Source.Value == Target.Value, Source.Error == NoError {
```
andersio: Could this be defined in an extension to `BindingTargetProtocol`?

Sources/UnidirectionalBinding.swift
```diff
@@ -11,167 +11,61 @@ precedencegroup BindingPrecedence {
...
+		return producer.observe(observer, lifetime: lifetime)
 	}
 }
 
+/// Describes a target which can be bound.
+public protocol BindingTargetProtocol: class {
+	associatedtype Value
+
+	/// The lifetime of `self`. The binding operators use this to determine when
+	/// the binding should be teared down.
```
sharplet: This should read "torn down"

Sources/UnidirectionalBinding.swift
```diff
@@ -11,167 +11,61 @@ precedencegroup BindingPrecedence {
...
-	///   - signal: A signal to bind.
-	///
-	/// - returns: A disposable that can be used to terminate binding before the
-	///            deinitialization of the target or the signal's `completed`
-	///            event.
+	associatedtype Error: Swift.Error
+	
 	@discardableResult
-	static func <~ <Source: SignalProtocol>(target: Self, signal: Source) -> Disposable? where Source.Value == Value, Source.Error == NoError
+	func observe(_ observer: Observer<Value, Error>, lifetime: Lifetime) -> Disposable?
```
sharplet: Why an optional Disposable?

andersio: `Signal.observe` returns an optional.

Sources/UnidirectionalBinding.swift
mdiep: Since it's generic over both sides, I think it makes sense to define it at the top level.

Sources/UnidirectionalBinding.swift
```diff
@@ -11,272 +11,126 @@ precedencegroup BindingPrecedence {
 
 infix operator <~ : BindingPrecedence
 
-/// Describes a target to which can be bound.
-public protocol BindingTargetProtocol: class {
+/// Describes a source which can be bound.
+public protocol BindingSourceProtocol {
```
andersio: Should it be a `class` bound protocol?

Sources/UnidirectionalBinding.swift
```diff
@@ -11,272 +11,126 @@ precedencegroup BindingPrecedence {
...
-		return target <~ producer.map(Value.init(reconstructing:))
+	/// The lifetime of `self`. The binding operators use this to determine when
+	/// the binding should be torn down.
+	var lifetime: Lifetime { get }
+
+	/// Consume a value from the binding.
+	func consume(_ value: Value)
+}
+
+/// Binds a signal to a target, updating the target's value to the latest
```
andersio: Please update the documentation.

Sources/UnidirectionalBinding.swift
```diff
@@ -11,272 +11,126 @@ precedencegroup BindingPrecedence {
...
-	/// - parameters:
-	///   - target: A target to be bond to.
-	///   - property: A property to bind.
-	///
-	/// - returns: A disposable that can be used to terminate binding before the
-	///            deinitialization of the target or the source property.
-	@discardableResult
-	public static func <~ <Source: PropertyProtocol>(target: Self, property: Source) -> Disposable where Source.Value == Value.Wrapped {
-		return target <~ property.producer
+/// Binds a source to a target, updating the target's value to the latest
```
andersio: Ditto.

Sources/UnidirectionalBinding.swift
```diff
@@ -11,272 +11,126 @@ precedencegroup BindingPrecedence {
...
-	/// - parameters:
-	///   - target: A target to be bond to.
-	///   - signal: A signal to bind.
-	///
-	/// - returns: A disposable that can be used to terminate binding before the
-	///            deinitialization of the target or the signal's `completed`
-	///            event.
+	associatedtype Error: Swift.Error
+	
+	/// Observe the binding source by sending any evenst to the given observer.
```
andersio: Typo: `evenst`.

Sources/Signal.swift
```diff
@@ -209,8 +209,8 @@ public final class Signal<Value, Error: Swift.Error> {
 	/// - parameters:
 	///   - observer: An observer to forward the events to.
 	///
-	/// - returns: An optional `Disposable` which can be used to disconnect the
-	///            observer.
+	/// - returns: A `Disposable` which can be used to disconnect the observer,
+	///            or `nil` if the signal has already completed.
```
andersio: [Nit] "already terminated" sounds more accurate.

Sources/UnidirectionalBinding.swift
```diff
@@ -312,15 +166,6 @@ public final class BindingTarget<Value>: BindingTargetProtocol {
...
-	public static func <~ <Source: SignalProtocol>(target: BindingTarget<Value>, signal: Source) -> Disposable? where Source.Value == Value, Source.Error == NoError {
-		return signal
-			.take(during: target.lifetime)
-			.observeValues { [setter = target.setter] value in
-				setter(value)
-			}
-	}
 }
 
 private let specificKey = DispatchSpecificKey<ObjectIdentifier>()
```
andersio: This seems unused.

Sources/UnidirectionalBinding.swift
```diff
@@ -312,15 +166,6 @@ public final class BindingTarget<Value>: BindingTargetProtocol {
 	public func consume(_ value: Value) {
 		setter(value)
 	}
-
-	@discardableResult
-	public static func <~ <Source: SignalProtocol>(target: BindingTarget<Value>, signal: Source) -> Disposable? where Source.Value == Value, Source.Error == NoError {
-		return signal
-			.take(during: target.lifetime)
-			.observeValues { [setter = target.setter] value in
-				setter(value)
```
andersio: This should not be removed. It is intentionally override to capture `setter` directly, so that `self` need not be retained and/or uniqued by its parent. The default implementation weakly captures the target, so the binding would be invalid.

_(Seems like it wasn't covered in the spec.)_

Sources/UnidirectionalBinding.swift
mdiep: `SignalProducer` isn't a class, which is why this isn't class bound.

andersio: Oops, overlooked that.

Sources/UnidirectionalBinding.swift
mdiep: I added a test for this in 1c339803382129ad53b5b9cfa0d74f1045629a0c.

Sources/UnidirectionalBinding.swift
```diff
@@ -73,29 +73,39 @@ public protocol BindingTargetProtocol: class {
...
 	<Target: BindingTargetProtocol, Source: BindingSourceProtocol>
 	(target: Target, source: Source) -> Disposable?
 	where Source.Value == Target.Value, Source.Error == NoError
 {
-	let disposable = source.observe(Observer(value: { [weak target] in target?.consume($0) }))
+	// Alter the semantics of `BindingTarget` to not require it to be retained.
+	// This is done here--and not in a separate function--so that all variants
+	// of `<~` can get this behavior.
+	let observer: Observer<Source.Value, NoError>
+	if let target = target as? BindingTarget<Source.Value> {
```
andersio: IIRC Swift will pick the most specific overload. So instead of a dynamic check, you might as well define an overload: 
```
@discardableResult
  public func <~
  	<Value, Source: BindingSourceProtocol>
  	(target: BindingTarget<Value>, source: Source) -> Disposable?
  	where Source.Value == Target.Value, Source.Error == NoError
```
Though the alternative is still having it defined in the protocol, and overriding it in `BindingTarget`.

mdiep: The issue I'm nervous about is if I want to constrain the _righthand_ side to be something else.

```swift
public func <~
     <Target: BindingTargetProtocol>
     (target: Target, source: FooSource) -> Disposable?
     where Target.Value == FooSource.Value
{
    … // This would call to another <~
}
```

But maybe that'd need to filter through the overload anyway?

andersio: I am not sure about the behavior. But there seems no point to assume the need of custom binding logic for varying types of sources.

Though I believe if you define it in the protocol, the matching priority would likely be given to the target.

PR #133, opened 2016-11-29
Use AnyError for `URLSession.reactive.data(with:)`

ikesyo:
https://github.com/antitypical/Result/pull/198

### DIFF
--- a/.Package.test.swift
+++ b/.Package.test.swift
@@ -3,7 +3,7 @@ import PackageDescription
 let package = Package(
     name: "ReactiveSwift",
     dependencies: [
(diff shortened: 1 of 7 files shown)


sharplet:
Is there a way we can provide a deprecation warning for this change? I'm unsure what the impact of this change would be on code expecting `NSError`.

ikesyo:
> Is there a way we can provide a deprecation warning for this change?

Something like this? 😉

sharplet:
Looks good

mdiep:
Why isn't `NSError` correct here? Isn't that what the Foundation API will have returned?

sharplet:
It's bridged as `Error`, so I think it makes us a better Swift 3 citizen to preserve that as accurately as possible.

PR #134, opened 2016-11-29
[Nit] Rename `FlattenStrategy` to `FlatteningStrategy`.

andersio:
One Java framework and one Scala framework use `FlattenStrategy`, while Rx and ReactiveObjC flattened the flattening operators.

👮


andersio:
So If another `Strategy` is to be made, should it follow this to be `[verb]Strategy`?

mdiep:
Yeah, but I'd probably frame it as `[operator]Strategy`.

Issue #135, opened 2016-11-29
Deadlock in binding to a MutableProperty


sharplet:
I'm seeing a deadlock in a binding from a signal to a mutable property. Here's the relevant part of the stack trace:

![image](https://cloud.githubusercontent.com/assets/147686/20729626/d5d6dc8c-b650-11e6-8fec-940b170009d9.png)

(2) and (19) are the same signal, hitting the recursive send. Here's the current implementation of `<~`:

```swift
	public static func <~ <Source: SignalProtocol>(target: Self, signal: Source) -> Disposable? where Source.Value == Value, Source.Error == NoError {
		return signal
			.take(during: target.lifetime)
			.observeValues { [weak target] value in
				target?.consume(value)
			}
	}
```

I _think_ what's happening is:

1. Signal at (19) sends a value
2. The `observeValues` block is called with the value
3. Upon reading `target`, which has gone out of scope, the target's `deinit` gets called
4. This triggers `lifetime.ended` to complete
5. Because of `take(during:)`, the signal immediately tries to complete (lifetime has ended)
6. Deadlock, because the same signal is in the middle of sending a value

Unfortunately I haven't been able to write a test case that reproduces this.

andersio:
Not sure if it is related: https://github.com/ReactiveCocoa/ReactiveSwift/issues/92#issuecomment-259833507.

But this should not be happening if the binding is truly the last weak reference to the `MutableProperty`.

PR #136, opened 2016-11-30
Failing Test Case: Delayed `Lifetime` deinit deadlocking `sendLock`. (#137)

andersio:
Related issue: #135 

#### Cause
1. A strong release atomically drops the refCount to 0. But deinitialization does not start right away.
2. A weak retain - the binding's observer closure in this case - manages to increment the refCount before the strong release flags the object as deallocating.
3. The deinitialization is thus delayed until exiting the scope of the observer closure, i.e. within the `sendLock`.

#### What's affected?
Everything in the following pattern:
```swift
signal
    .take(during: object.lifetime)
    .observeValues { [weak object] _ in }
```

#### Note
The test case would not fail in a uniprocessor environment.

#### Possible Solution
Allowing `completed` to be recursively sent like `interrupted` seems the only way, since:
1. We have no `autoreleasepool` for Swift objects.
2. The order cannot be controlled.

### DIFF
--- a/Tests/ReactiveSwiftTests/PropertySpec.swift
+++ b/Tests/ReactiveSwiftTests/PropertySpec.swift
@@ -265,6 +265,39 @@ class PropertySpec: QuickSpec {
 				property = nil
 				expect(isEnded) == true
 			}
+
+			it("should not deadlock") {
+				let queue: DispatchQueue
+
+				if #available(macOS 10.10, *) {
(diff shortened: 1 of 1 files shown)


andersio:
`NSRecursiveLock` unfortunately seems the easiest solution to this race condition. Handling `completed` like `interrupted` is like opening Pandora's box, racing with Swift's deallocation routines concurrently, and then having malloc broken or polluted somehow.

PR #137, opened 2016-11-30
Allow all termination events to be recursively sent.

### LINKED ISSUE #135: Deadlock in binding to a MutableProperty

andersio:
Fixes #135 and #136.

As described in #136, and demonstrated by [the test case](https://github.com/ReactiveCocoa/ReactiveSwift/pull/137/commits/2d239c50e1b97b95874f5c52e9cdc43a5d11a299):
> 1. A strong release atomically drops the refCount to 0. But deinitialization does not start right away.
> 2. A weak retain - the binding's observer closure in this case - manages to increment the refCount before the strong release flags the object as deallocating.
> 3. The deinitialization is thus delayed until exiting the scope of the observer closure, i.e. within the sendLock.

In the context of binding:
```swift
signal
    .take(during: target.lifetime)
    .observeValues { [weak target] value in
        target?.consume(value)
    }
```

```
             Thread 1             |            Thread 2            
----------------------------------|--------------------------------
  1. Received new `value` event.  |
                                  |
  2. Acquired `sendLock`.         |
                                  |
                                  | 3. `swift_release` on `target`.
                                  |    Decrement refCount to 0.
  4. `swift_weakLoadStrong`       |
     on `target`.                 |
     Increment refCount to 1.     |
                                  |     
  5a. Invoke `consume`.           | 5b. CAS to mark deallocating
                                  |     flag failed
                                  |
  6. `swift_release`              | (Deinit of `target` is supposed
     Decrement refCount to 0.     |  to run on this thread, but
                                  |  Thread 1 has stolen it due to
  7. CAS to mark deallocating     |  the weak retain having won the
     Flag succeeded.              |  race.)
                                  |
  8. `target` starts deinit.      |
     `Lifetime.init` sends        |
     `completed`.                 |
                                  |
  9. Upstream `take(until:)`      |
     gets `completed`, and forward|
     it to the downstream.        |
                                  |
  10. Try to acquire `sendLock`,  |
      but it is held by (2).      |
      Deadlock.                   |
```

So the only solution - except using a recursive lock or going asynchronous - is to handle `completed` like `interrupted`. `failed` do not need this behavior (yet?), but it is more convenient to implement it for all termination events. Can be reverted if this feels an unpleasant change.

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -27,8 +27,7 @@ public final class Signal<Value, Error: Swift.Error> {
 	/// The state of the signal. `nil` if the signal has terminated.
 	private let state: Atomic<SignalState<Value, Error>?>
 
-	/// Holds the status of the signal. Implicitly protected by `state` and
-	/// `sendLock`.
+	/// The status of the signal. It synchronizes with `sendLock`.
 	private var status: SignalStatus<Value, Error>
 
 	/// Used to ensure that events are serialized during delivery to observers.
@@ -46,7 +45,7 @@ public final class Signal<Value, Error: Swift.Error> {
 	///                that will act as an event emitter for the signal.
 	public init(_ generator: (Observer) -> Disposable?) {
 		state = Atomic(SignalState())
-		status = .green
+		status = .alive
 		sendLock = NSLock()
 		sendLock.name = "org.reactivecocoa.ReactiveSwift.Signal"
 
@@ -79,22 +78,23 @@ public final class Signal<Value, Error: Swift.Error> {
 					// So we do not need to guard it with an explicit lock.
 					//
 					// Specifically, senders are serialized by `sendLock`, and can react
-					// to and bump the status to `red` only after they see the status
-					// being `yellow`.
+					// to and bump the status to `terminated` only after they see the
+					// status being `terminating`.
 					//
-					// `yellow` can happen only once, since `state` being swapped with
-					// `nil` is a point of no return.
+					// `terminating` can happen only once in the lifetime of a `Signal`,
+					// since `state` being swapped with `nil` is a point of no return.
 					//
 					// Even in the case that both a previous sender and its successor see
-					// `yellow`, the senders are still bound to the `sendLock`. So
+					// `terminating`, the senders are still bound to the `sendLock`. So
 					// whichever sender loses the battle of acquring `sendLock` is
-					// guaranteed to be blocked by a status of `red`.
-					signal.status = .yellow(SignalTerminationSnapshot(state: state, event: event))
+					// guaranteed to see a status of `terminated`, and eventually be
+					// blocked.
+					signal.status = .terminating(SignalTerminationSnapshot(state: state, event: event))
 
 					if signal.sendLock.try() {
-						if case let .yellow(state) = signal.status {
+						if case let .terminating(state) = signal.status {
 							state.send()
-							signal.status = .red
+							signal.status = .terminated
 						}
 						signal.sendLock.unlock()
 						signal.swapDisposable()?.dispose()
@@ -105,36 +105,35 @@ public final class Signal<Value, Error: Swift.Error> {
 
 				signal.sendLock.lock()
 
-				if case .green = signal.status {
+				if case .alive = signal.status {
 					for observer in observers {
 						observer.action(event)
 					}
 
 					// Check if a downstream consumer or a concurrent sender has
 					// interrupted the signal.
-					if case let .yellow(state) = signal.status {
+					if case let .terminating(state) = signal.status {
 						state.send()
-						signal.status = .red
+						signal.status = .terminated
 						shouldDispose = true
 					}
 				}
 
 				signal.sendLock.unlock()
 
-				// Based on the implicit memory order, any updates to the
-				// `terminationSnapshot` should always be visible after `sendLock` is
-				// released. So we check it again and handle the termination if
-				// it has not been taken over.
-				if !shouldDispose, case let .yellow(state) = signal.status {
+				// Based on the implicit memory order, any updates to `status` should
+				// always be visible after `sendLock` is released. So we check it again
+				// and handle the termination if it has not been taken over.
+				if !shouldDispose, case .terminating = signal.status {
 					signal.sendLock.lock()
 
-					// The status check before acquring `sendLock` could be a false
+					// The `terminating` check before acquring `sendLock` could be a false
 					// positive, since it might race against other concurrent senders
 					// until the lock acquisition above succeeds. So we have to check
-					// again if the signal is still at `yellow`.
-					if case .yellow = signal.status {
+					// again if the status is still `terminating`.
+					if case let .terminating(state) = signal.status {
 						state.send()
-						signal.status = .red
+						signal.status = .terminated
 						shouldDispose = true
 					}
 
@@ -152,6 +151,10 @@ public final class Signal<Value, Error: Swift.Error> {
 		generatorDisposable = generator(observer)
 	}
 
+	/// Swap the generator disposable with `nil`.
+	///
+	/// - returns:
+	///   The generator disposable, or `nil` if it has been disposed of.
 	@inline(__always)
 	private func swapDisposable() -> Disposable? {
 		if let d = generatorDisposable {
@@ -241,14 +244,14 @@ public final class Signal<Value, Error: Swift.Error> {
 /// The status of a `Signal`.
 private enum SignalStatus<Value, Error: Swift.Error> {
 	/// The `Signal` is alive.
-	case green
+	case alive
 
 	/// The `Signal` has received a termination event, and is about to be
 	/// terminated.
-	case yellow(SignalTerminationSnapshot<Value, Error>)
+	case terminating(SignalTerminationSnapshot<Value, Error>)
 
 	/// The `Signal` has terminated.
-	case red
+	case terminated
 }
 
 /// A snapshot of the state of a `Signal` when it receives a termination event.
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -94,6 +94,13 @@ public final class Signal<Value, Error: Swift.Error> {
 			//   also be checked again after the locks are acquired. Fail gracefully
 			//   if the state has changed since the relaxed read, i.e. a concurrent
 			//   sender has already handled the termination event.
+			//
+			// Exploiting the relaxation of reads, please note that false positives
+			// are intentionally allowed in the `terminating` checks below. As a
+			// result, normal event deliveries need not acquire `updateLock`.
+			// Nevertheless, this should not cause the termination event being
+			// sent multiple times, since `shouldReallyTerminate` would not
+			// respond to false positives.
 
 			if event.isTerminating {
 				// Recursive events are disallowed for `value` events, but are permitted
@@ -124,21 +131,12 @@ public final class Signal<Value, Error: Swift.Error> {
 					signal.updateLock.unlock()
 
 					if signal.sendLock.try() {
-						// Acquire `updateLock`. If the termination has still not yet been
-						// handled, take it over and bump the status to `terminated`.
-						signal.updateLock.lock()
-
-						// It is either `terminating` with `newSnapshot` or `terminated`
-						// anyway. So retrieving the snapshot is unnecessary.
-						if case .terminating = signal.state {
-							signal.state = .terminated
-							signal.updateLock.unlock()
-
-							for observer in newSnapshot.observers {
-								observer.action(newSnapshot.associated)
+						// Check whether the termination event has been handled by a
+						// concurrent sender. If not, handle it.
+						if let snapshot = signal.shouldReallyTerminate() {
+							for observer in snapshot.observers {
+								observer.action(snapshot.associated)
 							}
-						} else {
-							signal.updateLock.unlock()
 						}
 
 						signal.sendLock.unlock()
@@ -178,27 +176,12 @@ public final class Signal<Value, Error: Swift.Error> {
 
 					// Check if the status has been bumped to `terminating` due to a
 					// concurrent or a recursive termination event.
-					//
-					// Note that as the check happens before acquiring `updateLock`, it
-					// could be a false positive, and thus a second check after the lock
-					// is acquired is mandatory.
-					if case .terminating = signal.state {
-						// Acquire `updateLock`. If the termination has still not yet been
-						// handled, take it over and bump the status to `terminated`.
-						signal.updateLock.lock()
-
-						if case let .terminating(snapshot) = signal.state {
-							signal.state = .terminated
-							signal.updateLock.unlock()
-
-							for observer in snapshot.observers {
-								observer.action(snapshot.associated)
-							}
-
-							shouldDispose = true
-						} else {
-							signal.updateLock.unlock()
+					if case .terminating = signal.state, let snapshot = signal.shouldReallyTerminate() {
+						for observer in snapshot.observers {
+							observer.action(snapshot.associated)
 						}
+
+						shouldDispose = true
 					}
 				}
 
@@ -208,28 +191,15 @@ public final class Signal<Value, Error: Swift.Error> {
 				// Check if the status has been bumped to `terminating` due to a
 				// concurrent termination event that has not been caught in the main
 				// protected section.
-				//
-				// Note that as the check happens before acquiring `updateLock`, it
-				// could be a false positive, and thus a second check after the lock is
-				// acquired is mandatory.
 				if !shouldDispose, case .terminating = signal.state {
 					signal.sendLock.lock()
 
-					// Acquire `updateLock`. If the termination has still not yet been
-					// handled, take it over and bump the status to `terminated`.
-					signal.updateLock.lock()
-
-					if case let .terminating(snapshot) = signal.state {
-						signal.state = .terminated
-						signal.updateLock.unlock()
-
+					if let snapshot = signal.shouldReallyTerminate() {
 						for observer in snapshot.observers {
 							observer.action(snapshot.associated)
 						}
 
 						shouldDispose = true
-					} else {
-						signal.updateLock.unlock()
 					}
 
 					signal.sendLock.unlock()
@@ -246,11 +216,35 @@ public final class Signal<Value, Error: Swift.Error> {
 		generatorDisposable = generator(observer)
 	}
 
+	/// Return the state snapshot if the signal is in the `terminating` state, or
+	/// `nil` otherwise.
+	///
+	/// Calling this method as a result of a false positive `terminating` check
+	/// is permitted. `nil` would be returned in this case.
+	///
+	/// - note: The `updateLock` would be acquired.
+	///
+	/// - returns:
+	///   The state snapshot associated with a termination event, or `nil`.
+	private func shouldReallyTerminate() -> SignalStateSnapshot<Value, Error, Event<Value, Error>>? {
+		// Acquire `updateLock`. If the termination has still not yet been
+		// handled, take it over and bump the status to `terminated`.
+		updateLock.lock()
+
+		if case let .terminating(snapshot) = state {
+			state = .terminated
+			updateLock.unlock()
+			return snapshot
+		}
+
+		updateLock.unlock()
+		return nil
+	}
+
 	/// Swap the generator disposable with `nil`.
 	///
 	/// - returns:
 	///   The generator disposable, or `nil` if it has been disposed of.
-	@inline(__always)
 	private func swapDisposable() -> Disposable? {
 		if let d = generatorDisposable {
 			generatorDisposable = nil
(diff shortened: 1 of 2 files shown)


andersio:
_It is probably missing a memory barrier for it to work on ARMv8. Gonna run tests on an iPhone to ensure things are right._

andersio:
The takeaway is that up till now Swift has no memory model. So whatever synchronization we do is depending on undocumented behavior of the compiler. It _seems_ the Swift compiler chain does follow program order though, so only fences are inserted only for the platforms that need it.

Tests/ReactiveSwiftTests/PropertySpec.swift
```diff
@@ -265,6 +265,39 @@ class PropertySpec: QuickSpec {
 				property = nil
 				expect(isEnded) == true
 			}
+
+			it("should not deadlock") {
```
NachoSoto: Nice test

Sources/Signal.swift
```diff
@@ -52,105 +52,105 @@ public final class Signal<Value, Error: Swift.Error> {
...
-				// So we would treat `interrupted` events specially. If it happens
-				// to occur while the `sendLock` is acquired, the observer call-out and
+				// So we would treat termination events specially. If it happens to
+				// occur while the `sendLock` is acquired, the observer call-out and
 				// the disposal would be delegated to the current sender, or
 				// occasionally one of the senders waiting on `sendLock`.
 				if let state = signal.state.swap(nil) {
-					// Writes to `interruptedState` are implicitly synchronized. So we do
+					terminationState = SignalTerminationState(state: state,
+					                                          event: event)
```
sharplet: Does this really need to be wrapped onto a new line?

andersio: Nope. 😹

andersio:
_It seems the Swift compiler in Linux has started to impose some resilience rules._

andersio:
In https://github.com/ReactiveCocoa/ReactiveSwift/pull/137/commits/60bbd0ce8e45b14cdc41d2b11cb97b850958ef22:

1. Swift `enum` is incredible. Two pieces of mutable states now become one. 🍰
2. Having states as stored properties _seems_ a little bit faster, probably due to locality and/or less ARC calls?

Sources/Signal.swift
```diff
@@ -237,9 +235,55 @@ public final class Signal<Value, Error: Swift.Error> {
...
+private enum SignalStatus<Value, Error: Swift.Error> {
+	/// The `Signal` is alive.
+	case green
+
+	/// The `Signal` has received a termination event, and is about to be
+	/// terminated.
+	case yellow(SignalTerminationSnapshot<Value, Error>)
+
+	/// The `Signal` has terminated.
+	case red
```
mdiep: Would `alive`/`terminating`/`terminated` make more sense here?

Sources/Signal.swift
```diff
@@ -27,6 +27,13 @@ public final class Signal<Value, Error: Swift.Error> {
 	/// The state of the signal. `nil` if the signal has terminated.
 	private let state: Atomic<SignalState<Value, Error>?>
 
+	/// Holds the status of the signal. Implicitly protected by `state` and
+	/// `sendLock`.
+	private var status: SignalStatus<Value, Error>
+
+	/// Used to ensure that events are serialized during delivery to observers.
+	private let sendLock: NSLock
```
mdiep: This looks like it could remain inside the `init`.

Sources/Signal.swift
```diff
@@ -159,9 +153,13 @@ public final class Signal<Value, Error: Swift.Error> {
...
-		if state.swap(nil) != nil {
-			// As the signal can deinitialize only when it has no observers attached,
-			// only the generator disposable has to be disposed of at this point.
+		// A signal can deinitialize only when it is not retained and has no
+		// active observers. So `state` need not be swapped.
+		//
+		// The generator disposable needs to be disposed of only if a `green` status
+		// is observed here, since `red` would mean it has already been disposed of,
+		// and `yellow` is impossible to be observed here.
+		if case .green = status {
```
mdiep: This looks like it could still use the `state.swap(nil)`, in which case `status` could remain part of the `init`.

andersio: It is now swapped out before disposal in https://github.com/ReactiveCocoa/ReactiveSwift/pull/137/commits/63a7e0d28db724191a6a2254458f5dbe91f0f196, so no check is needed.

andersio:
(Checking on how `enum` behaves in 32-bit platforms)

Sources/Signal.swift
```diff
@@ -39,131 +45,129 @@ public final class Signal<Value, Error: Swift.Error> {
...
-						// signal is really still alive.
-						if !terminated {
-							interrupt(state.observers)
-							shouldDispose = true
-						}
-
-						sendLock.unlock()
+					// Check if a downstream consumer or a concurrent sender has
+					// interrupted the signal.
+					if case let .terminating(state) = signal.status {
```
mdiep: Do we need this check here with the other check right below?

(If so, I think that'd be helpful to document in the code.)

mdiep:
This looks fine to me, but I'd like to get some more eyes on it. @sharplet @NachoSoto

Sources/Signal.swift
andersio: The check within the protected section guarantees that _recursive_ termination event is immediate.

The check after releasing `sendLock` is a safeguard for _concurrent_ termination event that is missed by the protected section.

Sources/Signal.swift
```diff
@@ -27,6 +27,12 @@ public final class Signal<Value, Error: Swift.Error> {
 	/// The state of the signal. `nil` if the signal has terminated.
 	private let state: Atomic<SignalState<Value, Error>?>
 
+	/// The status of the signal. It synchronizes with `sendLock`.
+	private var status: SignalStatus<Value, Error>
```
sharplet: I think having both `state` and `status` feels confusing. It's unclear to me whether they're two related concepts and should be together, or two independent concepts and perhaps could be named more descriptively?

andersio: Yeah, unfold the state into a separate `observer` and `retaining` would help. Probably not in this PR though, but in #139 which really touches this piece.

mdiep: Maybe `observers` and `retaining` could move into `.alive`?

andersio: Nice idea. Note that I have pulled in the Read-Copy-Update change from #139 to make this work. :tada:

andersio: Nice idea. Implemented in https://github.com/ReactiveCocoa/ReactiveSwift/pull/137/commits/ecab87173d8d1f436848f439e3058bdf895e368d. :tada:
Note that I have pulled in Read-Copy-Update from #139 to avoid introducing locks.

andersio:
Hmm caught a few missing bits. The latest commit https://github.com/ReactiveCocoa/ReactiveSwift/pull/137/commits/1681a4444d15fb0cff4a39dc0dd014d9791442b2 is the updated version with proper documentation.

Sources/Signal.swift
```diff
@@ -89,18 +98,32 @@ public final class Signal<Value, Error: Swift.Error> {
...
 					if signal.sendLock.try() {
-						if case let .terminating(state) = signal.status {
-							state.send()
-							signal.status = .terminated
+						// Is it terminating? (False positive is allowed.)
+						if case .terminating = signal.status {
+							// Acquire `statusLock`, and handle the termination if the signal
+							// is still at `terminating` state.
+							signal.statusLock.lock()
+							if case let .terminating(state) = signal.status {
```
andersio: As noted in the comment, read-copy-update is used to synchronise `status`. Here we take advantage of the wait-free reading to `status`, so that we acquire the `statusLock` only if `terminating` is being seen by the sender.

The second confirmation after acquiring `statusLock` is to ensure that the termination is really not yet handled, since it could have observed `terminating` while some other concurrent sender is handling it.

Sources/Signal.swift
```diff
@@ -150,13 +150,21 @@ public final class Signal<Value, Error: Swift.Error> {
...
+					// could be a false positive, and thus a second check after the lock
+					// is acquired is mandatory.
 					if case .terminating = signal.status {
-						// Acquire `statusLock`, and handle the termination if the signal
-						// is still at `terminating` state.
+						// Acquire `statusLock`. If the termination has still not yet been
+						// handled, take it over and bump the status to `terminated`.
 						signal.statusLock.lock()
+
 						if case let .terminating(state) = signal.status {
```
andersio: The double checking is to free the normal event delivery from having to acquire `statusLock`, by exploiting wait-free reading of read-copy-update.

Concurrent senders perform a relaxed read from `status`, and enter the termination handling routine only if they see `terminating`. However, as this is relaxed and some other may have already handled it and bumped the status to `terminated`, a second check is performed after acquiring `statusLock`.

Sources/Signal.swift
```diff
@@ -237,9 +309,73 @@ public final class Signal<Value, Error: Swift.Error> {
...
-private struct SignalState<Value, Error: Swift.Error> {
-	var observers: Bag<Signal<Value, Error>.Observer> = Bag()
-	var retainedSignal: Signal<Value, Error>?
+/// The status of a `Signal`.
+///
+/// - note: This three-case, multi-payload enum was verified to be a packed
+///         object reference on i386, x86-64, armv7 and armv8 under the Swift
+///         3.0.1 ABI. Should any changes be made to the enum, please first
+///         verify that the change would not cause the enum to spill over the
+///         size of an object reference on all platforms.
```
mdiep: Can you detail how this is done?

Sources/Signal.swift
```diff
@@ -24,8 +24,20 @@ public final class Signal<Value, Error: Swift.Error> {
...
 
-	/// The state of the signal. `nil` if the signal has terminated.
-	private let state: Atomic<SignalState<Value, Error>?>
+	/// The status of the signal.
+	///
+ 	/// `status` synchronizes using Read-Copy-Update. Reads are thus wait-free,
+ 	/// but writes are required not to mutate in place. This suits `Signal` as
+ 	/// reads to `status` happens on the critical path of event delivery, while
+ 	/// adding and removing observers or termination generally has a constant
+	/// occurrence.
```
mdiep: Does Swift guarantee that the read will be consistent in this case?

Sources/Signal.swift
```diff
@@ -237,9 +309,73 @@ public final class Signal<Value, Error: Swift.Error> {
...
+///            Copy the states and create a new instance instead.
+private final class SignalState<Value, Error: Swift.Error, U> {
+	/// The observers of the `Signal`.
+	fileprivate let observers: Bag<Signal<Value, Error>.Observer>
+
+	/// A self-retaining reference. It is set when there are one or more active
+	/// observers.
+	fileprivate let retaining: Signal<Value, Error>?
+
+	fileprivate let associatedStorage: U?
```
mdiep: Now it looks like it'd be clearer to make this `let terminatingEvent: Event<Value, Error>?` and go make to a `SignalState?`. A nil state would mean terminated, a non-nil terminating event would mean terminating, and everything else would be active.

Having to states does still seem confusing.

Otherwise maybe we should aim for a straightforward fix here and handle optimizations in a subsequent PR.

Sources/Signal.swift
andersio: (Packed) object references are the same size as the platform pointer size (4 or 8 bytes with the natural alignment), so reads (and writes that simply overwrite) are atomic as guaranteed by the architectures.

`statusLock` is introduced to serialise read-modify-write operations.

Sources/Signal.swift
andersio: I make a little snippet of `MemoryLayout`, and run them in the iPhone simulator (i386) and a iOS template app (both 32-bit and 64-bit builds) on iPhone SE.

While AArch32 is not exactly the same as ARMv7, they are backwards compatible.

andersio: Though with the cleanup you suggested, this enum would be removed.

Sources/Signal.swift
andersio: Relying on just `SignalState` is a significant regression, since now it has to load the reference and then load `terminationEvent` to check for termination in _all_ cases. The extra state provided by the enum does help here.

If you prefer to, I can revert 1681a44 and b53e796 and leave them to a separate PR.

Sources/Signal.swift
andersio: You can disassemble using LLDB to verify this. For example on ARMv8, it looks like:
```
->  0x103005b24 <+528>:  ldr    x25, [x20, #32]
    0x103005b28 <+532>:  lsr    x8, x25, #62
    0x103005b2c <+536>:  cbnz   x8, 0x103005be8           ; <+724> at Signal.swift:169
```
with `x20` being the reference to `signal`, which is an aligned pointer, and the offset being 32 bytes. So the resulting address is still an aligned location.
```
     x20 = 0x00000001712649c0
```

Sources/Signal.swift
andersio: Related:
https://github.com/apple/swift/blob/master/include/swift/ABI/System.h
https://github.com/apple/swift/blob/master/lib/IRGen/ExtraInhabitants.cpp
https://github.com/apple/swift/blob/master/lib/IRGen/GenEnum.cpp#L2765

andersio:
Just in case these would be missed:

On Read-Copy-Update and the atomicity of reads to `Signal.state`.
https://github.com/ReactiveCocoa/ReactiveSwift/pull/137#discussion_r90780060

On `SignalState` being laid out as a tagged pointer.
https://github.com/ReactiveCocoa/ReactiveSwift/pull/137#discussion_r90787077

On why the wrapping of `SignalState` helps.
https://github.com/ReactiveCocoa/ReactiveSwift/pull/137#discussion_r90780687

andersio:
I have refactored the code a bit. Hope it helps the code explaining itself.

Tests/ReactiveSwiftTests/PropertySpec.swift
```diff
@@ -265,6 +265,39 @@ class PropertySpec: QuickSpec {
...
+				let queue: DispatchQueue
+
+				if #available(macOS 10.10, *) {
+					queue = DispatchQueue.global(qos: .userInitiated)
```
liscio: I think that if you want to match "high priority" you should consider using the `.userInteractive` QoS instead. Doubt it really matters in a synthetic benchmark like this, but the asymmetry stood out to me. :)

Sources/Signal.swift
```diff
@@ -237,9 +335,67 @@ public final class Signal<Value, Error: Swift.Error> {
...
+///            `SignalStateSnapshot`. Copy the states and create a new instance.
+private final class SignalStateSnapshot<Value, Error: Swift.Error, U> {
+	/// The observers of the `Signal`.
+	fileprivate let observers: Bag<Signal<Value, Error>.Observer>
+
+	/// A self-retaining reference. It is set when there are one or more active
+	/// observers.
+	fileprivate let retaining: Signal<Value, Error>?
+
+	fileprivate let associated: U
```
mdiep: Could we split `SignalStateSnapshot` into 2 different classes?

```swift
private enum SignalState<Value, Error: Swift.Error> {
    case alive(AliveState)
    case terminating(TerminatingState)
    case terminated
}

private final class AliveState {
    let observers: Bag<Signal<Value, Error>.Observer>
    let retaining: Signal<Value, Error>?
}

private final class TerminatingState {
    let observers: Bag<Signal<Value, Error>.Observer>
    let event: Event<Value, Error>
}
```

AFAICS the terminating state doesn't need `retaining` and the alive state definitely doesn't need `event`. This would simplify things a bit, I think.

Sources/Signal.swift
```diff
@@ -38,132 +56,207 @@ public final class Signal<Value, Error: Swift.Error> {
...
+	/// `nil` otherwise.
+	///
+	/// Calling this method as a result of a false positive `terminating` check
+	/// is permitted. `nil` would be returned in this case.
+	///
+	/// - note: The `updateLock` would be acquired.
+	///
+	/// - returns:
+	///   The state snapshot associated with a termination event, or `nil`.
+	private func shouldReallyTerminate() -> SignalStateSnapshot<Value, Error, Event<Value, Error>>? {
```
mdiep: This function is only used like this:

```swift
if let snapshot = signal.shouldReallyTerminate() {
    for observer in snapshot.observers {
        observer.action(snapshot.associated)
    }
}  
```

That suggests that the code to notify observers should move into this function.

It probably also makes more sense as a block in `init` since it's only used there.

Sources/Signal.swift
```diff
@@ -38,132 +56,207 @@ public final class Signal<Value, Error: Swift.Error> {
...
+	/// Return the state snapshot if the signal is in the `terminating` state, or
+	/// `nil` otherwise.
+	///
+	/// Calling this method as a result of a false positive `terminating` check
+	/// is permitted. `nil` would be returned in this case.
+	///
+	/// - note: The `updateLock` would be acquired.
+	///
+	/// - returns:
+	///   The state snapshot associated with a termination event, or `nil`.
```
mdiep: Can you format the `- returns:` to match the existing ones?

```
- returns: The stat snapshot associated…
```

Sources/Signal.swift
```diff
@@ -38,132 +56,207 @@ public final class Signal<Value, Error: Swift.Error> {
...
+			//   using the snapshot.
+			//
+			//   `sendLock` must be acquired.
+			//
+			// - Replace the snapshot of a signal that is alive.
+			//   (e.g. observers bag manipulation)
+			//
+			//   `updateLock` must be acquired.
+			//
+			// - Transit from `alive` to `terminating`.
```
andersio: I guess this needs a rename too, hmm?
https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/Atomic.swift#L27

Sources/Signal.swift
```diff
@@ -237,9 +337,71 @@ public final class Signal<Value, Error: Swift.Error> {
...
+/// The state of a `Signal`.
+///
+/// `SignalState` is guaranteed to be laid out as a tagged pointer by the Swift
+/// compiler in the support targets of the Swift 3.0.1 ABI.
+///
+/// The Swift compiler has also an optimization for enums with payloads that are
+/// all reference counted, and at most one no-payload case.
+private enum SignalState<Value, Error: Swift.Error> {
+	/// The `Signal` is alive.
+	case alive(AliveState<Value, Error>)
```
andersio: These should be less confusing once we have them nested in Swift 3.1, e.g. Signal.State.Alive.

mdiep:
It looks like [this comment](https://github.com/ReactiveCocoa/ReactiveSwift/pull/137#discussion_r90869722) was missed. Can you take a look at that?

I'm starting to like the way this looks. ✨

andersio:
Addressed in https://github.com/ReactiveCocoa/ReactiveSwift/pull/137/commits/56862278084e0501013c0302f168440b9f6c8392. ☕️

Sources/Signal.swift
```diff
@@ -24,8 +24,26 @@ public final class Signal<Value, Error: Swift.Error> {
...
+	/// routine are thus wait-free. But modifications, e.g. inserting observers,
+	/// still have to be serialized, and are required not to mutate in place.
+	///
+	/// This suits `Signal` as reads to `status` happens on the critical path of
+	/// event delivery, while observers bag manipulation or termination generally
+	/// has a constant occurrence.
+	///
+	/// As `SignalState` is a packed object reference (a tagged pointer) that is
+	/// naturally aligned, reads to are guaranteed to be atomic on all supported
+	/// hardware architectures of Swift (ARM and x86).
```
mdiep: Do we know that this holds true for Linux as well?

mdiep:
This looks good to me. But since it's a very critical section of code, I'd like to get another set of eyes on it.

Sources/Signal.swift
andersio: The Swift ABI defines this, regardless of OSes. You would always have lowest 2 bits available for packing on 32-bit platform anyway, due to the 4-byte minimum alignment.

andersio: 2 bits are enough for 4 with-payload cases, or 3 with-payload cases plus 2^31 (or 2^30? not quite sure) no-payload cases.

mdiep:
Doesn't look like we'll get another review, so let's 🚢.

PR #139, opened 2016-12-02
Wait-free observers bag for `Signal`.

andersio:
1. Grabbing the observers is now wait-free, at the cost of insertion and removal of observers being more expensive. (Read-Copy-Update)
2. ~~Optimization for one-observer `Signal`s. (esp in `SignalProducer`)~~ _Gonna leave it for later._

#### Implementation note
Unless user explicitly synchronises the input and the removal of an observer, removing an observer would not free the observer from receiving events in a concurrent environment _immediately_, but only _eventually_. This has been the case since 3.x, since concurrent senders all grab a copy of the observer bag.

[Benchmark Source (A ported test from CwlSignal)](https://gist.github.com/andersio/70ded6fb313df4ebeed19e98b50ea77a)
i5 4258U, macOS 10.12.1, Xcode 8.1, `-Owholemodule`.

`master`
```
AsyncMap
Performance is 9.032114933 seconds versus expected 3.3. Rate is 110716.040198555 per second.

SingleSignal
Performance is 8.795302343 seconds versus expected 3.0. Rate is 1136970.57929553 per second.
```

`master` + Read Copy Update
```
AsyncMap
Performance is 6.655907553 seconds versus expected 3.3. Rate is 150242.471374061 per second.
SingleSignal
Performance is 5.766949619 seconds versus expected 3.0. Rate is 1734018.96334479 per second.
```

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -27,8 +27,7 @@ public final class Signal<Value, Error: Swift.Error> {
 	/// The state of the signal. `nil` if the signal has terminated.
 	private let state: Atomic<SignalState<Value, Error>?>
 
-	/// Holds the status of the signal. Implicitly protected by `state` and
-	/// `sendLock`.
+	/// The status of the signal. It synchronizes with `sendLock`.
 	private var status: SignalStatus<Value, Error>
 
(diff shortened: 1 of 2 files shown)


andersio:
Blocked by #137.

PR #140, opened 2016-12-03
Less hops in operator lifting.

andersio:
The PR proposes to overhaul how `SignalProducer` produces a new instance of `Signal`.

## Summary

1. The public API of `SignalProducer` remains untouched.

1. `startWithSignal(_:)` is no longer responsible of creating the `Signal`.

1. `SignalProducer.Builder` replaces the "start handler" as the internal representation of `SignalProducer`.

     The builder is responsible of creating an instance of `Signal` on request. It also creates a customised post-creation side effect and an interrupt handle for every produced `Signal`.

     Unlike `startWithSignal(_:)`, the caller is responsible of invoking the post-creation side effect i.e. `didCreate`. The swap in responsibility here gives a room of optimisation for lifted `Signal` operators. They can now dodge the unnecessary relay `Signal` that would otherwise be created by the old `startWithSignal(_:)`.

1. Condensed the description of `SignalProducer`.

## Result

For every lifted `Signal` operator applied, one less `Signal` would be produced. So given a producer chain `SignalProducer(signal).map.map.map`, it would now produce a graph of four `Signal`s, instead of seven.

### DIFF
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -2,24 +2,41 @@ import Dispatch
 import Foundation
 import Result
 
-/// A SignalProducer creates Signals that can produce values of type `Value` 
-/// and/or fail with errors of type `Error`. If no failure should be possible, 
-/// `NoError` can be specified for `Error`.
+/// A `SignalProducer` produces a `Signal` with an instance of repeatable work which
+/// yields events to the `Signal`, every time it is started with an observer.
 ///
-/// SignalProducers can be used to represent operations or tasks, like network
-/// requests, where each invocation of `start()` will create a new underlying
-/// operation. This ensures that consumers will receive the results, versus a
-/// plain Signal, where the results might be sent before any observers are
-/// attached.
+/// Even if multiple instances of work from the same `SignalProducer` are executing
+/// concurrently, produced `Signal`s would see only events from its own instance of work,
+/// but not from the others.
 ///
-/// Because of the behavior of `start()`, different Signals created from the
-/// producer may see a different version of Events. The Events may arrive in a
-/// different order between Signals, or the stream might be completely
-/// different!
+/// The _work_ performed by a `SignalProducer` is also known as the _post-creation side
+/// effect_.
+///
+/// `SignalProducer`s can be used to represent repeatable operations or tasks — like
+/// network requests — that are customizable, lazy and/or on demand.
 public struct SignalProducer<Value, Error: Swift.Error> {
 	public typealias ProducedSignal = Signal<Value, Error>
 
-	private let startHandler: (Signal<Value, Error>.Observer, Lifetime) -> Void
+	/// Wraps a closure which, when invoked, produces a new instance of `Signal`, a
+	/// customized `didCreate` post-creation side effect for the `Signal` and a disposable
+	/// to interrupt the produced `Signal`.
+	///
+	/// Unlike the safe `startWithSignal(_:)` API, `Builder` shifts the responsibility of
+	/// invoking the post-creation side effect to the caller, while it takes from the
+	/// caller the responsibility of the `Signal` creation.
+	///
+	/// The design allows producer lifting to be as efficient as native `Signal`
+	/// operators, by eliminating the unnecessary relay `Signal` imposed by the old
+	/// `startWithSignal(_:)`, regardless of the fact that lifted operators can rely on
+	/// the upstreams for producer interruption.
+	///
+	/// If the caller is a `Builder`, it must invoke the produced `didCreate` before
+	/// performing any of its own post-creation side effect.
+	fileprivate struct Builder {
+		fileprivate let make: () -> (signal: Signal<Value, Error>, didCreate: () -> Void, interruptHandle: Disposable)
+	}
+
+	fileprivate let builder: Builder
 
 	/// Initializes a `SignalProducer` that will emit the same events as the
 	/// given signal.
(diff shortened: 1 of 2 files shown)


Sources/SignalProducer.swift
```diff
@@ -217,6 +360,39 @@ extension SignalProducer: SignalProducerProtocol {
...
+	/// In other words, this will create a new `SignalProducer` which will apply
+	/// the given `Signal` operator to _every_ created `Signal`, just as if the
+	/// operator had been applied to each `Signal` yielded from `start()`.
+	///
+	/// - parameters:
+	///   - transform: An unary operator to lift.
+	///
+	/// - returns: A signal producer that applies signal's operator to every
+	///            created signal.
+	public func lift<U, F>(_ transform: @escaping (Signal<Value, Error>) -> Signal<U, F>) -> SignalProducer<U, F> {
```
andersio: These trampolines are made as `liftLeft` and `liftRight` cannot be included in `SignalProducerProtocol`, and `lift` is here for consistency. Having said that, these trampolines can be removed when Swift 3.1 lands, should we drop the protocol in favour of concrete same-type requirements.

andersio:
It seems intermediate signals do not get `interrupted` events but are disposed of directly. Fixing.

mdiep:
The idea here makes sense: why create more signals than are needed? But I find the implementation here to be difficult to follow. I think we need to try to clarify/simplify the code before merging. I'll swing back around to this after we get 1.0 out (or at least until we get 1.0 PRs merged).

andersio:
Not in a hurry. Gotta think of a better explanation to all this. :coffee:

andersio:
Let's define some terminologies.

1. Starting a `SignalProducer` graph would create a `Signal` graph.
2. The started `SignalProducer` and the produced `Signal` are considered the root node of the respective graphs.

Note that the order is the reverse of what we usually called upstream and downstream producers. The ultimate upstreams are the leaves in the graph, and the ultimate downstream (that is being explicitly started by the user) is the root of the graph.

#### Interruptible Producer
The graph is constructed with a **top-down** approach.

Start handler: `(Observer, CompositeDisposable) -> Void`

When a depth 0 `SignalProducer` is started, the start handler would be invoked after the observer has been setup properly. The start handler may further start one or more depth 1 `SignalProducer`s. The process is recursively repeated until a depth N `SignalProducer` no longer starts any depth N+1 `SignalProducer`.

For each `SignalProducer`, a first level of `Signal` is created so that it provides a hook for asynchronous interruption. The event emitter of this `Signal` would then be passed into the start handler. After that, depending on the start handler, more levels of `Signal` might be created, and it could be part of the aforementioned recursive `SignalProducer` starting process.

Operator lifting, for example, would create another level of `Signal` by applying a `Signal` operator. So if a depth N `SignalProducer` is a lifted producer and it is preceded by a `Signal` graph of height M, the `Signal` graph would be of height M+2 afterwards.

#### Lifted Producer
The graph is constructed with a **bottom-up** approach.

Start handler: `() -> (signal: Signal, start: () -> Void, interrupter: Disposable)`

Let's first assume only the leaves would be interruptible producers.

When a depth 0 `SignalProducer` is started, a depth 0 `Signal` is created to provide a hook for asynchronous interruption.

The start handler would then be invoked after the observer has been setup properly. It would fabricate a `Signal`, a start action and an interrupter. Events of the returned `Signal` would be forwarded to the depth 0 `Signal`, and the interrupter would be placed into the composite disposable of the depth 0 `SignalProducer`. Finally, the start action would be invoked.

The fabricated `Signal` is the resulting `Signal` of the `Signal` transformations applied by the depth 0 `SignalProducer` on the root node of the `Signal` graph of depth 1 to depth N `SignalProducer`.

Similarly, the fabricated start action is an aggregate of all start actions of depth 0 to depth N `SignalProducer`, but would still be invoked in a top-bottom order. It is more or less the same for the fabricated interrupter, except that **the interruptibility of the intermediate `SignalProducer` is irrelevant, as long as the interruption is guaranteed to be propagated to the leaves of the `SignalProducer` graph**.

Now, as the start handler now breaks apart the creation of `Signal` graph , the start action and the interruption into three separate entity, a start handler can decide whther or not it needs the ability to be individually interrupted for its own purpose, and just pass through the start action and the interrupter of the next level.

Generally speaking, lifted `Signal` operator do not need the individual interruptibility, since `Signal` cannot be interrupted, and it can rely on the `interrupted` event from the leaves of the `Signal` graph to clean up resources.

Therefore, when lifting a `Signal` operator - unary or binary - on one or more depth N `SignalProducer`s, it would create a depth N-1 `SignalProducer` (downstream) with a start handler that applies the `Signal` operator directly to the `Signal`s fabricated by the start handlers of the depth N `SignalProducer`s (upstreams).

In other words, it would be the leaves of the `SignalProducer` graph who first create their `Signal`s and have it returned to its upper level. The process repeats recursively until it reaches the root node. That is why it is a bottom-up approach.

andersio:
> I think we need to try to clarify/simplify the code before merging. 

I do hope to figure out a way to unify both "modes" to burn down the giant switch in binary `lift`. But I have run out of ideas, and got my hands busy on other stuff. 🐝

Feel free to criticise and play with the branch for a better way out.
 

/cc @ReactiveCocoa/reactivecocoa

andersio:
Heh, it turned out to be quite straightforward. 🙈

andersio:
Updated and polished.

mdiep:
Trying to get to this. It's on my list 😅

andersio:
Note that this addressed #227, since now downstreams are interrupted by upstreams. IOW `interrupted` would now respect the async operators applied.

Sources/SignalProducer.swift
```diff
@@ -382,14 +404,58 @@ extension SignalProducer {
 	/// - returns: A signal producer that applies signal's operator to every
 	///            created signal.
 	public func lift<U, F>(_ transform: @escaping (Signal<Value, Error>) -> Signal<U, F>) -> SignalProducer<U, F> {
-		return SignalProducer<U, F> { observer, lifetime in
-			self.startWithSignal { signal, innerDisposable in
-				lifetime.observeEnded(innerDisposable.dispose)
-				transform(signal).observe(observer)
-			}
+		return SignalProducer<U, F>(SignalProducer<U, F>.Builder {
```
mdiep: Basically: this could be `return SignalProducer<U, F> { ... }`.

mdiep:
Sorry this one took so long! 😅

PR #141, opened 2016-12-03
Add throwing versions of attempt(_:) and attemptMap(_:)

sharplet:
I think these make a lot of sense now that we have `AnyError`, and should make it much more natural to work with code that uses Swift error handling.

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -2229,4 +2229,71 @@ extension SignalProtocol where Error == NoError {
 			.promoteErrors(NewError.self)
 			.timeout(after: interval, raising: error, on: scheduler)
 	}
+
+	/// Apply a failable `operation` to values from `self` with successful
+	/// results forwarded on the returned signal and thrown errors sent as
+	/// failed events.
+	///
+	/// - parameters:
+	///   - operation: A failable closure that accepts a value.
+	///
+	/// - returns: A signal that forwards successes as `value` events and thrown
+	///            errors as `failed` events.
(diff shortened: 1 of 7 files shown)


Sources/Signal.swift
```diff
@@ -2076,4 +2076,71 @@ extension SignalProtocol where Error == NoError {
...
+	/// Apply a failable `operation` to values from `self` with successful
+	/// results forwarded on the returned signal and thrown errors sent as
+	/// failed events.
+	///
+	/// - parameters:
+	///   - operation: A failable closure that accepts a value.
+	///
+	/// - returns: A signal that forwards successes as `value` events and thrown
+	///            errors as `failed` events.
+	public func attempt(_ operation: @escaping (Value) throws -> Void) -> Signal<Value, AnyError> {
```
mdiep: It's small, but can we put these right after the other `attempt` and `attemptMap`?

mdiep:
I don't love this, but I think that's just because I don't love Swift's untyped error handling. 🙃

mdiep:
I'd still love to see an update to this.

sharplet:
Yep hoping to get back to it skon

Sources/Signal.swift
sharplet: The existing operators are in the middle of an unconstrained protocol extension, whereas these are in a protocol extension where `Error == NoError`. Do you want me move the existing operators so that these can all go together?

sharplet:
I went ahead and pushed 06a2d53 to group the different overloads together. If you'd prefer not to reorganise, let me know and I'll revert that change.

PR #157, opened 2016-12-16
`observe(on:)` for properties + gardening in `PropertyProtocol`.

andersio:
Only the initial value is emitted synchronously — the rest of the events is forwarded to the specified scheduler.

As for why not lifting `SignalProducer.observe(on:)`, it forwards events emitted during the synchronous start process to the scheduler, which means the initial value is delivered asynchronously.

### DIFF
--- a/Sources/Property.swift
+++ b/Sources/Property.swift
@@ -52,15 +52,23 @@ extension MutablePropertyProtocol {
 /// A composed property would retain its ultimate source, but not
 /// any intermediate property during the composition.
 extension PropertyProtocol {
-	/// Lifts a unary SignalProducer operator to operate upon PropertyProtocol instead.
+	/// Lifts a unary SignalProducer operator to operate upon PropertyProtocol
+	/// instead.
+	///
+	/// - parameters:
+	///   - transform: A unary `SignalProducer` transform to apply on `self`.
 	fileprivate func lift<U>(_ transform: @escaping (SignalProducer<Value, NoError>) -> SignalProducer<U, NoError>) -> Property<U> {
-		return Property(self, transform: transform)
+		return Property(unsafeProducer: transform(self.producer))
 	}
 
-	/// Lifts a binary SignalProducer operator to operate upon PropertyProtocol instead.
+	/// Lifts a binary SignalProducer operator to operate upon PropertyProtocol
+	/// instead.
+	///
+	/// - parameters:
+	///   - transform: A binary `SignalProducer` operator.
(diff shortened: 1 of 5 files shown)


Sources/Property.swift
```diff
@@ -76,6 +76,34 @@ extension PropertyProtocol {
...
+			return SignalProducer { observer, disposable in
+				var hasDeliveredFirst = false
+
+				disposable += producer.start { event in
+					if hasDeliveredFirst {
+						disposable += scheduler.schedule {
+							observer.action(event)
+						}
+					} else {
+						observer.action(event)
```
mdiep: Why is the first event not delivered on `scheduler`?

andersio: `property` requires the initial value to be synchronously delivered.

andersio: `Property(unsafeProducer:)` requires the initial value to be synchronously delivered.

mdiep: The implementation in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3345 seems a lot simpler. 😕

andersio: But it has to send the value twice, which is unpleasant.

andersio: Though it does look a bit weird. Perhaps `lift` shouldn't be used.

mdiep: This makes sense. But I think it deserves a comment in the source explaining what's going on here.

andersio:
@mdiep Updated in https://github.com/ReactiveCocoa/ReactiveSwift/pull/157/commits/216fa6ecb0f638a3ead9aad4af0d53a53c062410.

An alternative is to allow composed properties to define different producers for composition and public use. But it is quite a complexity.

andersio:
Hmm, but it won't be safe if people treat the producer as if it is `SignalProducer.observe(on:)`, and use implicitly protected resources with it.

It seems either we accept the compromised solution in https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3345, or we have to blown up the complexity.

sharplet:
Can you explain what's unsafe about it?

andersio:
@sharplet Let's say we have:
```
latestCommitProducer
    .observe(on: backgroundScheduler)
    .observe(nonThreadSafeCollectionWork)

latestCommitProperty
    .observe(on: backgroundScheduler)
    .producer
    .observe(nonThreadSafeCollectionWork)
```

It would be fine in `SignalProducer`, as all the events are forwarded to `backgroundScheduler` even if they are emitted by the start handler.

But for this version of property `observe(on:)`, the initial value is replayed synchronously. That means the observer is invoked concurrently w.r.t. `backgroundScheduler`.

PR #160, opened 2016-12-16
Add Operators And, Or, Not to the PropertyProtocol

cristianames92:
With this PR I intend to add the basic operators `And`, `Or` and `Not`  to the PropertyProtocol for Boolean implementations. This will allow to make combinations easier without making any side effect


Sources/Property.swift
```diff
@@ -366,6 +366,23 @@ extension PropertyProtocol {
 	}
 }
 
+extension PropertyProtocol where Value == Bool {
+	/// Returns a property with the oposite value
+	var not: Property<Value> { return self.map { !$0 } }
```
andersio: These properties and methods should be `public`. Please make a new line for the return statement too. :)

Sources/Property.swift
```diff
@@ -366,6 +366,23 @@ extension PropertyProtocol {
...
+	
+	///Combines the property with the given property in a manner described by
+	///`and()` operator.
+	func and(_ property: Property<Value>) -> Property<Value> {
```
andersio: Ditto.

Sources/Property.swift
```diff
@@ -366,6 +366,23 @@ extension PropertyProtocol {
...
+		return self.combineLatest(with: property).map { $0 && $1 }
+	}
+	
+	///Combines the property with the given property in a manner described by
+	///`or()` operator.
+	func or(_ property: Property<Value>) -> Property<Value> {
```
andersio: Ditto.

Sources/Property.swift
```diff
@@ -366,6 +366,23 @@ extension PropertyProtocol {
...
 }
 
+extension PropertyProtocol where Value == Bool {
+	/// Returns a property with the oposite value
+	var not: Property<Value> { return self.map { !$0 } }
+	
+	///Combines the property with the given property in a manner described by
+	///`and()` operator.
+	func and(_ property: Property<Value>) -> Property<Value> {
+		return self.combineLatest(with: property).map { $0 && $1 }
```
andersio: It'd be better to use `lift` instead.

Sources/Property.swift
```diff
@@ -366,6 +366,23 @@ extension PropertyProtocol {
...
+	}
+	
+	///Combines the property with the given property in a manner described by
+	///`or()` operator.
+	func or(_ property: Property<Value>) -> Property<Value> {
+		return self.combineLatest(with: property).map { $0 || $1 }
```
andersio: Ditto.

andersio:
(It would even be great if a `Signal` operator is made instead, and have lifted variants for `SignalProducer` and finally `Property`.)

Sources/Property.swift
sharplet: It's hard to say for sure without some concrete examples, but I wonder if this operator might read better as `negated`?

Sources/Property.swift
cristianames92: You mean to do something like this?
```
public func and(_ newProducer: SignalProducer<Bool, NoError>) -> Property<Value> {
		return self.lift<Bool> { (producer: SignalProducer<Value, NoError>) in
			producer.combineLatest(newProducer).map(&&)
		}
	}
```

using `lift<U>` grom the `PropertyProtocol`?
```
fileprivate func lift<U>(_ transform: @escaping (SignalProducer<Value, NoError>) -> SignalProducer<U, NoError>) -> Property<U> {
		return Property(self, transform: transform)
	}
```

If thats the case i don't see the point to use it when we can use `combineLatest`

cristianames92:
I'll add test cases for this tomorrow

Sources/Property.swift
cristianames92: @sharplet I think `not` is more conventional than `negated` as you are using propositional logic.

cristianames92:
@andersio all the comments are resolved or responded. Do you mind to take a look at it?

Sources/Property.swift
```diff
@@ -371,6 +371,25 @@ extension PropertyProtocol {
 	}
 }
 
+extension PropertyProtocol where Value == Bool {
+	/// Returns a property with the oposite value
+	public var not: Property<Value> {
```
andersio: `negated` perhaps. It is not a boolean but a container of booleans, and properties should be named as nouns whenever possible.

Sources/Property.swift
```diff
@@ -371,6 +371,25 @@ extension PropertyProtocol {
 	}
 }
 
+extension PropertyProtocol where Value == Bool {
+	/// Returns a property with the oposite value
```
andersio: Please follow the [documentation guidelines](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Documentation/DocumentingCode.md). 😸 

```swift
 	/// Start the producer, then block, waiting for the last value.
 	///
 	/// When a single value or error is sent, the returned `Result` will
 	/// represent those cases. However, when no values are sent, `nil` will be
 	/// returned.
 	///
 	/// - returns: Result when single `value` or `failed` event is received.
 	///            `nil` when no events are received.
 	public func last() -> Result<Value, Error>? {
 		return take(last: 1).single()
 	}
```

Sources/SignalProducer.swift
```diff
@@ -1878,6 +1878,25 @@ extension SignalProducerProtocol {
...
 
+extension SignalProducerProtocol where Value == Bool {
+	/// Returns a producer with the oposite value.
+	public var not: SignalProducer<Value, Error> {
+		return self.lift { $0.not }
+	}
+	
+	///Combines the producer with the given producer in a manner described by
+	///`and()` operator.
+	public func and(_ producer: SignalProducer<Value, Error>) -> SignalProducer<Value, Error> {
```
andersio: Please add overloads for both `and` and `or` that take a `Signal`.

Sources/SignalProducer.swift
```diff
@@ -1878,6 +1878,25 @@ extension SignalProducerProtocol {
...
+		return self.lift(Signal.and)(producer)
+	}
+	
+	///Combines the producer with the given producer in a manner described by
+	///`or()` operator.
+	public func or(_ producer: SignalProducer<Value, Error>) -> SignalProducer<Value, Error> {
```
andersio: Ditto.

cristianames92:
@andersio hope you enjoyed Christmas if you celebrate it. I want to tell you that i just add the changes that you've requested.

Sources/SignalProducer.swift
```diff
@@ -1878,6 +1878,55 @@ extension SignalProducerProtocol {
 	}
 }
 
+extension SignalProducerProtocol where Value == Bool {
+	/// Returns a producer with the oposite value.
+	/// - returns: Producer that when started, for each event will emit a new event instead
```
andersio: A blank line between them, please?

Sources/SignalProducer.swift
```diff
@@ -1878,6 +1878,55 @@ extension SignalProducerProtocol {
...
+	///            with the new value
+	public var negated: SignalProducer<Value, Error> {
+		return self.lift { $0.negated }
+	}
+	
+	///Combines the producer with the given producer in a manner described by
```
andersio: A whitespace between `///` and `Combines`. Same for lines below.

Sources/SignalProducer.swift
```diff
@@ -1878,6 +1878,55 @@ extension SignalProducerProtocol {
...
+	///`and()` operator.
+	/// - parameters:
```
andersio: A sample:
```
	/// Create a producer that computes a logical AND between the latest results of `self`
	/// and `signal`.
	///
	/// - parameters:
	///   - signal: Signal to be combined with `self`.
	///
	/// - returns: A producer that emits boolean logical AND results.
```

cristianames92:
@andersio The new requests have been done.

Sources/SignalProducer.swift
```diff
@@ -1878,6 +1878,59 @@ extension SignalProducerProtocol {
...
+	
+	/// Create a producer that computes a logical AND between the latest values of `self`
+	/// and `producer`.
+	///
+	/// - parameters:
+	///   - producer: Producer to be combined with `self`.
+	///
+	/// - returns: A producer that emits the logical AND results.
+	public func and(_ producer: SignalProducer<Value, Error>) -> SignalProducer<Value, Error> {
+		return self.lift(Signal.and)(producer)
```
andersio: Oh, one more thing: may you please use `liftLeft` to be consistent with `combineLatest`?

Sources/SignalProducer.swift
```diff
@@ -1878,6 +1878,59 @@ extension SignalProducerProtocol {
...
+	
+	/// Create a producer that computes a logical OR between the latest values of `self`
+	/// and `producer`.
+	///
+	/// - parameters:
+	///   - producer: Producer to be combined with `self`.
+	///
+	/// - returns: A producer that emits the logical OR results.
+	public func or(_ producer: SignalProducer<Value, Error>) -> SignalProducer<Value, Error> {
+		return self.lift(Signal.or)(producer)
```
andersio: Ditto.

andersio:
LGTM otherwise. Thank you for your work. 😸

cristianames92:
@andersio I've made the last changes you request and merged the changes in master.

Tests/ReactiveSwiftTests/SignalProducerSpec.swift
```diff
@@ -2433,6 +2433,111 @@ class SignalProducerSpec: QuickSpec {
...
+				}
+				
+				producer.negated.startWithValues { value in
+					expect(value).to(beFalse())
+				}
+			}
+		}
+		
+		describe("and attribute") {
+			it("should emmit true when both producers emmits the same value") {
```
andersio: s/emmit/emit/

cristianames92:
I missed that. Fixes are done

PR #161, opened 2016-12-17
`observe(on:)` for properties.

andersio:
Preceded by #157.
With some gardening work.

The issue below was addressed by https://github.com/ReactiveCocoa/ReactiveSwift/pull/161/commits/5ae85c6d5805f7b516cf3c84e4f0b928dd54680f.

~~I am hesitate to name it `observe(on:)`, as:~~

> It won't be safe if people treat the producer as if it is `Signal[Producer].observe(on:)`, and use implicitly protected resources with it.

~~`Signal.observe(on:)` forwards everything to the given scheduler. That means all events are serialised by the scheduler, and it is safe to use queue contained resources.~~

~~`PropertyProtocol.observe(on:)`, however, makes an exception for the replayed value. The same issue happens if you apply `replayLazily` to any asynchronous `SignalProducer`, but at least it is more explicit.~~

Example:
https://github.com/ReactiveCocoa/ReactiveSwift/pull/157#issuecomment-267748325

~~Any thoughts on a better name, or maybe we should just drop it?~~

### DIFF
--- a/ReactiveSwift.xcodeproj/project.pbxproj
+++ b/ReactiveSwift.xcodeproj/project.pbxproj
@@ -231,6 +231,7 @@
 		57A4D2461BA13F9700F7D4B1 /* tvOS-Framework.xcconfig */ = {isa = PBXFileReference; lastKnownFileType = text.xcconfig; path = "tvOS-Framework.xcconfig"; sourceTree = "<group>"; };
 		57A4D2471BA13F9700F7D4B1 /* tvOS-StaticLibrary.xcconfig */ = {isa = PBXFileReference; lastKnownFileType = text.xcconfig; path = "tvOS-StaticLibrary.xcconfig"; sourceTree = "<group>"; };
 		7DFBED031CDB8C9500EE435B /* ReactiveSwiftTests.xctest */ = {isa = PBXFileReference; explicitFileType = wrapper.cfbundle; includeInIndex = 0; path = ReactiveSwiftTests.xctest; sourceTree = BUILT_PRODUCTS_DIR; };
+		9A06E9431E03C3F8001B1C63 /* Thread+LinuxSupport.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "Thread+LinuxSupport.swift"; sourceTree = "<group>"; };
 		9A090C131DA0309E00EE97CA /* Reactive.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = Reactive.swift; sourceTree = "<group>"; };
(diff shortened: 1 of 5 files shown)


sharplet:
Even if the events aren't serialised by the scheduler, they should still be synchronised by the send mutex shouldn't they? I'd like to understand what the potential issues are, and what precisely you mean by "queue contained resources".

andersio:
@sharplet Let's say we have:
```swift
latestCommitProducer
    .observe(on: backgroundScheduler)
    .observe(nonThreadSafeCollectionWork)

latestCommitProperty
    .observe(on: backgroundScheduler)
    .producer
    .observe(nonThreadSafeCollectionWork)
```

It would be fine in `SignalProducer`, as all the events are forwarded to `backgroundScheduler` even if they are emitted by the start handler.

But for this version of property `observe(on:)`, the initial value is replayed synchronously. That means the observer is invoked concurrently w.r.t. `backgroundScheduler`.

sharplet:
In that example, even using `backgroundScheduler` doesn't prevent concurrency, because it may be a concurrent dispatch queue. Unless I'm missing something, it doesn't seem like there's anything inherently less safe about this than the `SignalProducer` version?

andersio:
@sharplet `SchedulerProtocol` requires serial execution.

andersio:
https://github.com/ReactiveCocoa/ReactiveSwift/pull/161/commits/5ae85c6d5805f7b516cf3c84e4f0b928dd54680f now enforces the initial value to be replayed on the specified scheduler. Moreover, properties derived from an `observed(on:)`-applied property would inherit this behavior, aligning with the behavior of its `Signal` and `SignalProducer` counterpart.

This is done by allowing composed properties to optionally apply a "persona" to its (sourcing) replayed producer.

andersio:
A TL;DR of https://github.com/ReactiveCocoa/ReactiveSwift/pull/161/commits/5ae85c6d5805f7b516cf3c84e4f0b928dd54680f:

1. Property operators use `PropertyProtocol.replayed` for composition. `PropertyProtocol.replayed` returns `Property._producer` if `Self` is `Property`, or `Self.producer` otherwise.

2. Composed property can be created with a "persona". `Property.producer` would then return `Property._producer` with the persona applied. I do not foresee any persona other than `start(on:)` though.

mdiep:
It seems to me that the issue here is that `unsafeProducer` isn't a sufficient model in the face of `observe(on:)`. To rectify that, I think we need to come up with a better model for asking for the current value and observing future changes atomically (akin to `MutableProperty.producer`).

I don't think this model is the one we're looking for. We should be looking for something that simplifies and removes the need for `unsafeProducer`. (Which sounds like a 2.0 task.)

It seems like that might require access to the `Atomic`. But I think it'd be good to take a step back, forget about `observe(on:)` for a moment, and find a way to rework `Property`s to not require synchronous values from `producer`.

andersio:
Can't say I disagree on that. AFAICT it would require `PropertyProtocol` to have a new requirement to expose its locking mechanism, e.g. `withValue` would do it implicitly.

PR #164, opened 2016-12-18
Rename `innerDisposable` to `inner`.

andersio:
`SerialDisposable.innerDisposable` -> `SerialDisposable.current`.
`ScopedDisposable.innerDisposable` -> `ScopedDisposable.wrapped`.

### DIFF
--- a/Sources/Deprecations+Removals.swift
+++ b/Sources/Deprecations+Removals.swift
@@ -52,6 +52,19 @@ extension Disposable {
 	public var disposed: Bool { fatalError() }
 }
 
+extension SerialDisposable {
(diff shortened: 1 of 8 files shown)


sharplet:
Works for me 👍

mdiep:
I kinda like these both using the same name. They're really the same thing; it's just that one is mutable and the other immutable.

Personally, I think `inner` or `disposable` makes the most sense. `Wrapped` makes more sense to me as a generic type than `wrapped` does as an accessor.

andersio:
@mdiep Reverted to `inner`.

PR #165, opened 2016-12-18
Fix broken non-optional to optional bindings.

andersio:


Tests/ReactiveSwiftTests/UnidirectionalBindingSpec.swift
```diff
@@ -11,49 +11,91 @@ class UnidirectionalBindingSpec: QuickSpec {
...
-				property.value = 2
-				expect(value) == 2
+					let property = MutableProperty(1)
+					optionalTarget <~ property
+					expect(value) == 1
+
+					optionalTarget = nil
+
+					property.value = 2
+					expect(value) == 2
```
andersio: This would fail without the fix.

Related Bug Report: https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3350

PR #166, opened 2016-12-18
[WIP] Composed properties internal refactoring + `observe(on:)`.

andersio:
Preceded by #161.

~~The internal of `Property` drops one level down — it is now built around `Signal` and `PropertyProtocol.withValue`.~~

The internal of `Property` now relies on `PropertyProtocol.signal` and `PropertyProtocol.withValue` for composition.

The use of `unsafeProducer` in `map`, `observe(on:)` and `zip(with:)` is replaced with `Property(initial:then:producerTransform:)`.

`observe(on:)` is also included to demonstrate how it interacts with the new internal.

#### Breaking Change
`PropertyProtocol` has a new requirement: `withValue(_:)`.

Serialized properties should invoke the supplied action in its protected section. In other words, `withValue(_:)` implicitly exposes the lock without handing out a specific instance of it.

The documentation now also stresses that reentrancy is a strict requirement. This is not something new however — `MutableProperty` has set the bound.

#### What's the difference from `replayLazily`?
Less `Signal`s would be involved. At compile time there is also an apparent "current value", which can be used directly in stateless operators like `map` and `observe(on:)`.

But the notion of "unsafe" initializer still has to remain, because some operators are stateful and need to be "trained" with an initial value. e.g. `combineLatest(with:)`, `skipRepeats`.

**Bonus point:** `Property.signal` no longer needs to be started from the producer.

#### Special Notes on `observe(on:)`
There are two things that have to be decided on:

1. Should the current value be replayed on the supplied scheduler?

2. If (1) is yes, should the descendants of a property inherits (1) from its ancestor?

This branch currently assumes ~~_yes_ for (1), and _no_ for (2)~~ YES for both. For the (2) to work, it would still require the notion of producer persona in #161.

### DIFF
--- a/Sources/Property.swift
+++ b/Sources/Property.swift
@@ -76,14 +76,16 @@ extension PropertyProtocol {
 	/// Lifts a unary stateful SignalProducer operator to operate upon
 	/// PropertyProtocol instead.
 	fileprivate func lift<U>(_ transform: @escaping (SignalProducer<Value, NoError>) -> SignalProducer<U, NoError>) -> Property<U> {
-		return Property(unsafeValues: transform(self.values))
+		return Property(unsafeValues: transform(self.values),
+		                persona: self.persona)
 	}
 
 	/// Lifts a binary stateful SignalProducer operator to operate upon
 	/// PropertyProtocol instead.
 	fileprivate func lift<P: PropertyProtocol, U>(_ transform: @escaping (SignalProducer<Value, NoError>) -> (SignalProducer<P.Value, NoError>) -> SignalProducer<U, NoError>) -> (P) -> Property<U> {
 		return { other in
-			return Property(unsafeValues: transform(self.values)(other.values))
+			return Property(unsafeValues: transform(self.values)(other.values),
+			                persona: self.persona)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


andersio:
Hmm, apparently there is a race condition between the cache updating and the producer start handler.

Sources/Property.swift
```diff
@@ -22,6 +34,10 @@ public protocol PropertyProtocol: class, BindingSourceProtocol {
 	/// completes when the property has deinitialized, or has no further
 	/// change.
 	var signal: Signal<Value, NoError> { get }
+
+	/// Perform an arbitrary action on the current value of the property.
+	/// Serialized properties should invoke `body` in its protected section.
+	func withValue<R>(_ body: (Value) throws -> R) rethrows -> R
```
mdiep: Would it make more sense to add something similar to `startWithSignal`?

```
public func observe(_ handler: (Value, Signal<Value, NoError>) -> ())
```

Then there wouldn't be so many requirements?

andersio: If you meant the requirements up top, they have been the case since `Property` is introduced.

andersio:
Doesn't feel like there is any better options. If https://github.com/ReactiveCocoa/ReactiveSwift/pull/166/commits/ad263cef6cf51bb23b4f7c99946e73317872cb5e is repealed (since CAS loop is not safe with Darwin QoS), the only option left is bringing back a replaying layer i.e. `replayLazily`, or _(novel idea alert)_ have the replaying mechanic integrated into `Signal`.

andersio:
Close it for now. Gonna go through it again when #182, #129 and #140 have all landed.

PR #173, opened 2016-12-20
Update Property documentation. [skip ci]

andersio:


gavrix:
`Property` docs are out of sync too.
```
	/// A producer for Signals that will send the property's current
	/// value, followed by all changes over time, then complete when the
	/// property has deinitialized or has no further changes.
	public var producer: SignalProducer<Value, NoError> {
		return _producer()
	}
```

PR #178, opened 2016-12-22
Rename `observe(_:during:)` to `connect(with:during:)`.

andersio:


mdiep:
Personally, I think `observe` is the better name. I think it's a very accurate description of the behavior. It's fundamentally the same thing as `Signal`'s `observe`.

If we're concerned about people being confused by autocomplete, then I think we should add some documentation that will show up in the autocomplete.

andersio:
With that point of view though, shouldn't we rename `start*` to `observe*`? They all do the same thing `observe(_:during:)` does.

andersio:
Not a bad compromise. Though I'd still like to see `<~` gets melded back into protocols eventually so that it'd become invisible.

liscio:
@mdiep My original distaste for `observe` is based on passed an `Observer` instance without acknowledging the `Observer` itself.

Maybe `observe(with:during:)` is the compromise? In fact, I'd prefer if most of the commonly-used trailing closure methods would name the observer/closure that is passed in, so it reads better in the cases where an instance is used instead.

For example, if we had a signal's pipe somewhere, just called `pipe` for simplicity:

```swift
someSignal.observe(pipe, during: some.lifetime)
someSignal.observe(using: pipe, during: some.lifetime) // reads better, IMO
```

(And forgive me if we've had this discussion before re: the other methods like `observeValues(_:)` vs `observeValues(using:)`. Brain might as well be applesauce at this point…)

andersio:
@liscio It is kind of a compromise.

Ideally we can drop the named variant with `<~` that comes with a different signature: `(Observer, Lifetime) <~ Source` instead of the top-level `Target <~ Source`, which kinda addresses the customisation issue mdiep once worried about.

But when I attempt to implement it, it is really a shithole... Apparently something is wrong with the type checker with operators. Though they are apparently fixed in Swift 3.1. So we could revisit these in 2.0.

liscio:
> But when I attempt to implement it, it is really a shithole... Apparently something is wrong with the type checker with operators. Though they are apparently fixed in Swift 3.1. So we could revisit these in 2.0.

A major reason why I gave up and wrote a *lot* of the new Capo 3.5 stuff in straight-up Objective-C. So much time that I didn't have to waste chasing down the "how to do this", "is it the compiler", "is it me", or "is it the API?" issues. It's amazing how much nicer life is with a compiler and tools that work!

(But now I'm back to writing more swift3 code and hating life again. Don't worry… 😄)

andersio:
Well 90% of the time Swift is nice. The shitstorm hits only when building abstractions with stupid diagnostics that do not help at all, and limitations that you'd wish Swift 4 to come earlier than later. 🙈

liscio:
That's the trouble I have, personally. I spend far too much time in the 10%. ;)

I don't want to give the impression that I dislike the language, or working with it. It's just that when I'm in a time crunch, I can't have my tools adding 20-40% overhead in added compile times, time spent spinning my wheels tracking down unhelpful error messages, etc.

Anyway, this isn't the place to gripe about swift! 🙊

mdiep:
> My original distaste for observe is based on passed an Observer instance without acknowledging the Observer itself.

That was based on the existing `Signal` method:

```swift
	public func observe(_ observer: Observer) -> Disposable? {
```

I can totally see where naming that parameter might be preferable. But I was just aiming for consistency with existing methods.

PR #179, opened 2016-12-22
Update package manager instructions in README. [skip ci]

andersio:


mdiep:
This should wait until we're actually ready to tag 1.0. Maybe I should be a part of that PR so we can also update the podspec?

andersio:
Ah. I opened this because I thought CocoaPods would fetch RCs with `~> 1.0` like Carthage. But it turns out to be not.

PR #180, opened 2016-12-23
Remove the `BindingTarget` hackery.

andersio:
There is a breaking bit, but should affect only those who conforms to `BindingTargetProtocol`.

### DIFF
--- a/Sources/Action.swift
+++ b/Sources/Action.swift
@@ -259,8 +259,8 @@ public protocol ActionProtocol: BindingTargetProtocol {
 }
 
 extension ActionProtocol {
-	public func consume(_ value: Input) {
-		apply(value).start()
+	public var consume: (Input) -> Void {
+		return { [weak self] in self?.apply($0).start() }
 	}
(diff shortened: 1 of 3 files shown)


mdiep:
I played around with this idea when I added `BindingSourceProtocol`, but it felt a little weird to me. 😕 Making `consume` a property didn't feel right.

I'd be curious to hear what others think.

andersio:
Maybe name it with a noun, heh? `valueEater`, perhaps.

andersio:
It might also worth making the scheduler explicit. In the two-way binding prototype, binding downstreams are required to specify its containing scheduler, or use `ImmediateScheduler` if it isn't contained.

https://github.com/ReactiveCocoa/ReactiveSwift/commit/7c37379ea6a1a8acbde995c3627558e9fdd90670#diff-86840718b7ab1350b885a0ab78a8e453R5

andersio:
Perhaps we can revisit it later.

Issue #181, opened 2016-12-23
Bidirectional Bindings.


andersio:
For two-way UI bindings, we'd need (as a summary of previous scattered discussions on Slack and various PR and issues):

1. [x] **Fine-grained Property Validations.**

2. [ ] **Decoupling of the presentation type and the actual stored type.**
    e.g. A property is stored as `Double`, but the View Model converts it into a `String` for presentation, and accepts `String` inputs that are convertible back into `Double` with its own rules.
   
3. [ ] **Write conflict handling.**

4. [ ] All of above in one thing, but configurable.

We could start iterating with basic building blocks, and eventually a two-way bindable construct. Some supporting constructs that might be needed:

1. [x] Changes to **`MutablePropertyProtocol` to support new features.**

1. [ ] **Standard binding "packages" for controls.**
    Let's call it `ControlPod` for the time being.
    e.g. `control.reactive.textPod` that packs `text`, `textValues` and `isEnabled` all together.

1. [ ] **Operator:** `<~>`.
    The operator allows `Bidirectional <~> ControlPod` and `Bidirectional <~> MutablePropertyProtocol`. Perhaps also `Action <~> ControlPod`?

andersio:
A second thought though: All of these proposed constructs look representable in `Action` in some way. Perhaps extending `Action` would be better take, hmm? But we would need an `Action` that guarantees to cache the latest value.

RuiAAPeres:
It feels to me, I might be wrong, that an Action is normally used between the UI and the ViewModel and tied to an user action (most of the times a tap on a `UIButton`).

andersio:
It is one of its use case. But `Action` in general is a serialised worker that can have an internal state, and when fired it executes on a snapshot of the state, together with an arbitrary input.

See the example in the [README](https://github.com/ReactiveCocoa/ReactiveSwift#action-a-serialized-worker-with-a-preset-action).

andersio:
[A gist](https://gist.github.com/andersio/fdbee89008a730e3e3a386d7d1fd5b3d) to illustrate the viability. I can think of various juicy ways to compose with wrapper based approach, esp. with conditional conformances. Not sure how the `Action` based variant can be composed though.

andersio:
[...]

PR #182, opened 2016-12-24
Mutable Validating Property.

andersio:
Related proposal: #181 

🎄

- [X] Self-validating mutable property.
- [X] Interactive playground.
- [x] Test Coverage.

#### TL;DR
For a more solid example, please [check the playground](https://github.com/ReactiveCocoa/ReactiveSwift/pull/182/files#diff-3991ca854a66db6dc382cc0bd2f65dbeR28).


ReactiveSwift.playground/Pages/ActionProperty.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,138 @@
+import ReactiveSwift
+import Result
+
```
mdiep: I really like having a playground page in a PR like this. That really helps demonstrate the idea. 👏 

But I think it'd be better to create something that approximates a real-life example. So create a faux reset password example or something.

Sources/ActionProperty.swift
```diff
@@ -0,0 +1,405 @@
+import Result
+
+/// A mutable, observable property that has an optionally failable action
+/// associated with the setter.
+public final class ActionProperty<Value, ActionError: Error>: ComposableMutablePropertyProtocol {
```
mdiep: Any similarities to `Action` aside, I don't think this concept should have `Action` in the name`.

Sources/ActionProperty.swift
```diff
@@ -0,0 +1,405 @@
...
+	///
+	/// - parameters:
+	///   - forward: The value transform to convert `Self.Value` to `U`.
+	///   - backward: The value transform to convert `U` to `Self.Value`.
+	///
+	/// - returns: A mapping `ActionProperty`.
+	public func map<U>(
+		forward: @escaping (Value) -> U,
+		backward: @escaping (U) -> Value
+	) -> ActionProperty<U, NoError> {
```
mdiep: It doesn't seem like bidirectionality should be required for this. But a few real life examples in the playground would probably demonstrate what you had in mind.

Sources/ActionProperty.swift
```diff
@@ -0,0 +1,405 @@
...
+	private let base: Property
+
+	init(_ base: Property) { self.base = base }
+
+	override func lock<R>(_ action: (()) throws -> R) rethrows -> R {
+		return try base.withValue { _ in return try action(()) }
+	}
+}
+
+private class ActionPropertyBoxBase<Value> {
```
mdiep: This abstraction seems to add complexity without much benefit.

Sources/ActionProperty.swift
andersio: https://gist.github.com/andersio/8f0ae67255b959e227aabb4944fa5b6d

Sources/ActionProperty.swift
andersio: It is an existential wrapper that boxes the root properties' lock. Thread safety stuff.

Sources/ActionProperty.swift
andersio: Generally speaking, it allows a particular value in the view layer to be decoupled from its view model layer type.

Sources/ActionProperty.swift
andersio: It is essential for thread safety, as far as I am concern. We need to run the validation logic within the root properties' protection, so that the whole thing is an atomic unit of work.

Even if the validation might not need this, the two-way binding mediator would need it to ensure data consistency.

Sources/ActionProperty.swift
andersio: Suggestions are always welcomed. 🙈

Sources/ActionProperty.swift
andersio: It is required too. Consider if you have `ActionProperty<Int, NoError>` mapped to `ActionProperty<String, NoError>`, if you perform a mutation through the later one, the operator needs to know how to map `String` back to `Int`.

andersio:
The playground has been updated with interactive elements.
🎄

andersio:
It seems there has to be a way to specify relationships too... say a `validateLatest(with:)`.

In the playground example, `emailConfirmation` depends on `email`. But `emailConfirmation` would not flip to be valid if we match `email` with it, instead of another way round.

RuiAAPeres:
Love this PR and the`map(forward:backward:)` is just beautiful. Thanks @andersio.

Sources/TransactionalProperty.swift
```diff
@@ -0,0 +1,509 @@
...
+	}
+
+	internal func revalidate() {
+		rootBox.lock {
+			action(cache.value)
+		}
+	}
+}
+
+public protocol _TransactionalPropertyProtocol: ComposableMutablePropertyProtocol {
```
andersio: Underscored as it is expected to be gone in Swift 3.1.

ReactiveSwift-UIExamples.playground/Pages/TransactionalProperty.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,133 @@
...
+		submit = Action(state: state, enabledIf: { $0.0 }) { state, _ in
+			return SignalProducer { observer, disposable in
+				observer.send(value: state.1)
+				observer.sendCompleted()
+			}
+		}
+	}
+}
+
+final class ViewController {
```
RuiAAPeres: `: UIViewController`. Did you miss the inheritance on purpose?

andersio: Hmm. kind of. The playground takes a `liveView` of `UIView` only. So unless a `UIWindow` is used, the VC hierarchy won't work.

RuiAAPeres: Also, what is the reasoning behind making the ViewModel a class? (I am quite curious, since I struggled with this decision)

andersio: It doesn't really matter. `struct` could do the job equally well in this case.

andersio: It semantically does not matter. But for view models in collection views, you _might_ prefer classes to structs, as when passing it around the ARC side effects happen on just the view model reference, not every single property of the struct.

But in the end, profile before optimization. :)

Sources/TransactionalProperty.swift
```diff
@@ -0,0 +1,690 @@
...
+	public var property: TransactionalProperty<Value, TransactionError> {
+		return self
+	}
+}
+
+public enum ValidationResult<Value, Error: Swift.Error> {
+	case success(Value)
+	case failure(Value, Error)
+
+	public var isFailed: Bool {
```
liscio: `didFail` would be a better name for this, or `isFailure` if you really want to match the case label.

Sources/TransactionalProperty.swift
```diff
@@ -0,0 +1,690 @@
...
+///
+/// ## Consistency when nested
+///
+/// `TransactionalProperty` would back-propagate the validation failure if a
+/// proposed value from an outer property fails with its inner property.
+///
+/// It would also evaluate values originated from the inner property. In other
+/// words, it is possible for `value` to be invalid, causing `validations`
+/// to be a `failure`. Rely on `validations` for asserting a pass in validation.
+public final class TransactionalProperty<Value, TransactionError: Error>: ComposableMutablePropertyProtocol {
```
liscio: Love the functionality, but not the name.

I don't sense anything that feels like a "transaction" when reading examples of using the class. Instead, I think that something like `PropertyEditor`, or the previously used `EditableProperty`, might better explain what the class does. Also, I find that the mixing of a "forward `map`" and `transform` (i.e. the `init` parameters vs the `map(forward:backward:)`) is a little confusing.

The thing that feels "weird" about `PropertyEditor`, though is that it behaves just like a property. So if I have a variable like `emailEditor` then it'd be weird to ask for `emailEditor.value` or `emailEditor.producer`.

Another idea that sprung up just now is a `{Mutable,}PropertyTransformer`, or maybe even `MutablePropertyAdapter` (with the former being my preferred of the two.) Those would tie in nicely with the transformation nomenclature that's used in places.

Sources/TransactionalProperty.swift
```diff
@@ -0,0 +1,690 @@
...
+	// - The default `map` with `NoError`.
+
+	/// Create a mutable mapped view to `self`.
+	///
+	/// - parameters:
+	///   - forward: The value transform to convert `Self.Value` to `U`.
+	///   - backward: The value transform to convert `U` to `Self.Value`.
+	///
+	/// - returns: A mapping `TransactionalProperty`.
+	public func map<U>(
```
liscio: If we changed the name from `TransactionalProperty` to `PropertyEditor`, and call the "value transformers" as they are, we could instead use this sort of signature:

```swift
public func editor<U>(
    transform: @escaping (Value) -> U,
    reverseTransform: @escaping (U) -> Value
) -> PropertyEditor<U, NoError> {
    return PropertyEditor(self, transform: transform) { proposedInput in
        return .success(reverseTransform(proposedInput))
    }
}
```

Or, if we use the `PropertyTransformer` naming, this could be:

```swift
public func valueTransformer<U>(
    forward: @escaping (Value) -> U,
    reverse: @escaping (U) -> Value
) -> PropertyEditor<U, NoError> {
    return PropertyTransformer(self, transform: forward) { proposedInput in
        return .success(reverse(proposedInput))
    }
}
```

I used `reverse` since it better matches the `NSValueTransformer` conventions. Also, something about `reverse` sounds more correct than `backward` to me for some reason.

Thoughts?

Sources/TransactionalProperty.swift
andersio: > The thing that feels "weird" about PropertyEditor, though is that it behaves just like a property. So if I have a variable like emailEditor then it'd be weird to ask for emailEditor.value or emailEditor.producer.

These are more like "committed" values. But since we can't rename conformances... hmm. Though it might be more appropriate if it does not conform to `MutablePropertyProtocol` at all, given that it has different semantics than a normal mutable property.

andersio:
Summary of changes

1. Renamed to `PropertyEditor`.
2. The failable setter is now `PropertyEditor.try`.
3. Committed values are now exposed through `PropertyEditor.commited`.
4. `[try]Reverse` instead of `[attempt]Backward`.

RuiAAPeres:
When I think about the name`PropertyEditor` my mind goes like this:

editor -> being able to be edited -> mutable -> `MutableProperty`?

Not really keen on this one.

liscio:
@RuiAAPeres Mutability is not the same as editing, necessarily. In this case, `PropertyEditor` is a "thing that makes edits to a property," often on behalf of the user. I'd be totally on your side if we had went with `EditableProperty` instead, which would apply to your logic more closely. (That would read like "property that is able to be edited" rather than "editor of properties.")

If it makes you feel any better, AppKit uses a similar naming to represent an editor in the [NSEditor](https://developer.apple.com/reference/appkit/1654405-nseditor) protocol. In the case of `NSEditor`, you can think of it as managing/reporting the editing state of the `representedObject` in a view controller, or the `objectValue` in a control.

What we have here is quite different than `NSEditor`, and it wraps up the following concepts from AppKit:

* [NSValueTransformer](https://developer.apple.com/reference/foundation/valuetransformer)
* [Property Validation](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/KeyValueCoding/ValidatingProperties.html)

I think the above bits combine well into a "thing that is making/managing/watching edits on a property."

Sources/PropertyEditor.swift
```diff
@@ -0,0 +1,680 @@
...
+/// It would also evaluate values originated from the inner property. In other
+/// words, it is possible for `value` to be invalid, causing `validations`
+/// to be a `failure`. Rely on `validations` for asserting a pass in validation.
+public final class PropertyEditor<Value, ValidationError: Error> {
+	/// The commited value of the property.
+	///
+	/// It does not guarantee that the value is valid with regard to `self`, since
+	/// changes might be initiated from the inner properties. Check `validations`
+	/// for the latest validation state.
+	public var commited: Property<Value>
```
liscio: Nit: "committed" has two 't's :)

Sources/PropertyEditor.swift
```diff
@@ -0,0 +1,680 @@
...
+				inner.property.action(innerResult)
+
+			case let .failure(error):
+				_validations.value = .failure(input, error)
+			}
+		}
+	}
+
+	@discardableResult
+	public func `try`(_ newValue: Value) {
```
liscio: I think `attemptSet()`(or if you don't mind the verbosity, `attemptToSet(value newValue: Value)` might be more appropriate here. The same applies elsewhere—I think I prefer `attempt` to `try` seeing how this has little/nothing to do with Swift exceptions.

(My distaste for `try` here is exacerbated by the fact that you also have actual uses of Swift's `try` keyword in the same file.)

Sources/PropertyEditor.swift
```diff
@@ -0,0 +1,680 @@
...
+				return .failure(transform(value), error)
+			}
+
+		case let .failure(value, error):
+			return .failure(transform(value), errorTransform(error))
+		}
+	}
+}
+
+// FIXME: Remove the type parameter that works around type checker weirdness.
```
liscio: I have no idea what the comment means. Maybe an example, or explain what the "type parameter" is in this case?

Sources/PropertyEditor.swift
```diff
@@ -0,0 +1,680 @@
...
+
+	/// The latest validation that have been made by the property.
+	public let result: Property<ValidationResult<Value, ValidationError>>
+
+	/// The action associated with the property.
+	private let action: (Value) -> Void
+
+	/// The existential box that wraps the synchronization mechanic of the root
+	/// property of the composed chain.
+	private let rootBox: PropertyEditorBoxBase<()>
```
liscio: Both the variable name and comment leave me confused after reading. I can only tell from usage that rootBox appears to guard accesses/changes to the originating property.

Sources/PropertyEditor.swift
```diff
@@ -0,0 +1,680 @@
...
+	/// It does not guarantee that the value is valid with regard to `self`, since
+	/// changes might be initiated from the inner properties. Check `validations`
+	/// for the latest validation state.
+	public var commited: Property<Value>
+
+	/// The lifetime of the property.
+	public let lifetime: Lifetime
+
+	/// The latest validation that have been made by the property.
+	public let result: Property<ValidationResult<Value, ValidationError>>
```
liscio: Either this is supposed to be called `validations`, or the comment for `commited` (sic) needs to be updated. I think it's probably just the latter.

Assuming it stays `result` I would suggest the comment reads:

/// The result of the last attempted edit on the underlying property

(Unless my understanding is broken, of course.)

Sources/PropertyEditor.swift
```diff
@@ -0,0 +1,680 @@
+import Result
+
+/// A mutable, observable property that has an optionally failable action
```
liscio: Unless I'm missing the point, this might instead read:

> An object that manages edits to an underlying MutableProperty using a setter action that can (optionally) validate proposed values.

Sources/PropertyEditor.swift
```diff
@@ -0,0 +1,680 @@
...
+/// associated with the setter.
+///
+/// ## Consistency when nested
```
liscio: I'm not sure I understand this section fully. Are you suggesting that a change in the underlying property might cause a failure to then get reported? The test cases appear to suggest this is the case, but the comment here is not so helpful.

Perhaps an inline, simple example would be helpful, similar to what we see all over the standard library comments.

Sources/PropertyEditor.swift
```diff
@@ -0,0 +1,680 @@
...
+///
+/// `PropertyEditor` would back-propagate the validation failure if a
+/// proposed value from an outer property fails with its inner property.
+///
+/// It would also evaluate values originated from the inner property. In other
+/// words, it is possible for `value` to be invalid, causing `validations`
```
liscio: `s/value/committed/`
`s/validations/result/`

Sources/PropertyEditor.swift
andersio: Oops.

Sources/PropertyEditor.swift
andersio: It refers to the redundant `<()>` type parameter in the inheritance clause below. Without it weird things happen.

RuiAAPeres:
Because I am not a native english speaker, I am not too precious about this, but in any case:

> In this case, PropertyEditor is a "thing that makes edits to a property,"

>That would read like "property that is able to be edited" rather than "editor of properties."

I find these very similar. 

From this PR, I am able to get some context on the why this is called as it is. I am unsure a person who is used to `Property` and `MutableProperty` would.

> AppKit uses a similar naming to represent an editor in the NSEditor protocol. In the case of NSEditor, you can think of it as managing/reporting the editing state of the representedObject in a view controller, or the objectValue in a control.

Good background. 👍 I think @andersio reasoning makes sense:

> +/// An editor that monitor, validates and commits changes to its root property.

andersio:
All comments were addressed. :)

RuiAAPeres:
LGTM. 👍 Great job as usual @andersio. 😊

Sources/PropertyEditor.swift
```diff
@@ -0,0 +1,698 @@
...
+
+extension PropertyEditor: BindingSourceProtocol {
+	public func observe(_ observer: Observer<Value, NoError>, during lifetime: Lifetime) -> Disposable? {
+		return committed.producer
+			.take(during: lifetime)
+			.start(observer)
+	}
+}
+
+public protocol _PropertyEditorProtocol: class {
```
mdiep: Why a `_`-prefixed protocol?

Sources/PropertyEditor.swift
```diff
@@ -0,0 +1,698 @@
...
+				_validations.value = .failure(input, error)
+			}
+		}
+	}
+
+	/// Attempt to edit the property with the given value.
+	///
+	/// - parameters:
+	///   - newValue: The proposed value.
+	public func attemptSet(_ newValue: Value) {
```
mdiep: Why let values be set in this direction? Bidirectionality adds a LOT of complexity, and I can only think of 1 instance where I've ever wanted something  like this.

Sources/PropertyEditor.swift
andersio: Expected to be removed in Swift 3.1.

andersio:
@mdiep 

> Like I mentioned above, I don't understand why this is bidirectional. At the very least, it seems like we'd want both a unidirectional and bidirectional version of this (I'd bet most validations are simple) in which case it makes sense to start with the simple, unidirectional version.
>
> (And now that I think about it, the one case I have done a bidirectional binding would have been better served by a simple bidirectional binding and two separate validations.)

Bidirectional _binding_ synchronises two entities. The bidirectional _mapping_ here decouples the presentation type and the stored type at the boundary of layers.

> Why let values be set in this direction? Bidirectionality adds a LOT of complexity, and I can only think of 1 instance where I've ever wanted something like this.

`PropertyEditor` is not a bidirectional construct. It is just a view to root property, and it merely blocks invalid values on the write path from propagating to the backing "root" property.

mdiep:
But why make it mutable? Why not make this a readonly, validated view of the property?

andersio:
It caters situations that have updates from both ends.

Let's say:
```
let validatedName = name.validate { ... }.validate { ... }
account.nameProperty <~>> name
account.nameProperty <~ validatedName.values
```

If you write back successfully validated values with a simple `<~` binding, it would form an infinite feedback loop. Of course it can be mitigated by value equality, but this is also what I am trying to dodge.

It also requires two separate properties in the VM to expose both the validation results and the point  of binding. In the playground the error context is aggregated. But isn't it more common in practice to have these HUDs on a per-control basis?

ReactiveSwift-UIExamples.playground/Pages/ValidatingProperty.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,153 @@
...
+	let termsAccepted: MutableProperty<Bool>
+
+	let reasons: Signal<String, NoError>
+
+	let submit: Action<(), (), FormError>
+
+	init(userService: UserService) {
+		email = MutableValidatingProperty<String, FormError>("") { input, error in
+			if !input.hasSuffix("@reactivecocoa.io") {
+				error = .invalidEmail
```
liscio: I'm not a huge fan of this pattern for returning an error. Would it not make more sense to return a `Result` here? This would also give the validator an opportunity to modify the attempted value coming in.

Code would look like:

``` swift
email = MutableValidatingProperty<String, FormError>("") { input, error in
    guard input.hasSuffix("@reactivecocoa.io") else { return .failure(.invalidEmail) }
    return .success(input)
}
```

Or, if we want to mess with the value and save the user a bit of time:

```
email = MutableValidatingProperty<String, FormError>("") { input, error in
    if input.isAlreadyPerfect {
        return .success(input)
    } else if !input.passesSomeArbitraryBadCharacterTestsAndCantBeFixed {
        return .failure(.invalidEmail)
    } else {
        let cleanInput = input.removingAllSuffixesAndWhitespace
        return .success(cleanInput + "@reactivecocoa.io")
}
```

Validation in Cocoa works in this way, where the `validateValue:forKeyPath:` call is given an in-out pointer so you can manipulate the value. I use this pattern myself in [FuzzMeasure](http://fuzzmeasure.com) when a user enters a value outside of an accepted range of values rather than barking at them with errors.

Now, it would be *really nice* if this API could somehow also allow a third, "Input was invalid but here is a new, valid value that I replaced the user's entry with" state. This ability is missing with the existing Cocoa behavior. It doesn't allow me to easily detect when the user's input was overridden so I could call attention to it in my UI. The user just sees the value flip to something different when the commit the edit on a text field right now.

But I suppose that's as easy as wrapping the `Value` itself in a new enum, so probably already possible using what you've proposed here.

For example:

``` swift
email = MutableValidatingProperty<UserInput<String>, FormError>("") { input, error in
    if input.isAlreadyPerfect {
        return .success(.unmodified(input))
    } else if !input.passesSomeArbitraryBadCharacterTestsAndCantBeFixed {
        return .failure(.invalidEmail)
    } else {
        let cleanInput = input.removingAllSuffixesAndWhitespace
        return .success(.modified(cleanInput + "@reactivecocoa.io", "Hey, we fixed that for you."))
}
```

Then, elsewhere in your code you'd just watch for those modified values and update the UI to call attention to the helpful messages.

Thoughts?

RuiAAPeres:
@liscio are you happy with the changes?

ReactiveSwift-UIExamples.playground/Pages/ValidatingProperty.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,149 @@
...
+
+		// A `Property` of the validated username.
+		//
+		// It outputs the valid username for the `Action` to work on, or `nil` if the form
+		// is invalid and the `Action` would be disabled consequently.
+		let validatedEmail = Property.combineLatest(email.result,
+		                                            emailConfirmation.result,
+		                                            termsAccepted)
+			.map { !$0.isFailure && !$1.isFailure && $2 }
+			.combineLatest(with: email)
```
mdiep: I don't actually think this is safe. `email` here isn't guaranteed to be the same one that passed the check on the line before. I think you need to read the value and the value portion together.

Sources/Action.swift
```diff
@@ -75,6 +75,10 @@ public final class Action<Input, Output, Error: Swift.Error> {
 	///   - execute: A closure that returns the `SignalProducer` returned by
 	///              calling `apply(Input)` on the action, optionally using
 	///              the current value of `state`.
+	public init<State: PropertyProtocol>(state property: State, enabledIf isEnabled: @escaping (State.Value) -> Bool, _ execute: @escaping (State.Value, Input) -> Result<Output, Error>) {
+		fatalError()
+	}
```
mdiep: What's this doing here?

Sources/Property.swift
```diff
@@ -50,6 +50,12 @@ extension MutablePropertyProtocol {
 	}
 }
 
+/// Represents a mutable property that can be composed.
```
mdiep: Can you document this a bit more explicitly? (i.e. explain what it means for a property to be composed.)

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,294 @@
...
+	public var value: Value {
+		get { return getter() }
+		set { setter(newValue) }
+	}
+
+	public let producer: SignalProducer<Value, NoError>
+
+	public let signal: Signal<Value, NoError>
+
+	/// The lifetime of the editor.
```
mdiep: s/editor/property/

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,294 @@
...
+	public let lifetime: Lifetime
+
+	/// Create an `MutableValidatingProperty` that presents a mutable validating
```
mdiep: s/an/a/

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,294 @@
...
+	/// `result` property of `self`.
+	///
+	/// - parameters:
+	///   - inner: The inner property which validated values are committed to.
+	///   - validator: The closure to invoke for any proposed value to `self`.
+	public init<Inner: ComposableMutablePropertyProtocol>(
+		_ inner: Inner,
+		_ validator: @escaping (Value) -> ValidatorOutput<Value, ValidationError>
+	) where Inner.Value == Value {
+		var mutesValueBackpropagation = false
```
mdiep: This variable deserves a comment. It's not immediately obvious what it does from the name.

mdiep: Maybe something like `isSettingInnerValue` would be a better name? Describe what's happening—not the effect.

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,294 @@
...
+				if let value = writebackValue {
+					mutesValueBackpropagation = true
+					inner.value = value
+					mutesValueBackpropagation = false
+				}
+			})
+		}
+	}
+
+	/// Create an `MutableValidatingProperty` that validates mutations before
```
mdiep: s/an/a/

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,294 @@
...
+	/// committing them.
+	///
+	/// If `success` is returned by `validator`, the associated value would be
+	/// committed to `inner`. Otherwise, the failure would be exposed by the
```
mdiep: `inner` doesn't exist here

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,294 @@
...
+	/// `result` property of `self`.
+	///
+	/// - parameters:
+	///   - initial: The initial value of the property. It is not required to
+	///              pass the validation as specified by `validator`.
```
mdiep: It seems like we should maybe require this? That seems very weird.

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,294 @@
...
+	///   - validator: The closure to invoke for any proposed value to `self`.
+	public convenience init(
+		_ initial: Value,
+		_ validator: @escaping (Value) -> ValidatorOutput<Value, ValidationError>
+	) {
+		self.init(MutableProperty(initial), validator)
+	}
+
+	/// Create an `MutableValidatingProperty` that presents a mutable validating
```
mdiep: s/an/a/

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,294 @@
...
+					case let .substitution(substitutedValue, _, _):
+						s.value = substitutedValue
+					case let .success(value):
+						s.value = value
+					}
+				}
+		}
+	}
+
+	/// Create an `MutableValidatingProperty` that validates mutations before
```
mdiep: s/an/a/

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,294 @@
...
+	/// committing them.
+	///
+	/// If `success` is returned by `validator`, the associated value would be
+	/// committed to `inner`. Otherwise, the failure would be exposed by the
```
mdiep: `inner` doesn't exist here

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,294 @@
...
+	///   - validator: The closure to invoke for any proposed value to `self`.
+	public convenience init<Other: PropertyProtocol>(
+		_ initial: Value,
+		with other: Other,
+		_ validator: @escaping (Value, Other.Value) -> ValidatorOutput<Value, ValidationError>
+	) {
+		self.init(MutableProperty(initial), with: other, validator)
+	}
+
+	/// Create an `MutableValidatingProperty` that presents a mutable validating
```
mdiep: s/an/a/

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,294 @@
...
+	///   - validator: The closure to invoke for any proposed value to `self`.
+	public convenience init<U, E: Swift.Error>(
+		_ inner: MutableProperty<Value>,
+		with other: MutableValidatingProperty<U, E>,
+		_ validator: @escaping (Value, U) -> ValidatorOutput<Value, ValidationError>
+	) {
+		self.init(inner, with: other, validator)
+	}
+
+	/// Create an `MutableValidatingProperty` that validates mutations before
```
mdiep: s/an/a/

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,294 @@
...
+	/// committing them.
+	///
+	/// If `success` is returned by `validator`, the associated value would be
+	/// committed to `inner`. Otherwise, the failure would be exposed by the
```
mdiep: `inner` doesn't exist here

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,294 @@
...
+		}
+	}
+}
+
+public enum ValidatorOutput<Value, Error: Swift.Error> {
+	case success
+
+	case substitution(Value, Error)
+
+	case failure(Error)
```
mdiep: These could use documentation

liscio:
> @liscio are you happy with the changes?

@RuiAAPeres Do I ever seem happy with changes? ;)

@andersio and I spoke briefly on the Slack about whether `Error` is appropriate for the replaced value case, or if instead that scenario should be eliminated in favor of the API user supplying their own `Value` that allows for replacement. But the discussion didn't really result in anything constructive (my fault—been swamped.)

My worry specifically was that appending to the "Error Space" isn't going to work if you consider that suddenly you need to handle that kind of error everywhere something can go wrong (even though this represents that you've just worked around the error, and it's no longer a problem.)

So, yeah—I think the replacement case might have to get backed out after all, and I'll have to think on an alternative for that scenario. Maybe mapping errors through another adapter that "massages" bad values is a better approach. (Not part of this PR/construct, though—it might have to exist simply as a "recipe" in the future.)

andersio:
@liscio It wasn't required to be handled though. The substituted state is just an aggregate of the proposed value, the substituted value, and an optional error context. Each pieces of them can very well be ignored at one's will.

Note that https://github.com/ReactiveCocoa/ReactiveSwift/pull/182/commits/09022641487305ed17b92824455967d98dcc776e has introduced the notion to replace a failed value with a predefined value.

Sources/ValidatingProperty.swift
andersio: Consider a form field that requires at least three characters. Initially it has to be invalid.

liscio:
> The substituted state is just an aggregate of the proposed value, the substituted value, and an optional error context.

@andersio While the error is optional, it shares the same `Error` type as the `.error` case which is what I am feeling strange about. If you choose to replace the value, the context supplied with that replacement must be the same kind of error as an actual failure.

I don't know that this is actually a problem or not, so I think for the interest of pushing this forward I'll say that making the error optional is probably a good compromise for now. We can adjust this later on if we find it troublesome in practice.

Other than @mdiep's pending documentation requests I should be good for this going in.

Sources/Property.swift
```diff
@@ -50,6 +50,28 @@ extension MutablePropertyProtocol {
 	}
 }
 
+/// Represents a mutable property that can be safety composed by exposing its
+/// synchronization mechanic through the defined closure-based interface.
+public protocol ComposableMutablePropertyProtocol: MutablePropertyProtocol {
```
mdiep: Should we move this into `MutablePropertyProtocol` in 2.0?

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,307 @@
...
+		if case .failure = self {
+			return true
+		} else {
+			return false
+		}
+	}
+
+	/// Extract the value if the validation is passed, or a substituted value
+	/// for failure is provided.
+	public var value: Value? {
```
mdiep: Do you have a use case in mind for this?

I think this maybe shouldn't return an `Optional`. You could still check `isFailure` if you didn't want failed values. This would simplify the code that handles dependencies, e.g.

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,307 @@
...
+				if let s = self {
+					switch s.result.value {
+					case let .failure(failedValue, _):
+						s.value = failedValue
+					case let .substitution(substitutedValue, _, _):
+						s.value = substitutedValue
+					case let .success(value):
+						s.value = value
+					}
+				}
```
mdiep: It took me 5-10 minutes to understand what was happening here. We should add a comment here (and below) explaining why `s.value` is being set.

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,307 @@
...
+		let other = Property(other)
+
+		self.init(inner) { input in
+			return validator(input, other.value)
+		}
+
+		other.signal
+			.take(during: lifetime)
+			.observeValues { [weak self] _ in
+				if let s = self {
```
mdiep: This would probably be better as a `guard`. (Same below.)

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,307 @@
...
+	///   - initial: The initial value of the property. It is not required to
+	///              pass the validation as specified by `validator`.
+	///   - other: The property that `validator` depends on.
+	///   - validator: The closure to invoke for any proposed value to `self`.
+	public convenience init<U, E: Swift.Error>(
+		_ initial: Value,
+		with other: MutableValidatingProperty<U, E>,
+		_ validator: @escaping (Value, U) -> ValidatorOutput<Value, ValidationError>
+	) {
+		let otherValidations = other.result
```
mdiep: Is there a reason that this is outside the `init`? If so, can you add a comment explaining why?

Sources/ValidatingProperty.swift
```diff
@@ -0,0 +1,307 @@
...
+			if let s = self {
+				switch s.result.value {
+				case let .failure(failedValue, _):
+					s.value = failedValue
+				case let .substitution(substitutedValue, _, _):
+					s.value = substitutedValue
+				case let .success(value):
+					s.value = value
+				}
+			}
```
mdiep: Indentation is off for this block.

Sources/Property.swift
andersio: I'd think so.

Sources/ValidatingProperty.swift
andersio: It just makes it easier to get a valid value during composition without case switching, akin to `Result.value`.

Sources/ValidatingProperty.swift
andersio: Capturing only `other.result` but not `other`.

mdiep:
Thanks for all your work on this! Sorry I was so slow with reviews. 🙈

PR #191, opened 2016-12-26
Swift Orange brand assets.

andersio:
Preview: https://github.com/ReactiveCocoa/ReactiveSwift/blob/b0dedfdd702e750cfe6089a5595ca6adcbc5687c/README.md


mdiep:
1. Do we need to keep the blue logos around?

2. The _Documentation_ and _Join us on Slack_ ones seem like they should be darker or less pastel—more like the Swift orange. (With white text if necessary.)

andersio:
> Do we need to keep the blue logos around?

Removed.

> The Documentation and Join us on Slack ones seem like they should be darker or less pastel—more like the Swift orange. (With white text if necessary.)

I'd prefer them to be more pastel though for contrasting with the logo that is already heavy in color.

PR #195, opened 2016-12-28
[skip ci] Minor fixes in documentation

dmcrodrigues:
I caught an example that was still assuming the semantics of a cold signal after the migration to use hot signals (`Signal`).


mdiep:
Thanks!

dcaunt:
Do you think there is any value in adding some or all of the content in this file to the playground?

Issue #197, opened 2016-12-29
Port groupBy/uncollect from Rex to ReactiveSwift?


ynnckcmprnl:
[Rex](https://github.com/RACCommunity/Rex) contained a [groupBy](https://github.com/RACCommunity/Rex/blob/0.12.0/Source/SignalProducer.swift#L18) extension, which came in handy occasionally. With most of Rex' functionality moved to ReactiveCocoa, would it make sense to add/port the groupBy extension to ReactiveSwift?

I had a go myself at updating the groupBy extension to ReactiveSwift 1.0 & Swift 3.0 but couldn't quite make out how to address the removal of `SignalProducer.buffer(_ capacity: Int)`.

Another extension which might be worth adding/porting is uncollect, I can open a pull request if desired.

andersio:
Would you mind to provide a reduced use case for both operators? New operators are definitely welcomed, just that we wanna keep it tight and straight.

`buffer(upTo:)` can be mimicked by:
```swift
let (signal, observer) = Signal<U, Error>.pipe()
let producer = SignalProducer(signal).replayLazily(upTo: n)
producer.start().dispose()
```

andersio:
`uncollect` already exists as an overload of `flatten`, by the way.

dmcrodrigues:
Regarding `uncollect` an almost equivalent is `flatten` available in ReactiveSwift for both `Signal` and `SignalProducer`. However, there's a constraint about the error type, `Error == NoError`, which is not defined on `uncollect` and removing it leads to an _Ambiguous use of 'flatMap(_:transform:)' which kind of makes sense since there are two versions of `flatMap` to accommodate both `Error` and `NoError`. I don't know if this was the reason behind this constraint but I don't see a reason to restrict this operator only for signals without errors.

Also, since we are talking about `flatten` as an alternative to `uncollect`, I'm having some difficulty to understand how relevant a strategy is to flatten a sequence, between picking `latest`, `merge`, or `concat` there's any difference?

andersio:
Agree. The strategies do not matter when the inner sequence is immediate.

dmcrodrigues:
@andersio If that's the case, maybe we can remove the strategy from `flatten` when the value is a `Sequence`? Introduces unnecessary confusion around the usage of the operator and can be misleading.

mdiep:
> If that's the case, maybe we can remove the strategy from flatten when the value is a Sequence?

[...]

PR #200, opened 2016-12-30
Port `groupBy` from REX

iv-mexx:
As per #197, this is a Port of `groupBy` from [REX](https://github.com/RACCommunity/Rex/pull/159/).

I've only changed it to use `Atomic` instead of the `NSRecursiveLock`

### DIFF
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -1942,47 +1942,41 @@ extension SignalProducerProtocol {
 	/// - returns: A producer of producers amits one producer for each group and forwards
 	///            each value from the original producer to the inner producer corresponding 
 	///            to the group to which the value belongs to (as determined by the key)
-	public func group<Key: Hashable>(by grouping: @escaping (Value) -> Key) -> SignalProducer<(Key, SignalProducer<Value, Error>), Error> {
 	public func groupBy<Key: Hashable>(_ grouping: @escaping (Value) -> Key) -> SignalProducer<(Key, SignalProducer<Value, Error>), Error> {
 		return SignalProducer<(Key, SignalProducer<Value, Error>), Error> { observer, disposable in
-			var groups: [Key: Signal<Value, Error>.Observer] = [:]
-			
-			let lock = NSRecursiveLock()
-			lock.name = "me.neilpa.rex.groupBy"
+			let groups = Atomic<[Key: Signal<Value, Error>.Observer]>([:])
 			
 			self.start { event in
 				switch event {
 				case let .value(value):
 					let key = grouping(value)
-					
-					lock.lock()
-					var group = groups[key]
-					if group == nil {
-						let (signal, innerObserver) = Signal<Value, Error>.pipe()
-						let producer = SignalProducer(signal).replayLazily(upTo: Int.max)
-						
-						// Start the buffering immediately.
-						producer.start()
-						observer.send(value: (key, producer))
-						
-						groups[key] = innerObserver
-						group = innerObserver
+					var group: Signal<Value, Error>.Observer?
+					groups.modify {
+						group = $0[key]
+						if group == nil {
+							let (signal, innerObserver) = Signal<Value, Error>.pipe()
+							let producer = SignalProducer(signal).replayLazily(upTo: Int.max)
+							
+							// Start the buffering immediately.
+							producer.start()
+							observer.send(value: (key, producer))
+							
+							$0[key] = innerObserver
+							group = innerObserver
+						}
 					}
-					lock.unlock()
-					
-					group!.send(value: value)
-					
+					group!.send(value: value)					
 				case let .failed(error):
 					observer.send(error: error)
-					groups.values.forEach { $0.send(error: error) }
+					groups.value.values.forEach { $0.send(error: error) }
 					
 				case .completed:
 					observer.sendCompleted()
-					groups.values.forEach { $0.sendCompleted() }
+					groups.value.values.forEach { $0.sendCompleted() }
 					
 				case .interrupted:
 					observer.sendInterrupted()
-					groups.values.forEach { $0.sendInterrupted() }
+					groups.value.values.forEach { $0.sendInterrupted() }
 				}
 			}
 		}
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -1943,7 +1943,7 @@ extension SignalProducerProtocol {
 	///            each value from the original producer to the inner producer corresponding 
 	///            to the group to which the value belongs to (as determined by the key)
 	public func group<Key: Hashable>(by grouping: @escaping (Value) -> Key) -> SignalProducer<(Key, SignalProducer<Value, Error>), Error> {
-		return SignalProducer<(Key, SignalProducer<Value, Error>), Error> { observer, disposable in
+		return SignalProducer { observer, disposable in
 			let groups = Atomic<[Key: Signal<Value, Error>.Observer]>([:])
 			
 			disposable += self.start { event in
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -1950,10 +1950,10 @@ extension SignalProducerProtocol {
 				switch event {
 				case let .value(value):
 					let key = grouping(value)
-					var group: Signal<Value, Error>.Observer?
-					groups.modify {
-						group = $0[key]
-						if group == nil {
+					let group: Signal<Value, Error>.Observer = groups.modify { groups in
+						if let group = groups[key] {
+							return group
+						} else {
 							let (signal, innerObserver) = Signal<Value, Error>.pipe()
 							let producer = SignalProducer(signal).replayLazily(upTo: Int.max)
 							
@@ -1961,11 +1961,12 @@ extension SignalProducerProtocol {
 							producer.start().dispose()
 							observer.send(value: (key, producer))
 							
-							$0[key] = innerObserver
-							group = innerObserver
+							groups[key] = innerObserver
+							return innerObserver
 						}
 					}
-					group!.send(value: value)					
+					group.send(value: value)
+					
 				case let .failed(error):
 					observer.send(error: error)
 					groups.value.values.forEach { $0.send(error: error) }
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -1942,7 +1942,7 @@ extension SignalProducerProtocol {
 	/// - returns: A producer of producers amits one producer for each group and forwards
 	///            each value from the original producer to the inner producer corresponding 
 	///            to the group to which the value belongs to (as determined by the key)
-	public func groupBy<Key: Hashable>(_ grouping: @escaping (Value) -> Key) -> SignalProducer<(Key, SignalProducer<Value, Error>), Error> {
+	public func group<Key: Hashable>(by grouping: @escaping (Value) -> Key) -> SignalProducer<(Key, SignalProducer<Value, Error>), Error> {
 		return SignalProducer<(Key, SignalProducer<Value, Error>), Error> { observer, disposable in
 			let groups = Atomic<[Key: Signal<Value, Error>.Observer]>([:])
 			
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -1929,6 +1929,64 @@ extension SignalProducerProtocol {
 			}
 		}
 	}
+	
+	/// Creates a new `SignalProducer` that will bucket each received value into
+	/// a group based on the key returned from `grouping`. 
+	///
+	/// Termination events on the original signal are also forwarded to each
+	/// producer group.
+	///
+	/// - parameters:
+	///   - grouping: a closure that determines the grouping key for a given value
+	///
+	/// - returns: A producer of producers amits one producer for each group and forwards
+	///            each value from the original producer to the inner producer corresponding 
+	///            to the group to which the value belongs to (as determined by the key)
+	public func group<Key: Hashable>(by grouping: @escaping (Value) -> Key) -> SignalProducer<(Key, SignalProducer<Value, Error>), Error> {
+	public func groupBy<Key: Hashable>(_ grouping: @escaping (Value) -> Key) -> SignalProducer<(Key, SignalProducer<Value, Error>), Error> {
+		return SignalProducer<(Key, SignalProducer<Value, Error>), Error> { observer, disposable in
+			var groups: [Key: Signal<Value, Error>.Observer] = [:]
+			
+			let lock = NSRecursiveLock()
+			lock.name = "me.neilpa.rex.groupBy"
+			
+			self.start { event in
+				switch event {
+				case let .value(value):
+					let key = grouping(value)
+					
+					lock.lock()
+					var group = groups[key]
+					if group == nil {
+						let (signal, innerObserver) = Signal<Value, Error>.pipe()
+						let producer = SignalProducer(signal).replayLazily(upTo: Int.max)
+						
+						// Start the buffering immediately.
+						producer.start()
+						observer.send(value: (key, producer))
+						
+						groups[key] = innerObserver
+						group = innerObserver
+					}
+					lock.unlock()
+					
+					group!.send(value: value)
+					
+				case let .failed(error):
+					observer.send(error: error)
+					groups.values.forEach { $0.send(error: error) }
+					
+				case .completed:
+					observer.sendCompleted()
+					groups.values.forEach { $0.sendCompleted() }
+					
+				case .interrupted:
+					observer.sendInterrupted()
+					groups.values.forEach { $0.sendInterrupted() }
+				}
+			}
+		}
+	}
 }
 
 extension SignalProducerProtocol where Value == Bool {
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -1946,7 +1946,7 @@ extension SignalProducerProtocol {
 		return SignalProducer<(Key, SignalProducer<Value, Error>), Error> { observer, disposable in
 			let groups = Atomic<[Key: Signal<Value, Error>.Observer]>([:])
 			
-			self.start { event in
+			disposable += self.start { event in
 				switch event {
 				case let .value(value):
 					let key = grouping(value)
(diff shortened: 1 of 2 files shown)


andersio:
I wonder if it makes better sense to return a `Property<[Key: Value]>`.

Sources/SignalProducer.swift
```diff
@@ -1877,6 +1877,58 @@ extension SignalProducerProtocol {
...
+					let key = grouping(value)
+					var group: Signal<Value, Error>.Observer?
+					groups.modify {
+						group = $0[key]
+						if group == nil {
+							let (signal, innerObserver) = Signal<Value, Error>.pipe()
+							let producer = SignalProducer(signal).replayLazily(upTo: Int.max)
+							
+							// Start the buffering immediately.
+							producer.start()
```
andersio: This can be disposed.

Sources/SignalProducer.swift
```diff
@@ -1877,6 +1877,58 @@ extension SignalProducerProtocol {
...
+		return SignalProducer<(Key, SignalProducer<Value, Error>), Error> { observer, disposable in
+			let groups = Atomic<[Key: Signal<Value, Error>.Observer]>([:])
+			
+			self.start { event in
+				switch event {
+				case let .value(value):
+					let key = grouping(value)
+					var group: Signal<Value, Error>.Observer?
+					groups.modify {
+						group = $0[key]
```
andersio: [Nit] For multi-line modifications, it'd be better to use named parameter.

Sources/SignalProducer.swift
```diff
@@ -1877,6 +1877,58 @@ extension SignalProducerProtocol {
...
+	///            to the group to which the value belongs to (as determined by the key)
+	public func groupBy<Key: Hashable>(_ grouping: @escaping (Value) -> Key) -> SignalProducer<(Key, SignalProducer<Value, Error>), Error> {
+		return SignalProducer<(Key, SignalProducer<Value, Error>), Error> { observer, disposable in
+			let groups = Atomic<[Key: Signal<Value, Error>.Observer]>([:])
+			
+			self.start { event in
+				switch event {
+				case let .value(value):
+					let key = grouping(value)
+					var group: Signal<Value, Error>.Observer?
```
andersio: You can dodge the need of an optional by using:
```swift
let group: Signal<Value, Error>.Observer = groups.modify {
```

Sources/SignalProducer.swift
```diff
@@ -1877,6 +1877,58 @@ extension SignalProducerProtocol {
...
+							// Start the buffering immediately.
+							producer.start()
+							observer.send(value: (key, producer))
+							
+							$0[key] = innerObserver
+							group = innerObserver
+						}
+					}
+					group!.send(value: value)					
+				case let .failed(error):
```
andersio: [Nit] Blank line before the pattern.

Sources/SignalProducer.swift
```diff
@@ -1877,6 +1877,58 @@ extension SignalProducerProtocol {
...
+	///   - grouping: a closure that determines the grouping key for a given value
+	///
+	/// - returns: A producer of producers amits one producer for each group and forwards
+	///            each value from the original producer to the inner producer corresponding 
+	///            to the group to which the value belongs to (as determined by the key)
+	public func groupBy<Key: Hashable>(_ grouping: @escaping (Value) -> Key) -> SignalProducer<(Key, SignalProducer<Value, Error>), Error> {
+		return SignalProducer<(Key, SignalProducer<Value, Error>), Error> { observer, disposable in
+			let groups = Atomic<[Key: Signal<Value, Error>.Observer]>([:])
+			
+			self.start { event in
```
andersio: The start disposable should be added to the producer disposable.

Sources/SignalProducer.swift
```diff
@@ -1877,6 +1877,58 @@ extension SignalProducerProtocol {
...
+	///
+	/// - parameters:
+	///   - grouping: a closure that determines the grouping key for a given value
+	///
+	/// - returns: A producer of producers amits one producer for each group and forwards
+	///            each value from the original producer to the inner producer corresponding 
+	///            to the group to which the value belongs to (as determined by the key)
+	public func groupBy<Key: Hashable>(_ grouping: @escaping (Value) -> Key) -> SignalProducer<(Key, SignalProducer<Value, Error>), Error> {
+		return SignalProducer<(Key, SignalProducer<Value, Error>), Error> { observer, disposable in
+			let groups = Atomic<[Key: Signal<Value, Error>.Observer]>([:])
```
andersio: It doesn't seem necessary to specialise it. Is it a workaround to the compiler's weirdness or something?

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
```diff
@@ -1666,5 +1666,47 @@ class SignalProducerLiftingSpec: QuickSpec {
 				expect(latestValues?.1) == 2
 			}
 		}
+		
+		describe("groupBy") {
+			let (signal, observer) = Signal<Int, NoError>.pipe()
```
andersio: Please wrap test cases in `it`.

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
```diff
@@ -1666,5 +1666,47 @@ class SignalProducerLiftingSpec: QuickSpec {
...
+			let producer = SignalProducer<Int, NoError>(signal)
+			var evens: [Int] = []
+			var odds: [Int] = []
+			let disposable = CompositeDisposable()
+			var interrupted = false
+			var completed = false
+			
+			disposable += producer
+				.groupBy { $0 % 2 == 0 }
+				.start(Observer(value: { key, group in
```
andersio: Feels like `start` that takes an event action would be nicer here.

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
```diff
@@ -1666,5 +1666,47 @@ class SignalProducerLiftingSpec: QuickSpec {
...
+			expect(odds) == [1]
+			
+			observer.send(value: 3)
+			expect(evens) == [2]
+			expect(odds) == [1, 3]
+			
+			disposable.dispose()
+			
+			observer.send(value: 1)
+			expect(interrupted) == true
```
andersio: These should be checked immediately after the disposal.

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
```diff
@@ -1666,5 +1666,47 @@ class SignalProducerLiftingSpec: QuickSpec {
 				expect(latestValues?.1) == 2
 			}
 		}
+		
+		describe("groupBy") {
```
andersio: It would be great if there are more test cases. At least this single test cases apparently did not catch the undisposed upstreams, caused by [this line](https://github.com/ReactiveCocoa/ReactiveSwift/pull/200/files#diff-ceb9f0d4c919bbf25ca4e3c8b7036b8bR1897).

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
iv-mexx: woops

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
iv-mexx: yeah right, i've split this up into 2 separated cases and extended them to check the lifetime of the inner producers

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
iv-mexx: alright! 👍

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
andersio: Perhaps one per terminal event, hmm?

Sources/SignalProducer.swift
iv-mexx: sure

Sources/SignalProducer.swift
iv-mexx: done

Sources/SignalProducer.swift
iv-mexx: nice, didn't know about this!

Sources/SignalProducer.swift
iv-mexx: I'm afraid I don't know what exactly you mean.

Sources/SignalProducer.swift
iv-mexx: fixed

Sources/SignalProducer.swift
iv-mexx: ok 👍

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
iv-mexx: Sure I can do that. 
Should we check for all events that only this event is sent (e.g. when sending completed, also check that interrupted was not received), or can we trust that when `completed` is received, `interrupted` can not be received?

andersio: I think asserting just the specific event is enough. The `Signal` contract guarantees only one would ever be sent. If multiple terminal events are sent, that is a fault in the basics and should have been caught somewhere else.

Sources/SignalProducer.swift
andersio: I mean if you really need to write `SignalProducer<(Key, SignalProducer<Value, Error>), Error>` when it can be inferred.

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
```diff
@@ -1666,5 +1666,102 @@ class SignalProducerLiftingSpec: QuickSpec {
...
+				expect(evens) == [2]
+				expect(odds) == [1, 3]
+				
+				disposable.dispose()
+				
+				observer.send(value: 4)
+				expect(evens) == [2]
+				expect(odds) == [1, 3]
+			}
+			it("should terminate correctly on disposal") {
```
andersio: [Nit] A blank line please. 🙏

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
```diff
@@ -1666,5 +1666,102 @@ class SignalProducerLiftingSpec: QuickSpec {
...
+				
+				disposable += producer
+					.groupBy { $0 % 2 == 0 }
+					.startWithValues { key, group in
+						if key {
+							group.startWithValues { evens.append($0)}
+						} else {
+							group.startWithValues { odds.append($0)}
+						}
+				}
```
andersio: [Nit] This close bracket could be aligned better.

Sources/SignalProducer.swift
```diff
@@ -1877,6 +1877,59 @@ extension SignalProducerProtocol {
...
+	/// Termination events on the original signal are also forwarded to each
+	/// producer group.
+	///
+	/// - parameters:
+	///   - grouping: a closure that determines the grouping key for a given value
+	///
+	/// - returns: A producer of producers amits one producer for each group and forwards
+	///            each value from the original producer to the inner producer corresponding 
+	///            to the group to which the value belongs to (as determined by the key)
+	public func groupBy<Key: Hashable>(_ grouping: @escaping (Value) -> Key) -> SignalProducer<(Key, SignalProducer<Value, Error>), Error> {
```
andersio: By the API convention though, it seems it should be `group(by:)`. The parameter perhaps should be named `classifier` or something?

iv-mexx: I've tried that for a short moment, but on the call site that would then look like `producer.group { ... }` instead of `producer.groupBy { ... } `when using trailing closure syntax. 
Not sure if thats a problem though...

andersio: We already have `skip(while:)` and `take(while:)` working in this way.

iv-mexx: Alright! 👍

Sources/SignalProducer.swift
iv-mexx: Oh that line. Yeah that works without explicit annotations 👍

iv-mexx:
@andersio all done 😃

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
iv-mexx: Alright

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
```diff
@@ -1749,19 +1727,132 @@ class SignalProducerLiftingSpec: QuickSpec {
...
+			observer.send(value: 2)
+			observer.sendCompleted()
+			
+			expect(completed) == true
+			expect(interrupted) == false
+			expect(evenCompleted) == true
+			expect(oddCompleted) == true
+		}
+		
+		it("should terminate correctly receiving an error event") {
```
andersio: s/error/failed/

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
```diff
@@ -1749,19 +1727,132 @@ class SignalProducerLiftingSpec: QuickSpec {
 						case .failed:
 							break
 						}
-					}
+				}
```
andersio: I actually meant that the bracket should aligned with `.start`. Sorry if I wasn't clear enough. 😸

iv-mexx:
@mdiep ping

mdiep:
Oops! Sorry @iv-mexx. I put this back on my todo list and will get to it this week.

Sources/SignalProducer.swift
```diff
@@ -1877,6 +1877,59 @@ extension SignalProducerProtocol {
...
+			disposable += self.start { event in
+				switch event {
+				case let .value(value):
+					let key = grouping(value)
+					let group: Signal<Value, Error>.Observer = groups.modify { groups in
+						if let group = groups[key] {
+							return group
+						} else {
+							let (signal, innerObserver) = Signal<Value, Error>.pipe()
+							let producer = SignalProducer(signal).replayLazily(upTo: Int.max)
```
mdiep: Is replaying values an essential part of `group(by:)`? This is my primary hesitation. In general, we try to discourage the use of replaying (which is why `buffer` was removed).

(You could just as easily send the `Signal` and not send values _on_ the `Signal` until it's been sent—giving observers a chance to observe the signal.)

If you feel that it is, I'd love to see a larger code sample that demonstrates why.

Sources/SignalProducer.swift
```diff
@@ -1877,6 +1877,59 @@ extension SignalProducerProtocol {
...
+	/// Creates a new `SignalProducer` that will bucket each received value into
+	/// a group based on the key returned from `grouping`. 
+	///
+	/// Termination events on the original signal are also forwarded to each
+	/// producer group.
+	///
+	/// - parameters:
+	///   - grouping: a closure that determines the grouping key for a given value
+	///
+	/// - returns: A producer of producers amits one producer for each group and forwards
```
neilpa: typo `amits`

Sources/SignalProducer.swift
neilpa: I _think_ so, otherwise consumers could miss events without realizing it. Hence why I erred on the side of buffering. I could be convinced otherwise, but I'm not sure how to best communicate those semantics at the operator level.

mdiep: I think returning `Signal`s instead of `SignalProducer`s communicates the semantics.

But I think seeing an actual example of this operator will make it easier to see whether that's a viable change.

iv-mexx: I've posted an example how we use the `groupBy` in a project in the Issue: https://github.com/ReactiveCocoa/ReactiveSwift/issues/197#issuecomment-269752875

iv-mexx: If I'm reading this correctly, [RxJS](https://github.com/ReactiveX/rxjs/blob/master/src/operator/groupBy.ts) is using Hot Observables for the groups (via `refCount)

NachoSoto: As much as I agree with @mdiep about the semantics, in my experience having buffering in this operator (specifically `RxJava`'s `Observable#groupBy`) has been valuable .

andersio: Perhaps should the buffering be an option?

andersio: @mdiep 

> (You could just as easily send the Signal and not send values on the Signal until it's been sent—giving observers a chance to observe the signal.)

While this is the ideal case, it seems the buffering is essential, say when using with `flatten(.concat)`, which can delay the establishment of the observation.

mdiep:
I've been thinking about this quite a bit. The replaying of values makes me pretty uncomfortable, so I've felt a bit stuck and unsure how to resolve this.

But after thinking about it, I no longer think that `groupBy` makes sense as a built-in operator. 😞 It requires semantics that we've said we'd like to discourage (for good reason, I think). And the example can be resolved by introducing a `class`, which implies the correct semantics.

```swift
class BluetoothManager {
  private let allBeaconsSignal: Signal<CLBeacon, NoError>

  private var _beacons: [(Int, Int): (Signal<_, _>, Signal<_, _>.Observer)]
  public let beacons: Signal<(CLBeacon, Signal<_, _>), NoError>
  private let observer: Signal<(CLBeacon, Signal<_, _>), NoError>.Observer

  init() {
     // I'm not exactly sure how this should all work. I was working off the given example code,
     // which is tightly focused on the groupBy. But this will hopefully communicate the idea,
     // which is that the class is responsible for grouping values, notifying interested parties
     // that new values have come in and giving them an opportunity to subscribe to events.
     let (beacons, observer) = Signal<(CLBeacon, Signal), NoError>.pipe()
     allBeaconsSignal.observeValues { beacon in
       let identifier = (beacon.major, beacon.minor)
       if _beacons[identifier] == nil {
         let (signal, observer) = Signal<_, _>.pipe()
         self.observer.send(value: (beacon, signal))
       }
     }
  }
}
```

But if the rest of the core team feels that this _does_ make sense to add, they can definitely override me. 😄

andersio:
In lieu of the rise of unidirectional data flow in the Swift community, perhaps the class of problems intended for `groupBy` is better solved that way.

Closing because of inactivity.

PR #202, opened 2016-12-30
combineLatest(with:) drops events sent synchronously on start

sharplet:
When multiple events are sent synchronously in a `SignalProducer`'s start handler, `combineLatest(with:)` appears to be dropping events and only emitting the latest.

The test fails with

```
expected to equal <[1-1, 2-1]>, got <[2-1]>
```

### DIFF
--- a/Tests/ReactiveSwiftTests/SignalProducerSpec.swift
+++ b/Tests/ReactiveSwiftTests/SignalProducerSpec.swift
@@ -780,6 +780,18 @@ class SignalProducerSpec: QuickSpec {
 				expect(ids) == [1, 2]
 				expect(values._bridgeToObjectiveC()) == [[1, 2]]._bridgeToObjectiveC()
 			}
+
+			it("should combine synchronous events") {
+				var values: [String] = []
+
+				SignalProducer<Int, NoError>([1, 2])
+					.combineLatest(with: SignalProducer(value: 1))
+					.startWithValues { a, b in
+						values.append("\(a)-\(b)")
+					}
+
+				expect(values) == ["1-1", "2-1"]
+			}
 		}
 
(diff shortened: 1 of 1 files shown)


sharplet:
So I believe the issue is a race in https://github.com/ReactiveCocoa/ReactiveSwift/blob/a458a5a754a2fe87ec16cb669f6beb5ecdf0fd02/Sources/Signal.swift#L903-L904:

```swift
			disposable += self.observeWithStates(signalState, otherState, lock, observer)
			disposable += other.observeWithStates(otherState, signalState, lock, observer)
```

The two streams aren't serialized with respect to each other, so when the first is observed, if there are multiple synchronous events, they are lost before the second observation is started.

andersio:
It looks more like an issue due to the two start handlers being invoked synchronously. At the `Signal` level, it is correct.

sharplet:
If I rewrite the test like so, it passes:

```swift
it("should combine synchronous events") {
	var values: [String] = []

	SignalProducer<Int, NoError>([1])
		.combineLatest(with: SignalProducer([1, 2]))
		.startWithValues { a, b in
			values.append("\(a)-\(b)")
		}

	expect(values) == ["1-1", "1-2"]
}
```

andersio:
`combineLatest` uses `liftLeft`. So in the failing one, the 2-element sequence would start first.

sharplet:
I'm going to close this for now, now that I understand how ordering affects the semantics of `combineLatest`. In the particular case I was working on, flipping the order of the producers does resolve my problem.

andersio:
It think it worths mentioning in the documentation that `combineLatest` picks only the latest value emitted by the start handler of the LHS producer.

Issue #97, opened 2016-11-15
Replaying success but not errors


sharplet:
In my current project I have a 3-step process where the output of each signal producer becomes the input of the next. This process should externally behave like a single atomic operation, essentially a state machine. If any step fails, then retrying the process should start again from the failed step, but use the cached output from previously successful steps.

I'd really like to use `replayLazily()` for this, but can't because it will also replay failures. I haven't been able to work out a way to compose existing operators to get the semantics I'm looking for.

Is this something that could use a new operator? Is there a way I can get this with existing operators?

I kind of want `replayLazily`, only failure events should invalidate the cache instead of being replayed.

andersio:
A similar issue: https://github.com/ReactiveCocoa/ReactiveSwift/issues/83

mdiep:
Can you give a little more detail about what you're trying to do?

sharplet:
The idea is basically to create a state machine, where an action will restart from the last point it failed:

```swift
let step1: SignalProducer<A, AnyError>
let step2: (A) -> SignalProducer<B, AnyError>
let step3: (B) -> SignalProducer<C, AnyError>

let task = step1
  .replaySuccessLazily() // don't love this name
  .flatMap(.concat, transform: step2)
  .replaySuccessLazily()
  .flatMap(.concat, transform: step3)
  .replaySuccessLazily()

let action = Action { _ in task }
```

andersio:
It seems the prominent need here is a mean to "fail gracefully" by invalidating the cached values and restarting the producer. Composing with`retry(upTo:)` can do the later but not the former.

It is like how #83 asked for a producer that invalidates the cached image upon a timeout event, and restart the underlying producer.

I'd say we may introduce a specialized `replayLazily` overload on `SignalProducerProtocol where Error == ReplayCacheInvalidation`.

mdiep:
It seems like you should move the retries into the steps.

```swift
let step1: SignalProducer<A, AnyError>
let step2: (A) -> SignalProducer<B, AnyError>
let step3: (B) -> SignalProducer<C, AnyError>

let task = step1
  .retry(upTo: 3)
  .flatMap(.concat) { step2($0).retry(upTo: 3) }
  .flatMap(.concat) { step3($0).retry(upTo: 3) }

let action = Action { _ in task }
```

Would that work?

sharplet:
[...]

PR #203, opened 2016-12-30
Allow replayed producers to be restarted after termination

sharplet:
Related to #97. I've been to get together a proof of concept of the API described in https://github.com/ReactiveCocoa/ReactiveSwift/issues/97#issuecomment-263199750, but running into subtle issues with the lifetime semantics of composed Properties (which use `replayLazily` internally).

Currently, the start handler goes out of scope the first time it is started, which means no strong reference is kept to the inner producer after starting. In order to allow the replayed producer to be restarted, we need to keep a reference to it (see `startHandler`), but keeping this reference around breaks a couple of tests verifying when a composed property sends a `completed` event.

It looks like `Property.init(unsafeProducer:)` is implicitly relying on the fact that `replayLazily` discards its reference to the inner producer after the first time it's started. I've tried a few different things, but I haven't been able to work out an alternate implementation of the initialiser that doesn't rely on that fact.

`Property` lifetime management still feels pretty hard! I _think_ in part because `SignalProducer` is a struct, I'm finding it difficult to reason about who owns a reference to what, and why the test is failing. I'd really appreciate some help working out what's going on here.

### DIFF
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -1867,7 +1867,7 @@ extension SignalProducerProtocol {
 					// Start the underlying producer if it has never been started.
 					start.swap(nil)?()
 
-					// Terminate the replay loop.
+					// Terminate the observation loop.
 					return
 
 				case let .failure(error):
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -1812,14 +1812,17 @@ extension SignalProducerProtocol {
 	/// This operator is only recommended when you absolutely need to introduce
 	/// a layer of caching in front of another `SignalProducer`.
 	///
-	/// - precondtion: `capacity` must be non-negative integer.
+	/// - precondition: `capacity` must be non-negative integer.
 	///
 	/// - parameters:
-	///   - capcity: Number of values to hold.
+	///   - capacity: Number of values to hold.
+	///   - policy: A `ReplayCachePolicy` that responds to termination events,
+	///             optionally allowing the terminated, replayed producer to be
+	///             restarted.
 	///
 	/// - returns: A caching producer that will hold up to last `capacity`
 	///            values.
-	public func replayLazily(upTo capacity: Int) -> SignalProducer<Value, Error> {
+	public func replayLazily(upTo capacity: Int, policy: ReplayCachePolicy<Error> = .alwaysReplay) -> SignalProducer<Value, Error> {
 		precondition(capacity >= 0, "Invalid capacity: \(capacity)")
 
 		// This will go "out of scope" when the returned `SignalProducer` goes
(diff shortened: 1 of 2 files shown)


sharplet:
I've pushed d7b6686bba9a5f939ee926e3f7e82e7789630344 to show where I'm currently going with the replay cache policy idea. Once CI completes, it should hopefully show that 6cab6a3bde2c92a8fd589828bd765785e03594ab is the minimal change that breaks the composed `Property` lifetime behaviours, and d7b6686bba9a5f939ee926e3f7e82e7789630344 adds support for retrying after failure but without breaking the existing tests for `replayLazily()`. This is still very much a work in progress, as I got fairly set back by the property lifetime stuff and haven't tested the new stuff yet.

Sources/SignalProducer.swift
```diff
@@ -1866,9 +1894,11 @@ extension SignalProducerProtocol {
 					}
 
 					// Start the underlying producer if it has never been started.
-					start.swap(nil)?()
+					if start.swap(false) {
+						startHandler()
+					}
 
-					// Terminate the replay loop.
+					// Terminate the observation loop.
```
andersio: I don't feel the need to rename it. 🤔

Sources/SignalProducer.swift
```diff
@@ -1830,17 +1833,43 @@ extension SignalProducerProtocol {
 		let lifetime = Lifetime(lifetimeToken)
 
 		let state = Atomic(ReplayState<Value, Error>(upTo: capacity))
+		let start = Atomic(true)
 
-		let start: Atomic<(() -> Void)?> = Atomic {
-			// Start the underlying producer.
+		// Start the underlying producer.
+		func startHandler() {
```
andersio: `replayLazily` retains `self` only when it is not ever started. Once it is started, it throws the start handler (which retained `self`) away. This is why there is a `start` of type `Atomic<(() -> Void)?>`. The behaviour — at least for the default policy — should not be changed.

Sources/SignalProducer.swift
```diff
@@ -1830,17 +1833,43 @@ extension SignalProducerProtocol {
...
-				.start { event in
-					let observers: Bag<Signal<Value, Error>.Observer>? = state.modify { state in
-						defer { state.enqueue(event) }
-						return state.observers
-					}
-					observers?.forEach { $0.action(event) }
+				.startWithSignal { events, _ in
+					let events = SignalProducer(events)
+
+					let invalidated = events
```
andersio: It seems all these could be flattened into one event action.

Sources/SignalProducer.swift
andersio: I'd support an eager release of the start handler whenever it is possible though.

Sources/SignalProducer.swift
```diff
@@ -1830,17 +1833,43 @@ extension SignalProducerProtocol {
...
 		let state = Atomic(ReplayState<Value, Error>(upTo: capacity))
+		let start = Atomic(true)
 
-		let start: Atomic<(() -> Void)?> = Atomic {
-			// Start the underlying producer.
+		// Start the underlying producer.
+		func startHandler() {
 			self
+				.materialize()
 				.take(during: lifetime)
```
andersio: The lifetime is what causes the failing test cases. The token tracks all outstanding producers, and breaks the observation only if there is none of them left.

andersio:
The source of the problem is the guarantee of all properties emitting its latest value, even if it has been deinitialized.

In these failing test cases:

1. the flattened property is retained till the end of the scope.
2. As it was made to retain the upstream producer in `startHandler`, this means ultimately the outer property's producer is retained too.
3. The producer retains the `Atomic` storage, which retains the last property being sent.

This also explains why `latest` doesn't fail — the flattened property has deinitialized immediately after use.

I don't think there is any workaround.

andersio:
I am still leaning to an overload constrained to a special error type. The PR in its current form would allow `failure` being silently absorbed by `replayLazily`, regardless of being intentional or not.

andersio:
We might consider instead generalizing this as "autoconnect()" operator like Combine and Rx when there is an interest in expanding primitive operators.

Closing because of inactivity.

PR #206, opened 2016-12-31
[Internal] Manually allocate memory for the atomic counter in UIScheduler.

andersio:
Related: https://lists.swift.org/pipermail/swift-users/Week-of-Mon-20161205/004147.html

> _Joe Groff jgroff at apple.com_
> If you apply them to memory you allocated manually with malloc/free on UnsafeMutablePointer's allocation methods, then yeah, they should work as they do in C. That's the safest way to use these functions today. Passing a Swift `var` inout to one of these functions does not guarantee that accesses to that var will maintain atomicity, since there may be bridging or reabstracting conversions happening under the hood.

### DIFF
--- a/Sources/Scheduler.swift
+++ b/Sources/Scheduler.swift
@@ -97,7 +97,20 @@ public final class UIScheduler: SchedulerProtocol {
 	#if os(Linux)
 	private var queueLength: Atomic<Int32> = Atomic(0)
 	#else
-	private var queueLength: Int32 = 0
+	// `inout` references do not guarantee atomicity. Use `UnsafeMutablePointer`
+	// instead.
+	//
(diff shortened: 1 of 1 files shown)


Sources/Scheduler.swift
```diff
@@ -97,7 +97,16 @@ public final class UIScheduler: SchedulerProtocol {
 	#if os(Linux)
 	private var queueLength: Atomic<Int32> = Atomic(0)
 	#else
-	private var queueLength: Int32 = 0
+	private let queueLength: UnsafeMutablePointer<Int32> = {
```
mdiep: Can you add a comment here explaining why `UnsafeMutablePointer` is used and linking to the thread?

PR #211, opened 2017-01-04
Let's make `Atomic` more simple, efficient and reusable

dmcrodrigues:
`Atomic` is an important part of ReactiveSwift but it's also something additive for any project using `ReactiveSwift` since it's publicly available for instantiation and can be used outside RAC codebases. However, it's only provided in a non-recursive version even when internally there's also a recursive implementation.

The idea behind this PR is provide both a recursive and non-recursive versions of `Atomic` and simplify the current implementation by reducing types, promoting consistency and improve performance.

This is achieved by exposing a strategy about the locking wanted plus a new initializer. I think `didSet` can be useful in certain use cases, internally it's useful for example, and it could be something to publicly expose to provide an extra flexibility or we could left this out.

```swift
public enum AtomicLocking {
    case recursive
    case nonRecursive
}

public class Atomic<Value> {

    public convenience init(_ value: Value) {}

    public init(_ value: Value, locking: AtomicLocking = .nonRecursive, didSet action: ((Value) -> Void)? = nil)
}
```

Having a single `Atomic` with this kind of parameterization avoids having two types for each version: recursive and non-recursive, plus avoiding both a protocol and extension to share implementation between them when the only thing that changes is the lock used. This was originally discussed and introduced in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3079.

This also promotes consistency since `Atomic` is currently using a POSIX's mutex but `RecursiveAtomic` is using a `NSRecursiveLock` instead of a POSIX's mutex configured as recursive. And there's also the question of performance.

#### Test
- Xcode 8.2.1
- Swift 3.0.2
- Compiled with `-O -whole-module-optimization`
- 10 million iterations

[Source Code](https://gist.github.com/dmcrodrigues/e879828a665dd3a723789a7516bf6933)

##### 1. No recursive lock

- `NSRecursiveLock`: 0.662 sec. (1.521% STDEV)
- `pthread_mutex_t`: 0.506 sec. (2.167% STDEV)

##### 2. Lock with recursion = 2

- `NSRecursiveLock`:  1.209 sec. (1.284% STDEV) 
- `pthread_mutex_t`: 0.733 sec. (4.105% STDEV)

##### 3. Lock with recursion = 4

- `NSRecursiveLock`:  1.885 sec. (1.401% STDEV)
- `pthread_mutex_t`: 1.001 sec. (1.572% STDEV)

##### 4. Lock with recursion = 8

- `NSRecursiveLock`:  3.233 sec. (1.800% STDEV)
- `pthread_mutex_t`: 1.437 sec. (1.204% STDEV)

---

Like was expected, a POSIX's mutex configured as recursive shows it as being much more efficient than `NSRecursiveLock` especially when there's some recursion involved. 

I don't know the reasoning behind the adoption of `NSRecursiveLock` for `RecursiveAtomic` but considering results, adopting a POSIX's mutex seems a good option in terms of consistency and performance, unless there's some detail or limitation that I may not be aware.

In case of positive feedback about this change there's some documentation missing that should be added.

### DIFF
--- a/Sources/Atomic.swift
+++ b/Sources/Atomic.swift
@@ -125,12 +125,11 @@ internal struct UnsafeAtomicState<State: RawRepresentable>: AtomicStateProtocol 
 #endif
 }
 
-final class PosixThreadMutex: NSLocking {
-	private var mutex = pthread_mutex_t()
+private final class Lock {
+	private var mutex: pthread_mutex_t
 
-	init() {
-		let result = pthread_mutex_init(&mutex, nil)
-		precondition(result == 0, "Failed to initialize mutex with error \(result).")
+	private init(_ mutex: pthread_mutex_t) {
+		self.mutex = mutex
 	}
 
 	deinit {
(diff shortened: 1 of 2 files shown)


andersio:
Reentrant lock has a fundamentally different semantic, and I am not fond of having both under the same type. Places that actually need reentrant locks are limited either, so this is why it isn't part of the public API.

+1 on pthread_mutex in release build. I think we should use `NSLock` in PosixThreadMutex for debug builds, heh, so that deadlocks would not be a silent halt.

andersio:
As a side note, I am eyeing for [recursive writes in `MutableProperty`](https://github.com/ReactiveCocoa/ReactiveSwift/commit/2c12224f8234f90df57822b82c582dad13e25891), which requires entrance counting.

mdiep:
I'm curious: is there a reason that you are looking to do this? Do you have a case where you need a recursive lock?

andersio:
The involved entities have been significantly changed lately, esp with the introduction of the adaptive `Lock`. `MutableProperty` had also adopted a recursive pthread mutex. If you still feel that a recursive `Atomic` is necessary, feel free to bring it up again.

PR #217, opened 2017-01-10
Improved docs on EventLogger

NachoSoto:
Unfortunately, since Swift dropped named arguments on closures, Xcode provides no clues as to what they are.
All you can see when opening the compiled framework on Xcode is:
```swift
/// A type that represents an event logging function.
public typealias EventLogger = (String, String, String, String, Int) -> Swift.Void
```

So hopefully this will make it a bit easier to use.


mdiep:
🤘

PR #218, opened 2017-01-11
Allow recursive value sending in MutableProperty.

andersio:
This PR adds the ability to set the value of `MutableProperty` recursively. Note that it does not change the `Signal` contract which prohibits recursive value sending. It would only become part of the Property contract.

This means the following snippet is now valid:
```swift
property.producer.startWithValues { value in
    if !fulfilsTerminationCondition {
        property.value = action(value)
    }
}
```

#### Use Case

Conflict resolution in an asynchronous bidirectional binding. For example, when `propertyA` emits a `value` event, the binding observes it and [may subsequently overwrite it synchronously with a new conflict-resolved value based on the merge policy](https://github.com/ReactiveCocoa/ReactiveSwift/blob/58f55b990e89f7e1b3c6803122099fe3050ab212/Documentation/BidirectionalBindings.md#consistency-guarantees).

While it is possible to perform the writeback asynchronously, it could result in windows of visible conflicting values that is hard to workaround.

### DIFF
--- a/Sources/Atomic.swift
+++ b/Sources/Atomic.swift
@@ -193,60 +193,6 @@ public final class Atomic<Value>: AtomicProtocol {
 	}
 }
 
-
-/// An atomic variable which uses a recursive lock.
-internal final class RecursiveAtomic<Value>: AtomicProtocol {
-	private let lock: NSRecursiveLock
-	private var _value: Value
-	private let didSetObserver: ((Value) -> Void)?
-
-	/// Initialize the variable with the given initial value.
-	/// 
-	/// - parameters:
-	///   - value: Initial value for `self`.
-	///   - name: An optional name used to create the recursive lock.
-	///   - action: An optional closure which would be invoked every time the
-	///             value of `self` is mutated.
-	internal init(_ value: Value, name: StaticString? = nil, didSet action: ((Value) -> Void)? = nil) {
-		_value = value
-		lock = NSRecursiveLock()
-		lock.name = name.map(String.init(describing:))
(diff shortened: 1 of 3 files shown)


mdiep:
> While it is possible to perform the writeback asynchronously, it could result in windows of visible conflicting values that is hard to workaround.

I believe that still possible. Since you don't know what order the observers will be notified, an observer could be notified before one that overwrites it. It seems like we might need a different solution for this sort of thing.

andersio:
> I believe that's still possible. Since you don't know what order the observers will be notified, an observer could be notified before one that overwrites it. It seems like we might need a different solution for this sort of thing.

As written in [the draft](https://github.com/ReactiveCocoa/ReactiveSwift/blob/58f55b990e89f7e1b3c6803122099fe3050ab212/Documentation/BidirectionalBindings.md#consistency-guarantees), the order among multiple downstreams is not guaranteed.

What would be guaranteed is that whenever a downstream enqueues a piece of work to the upstream scheduler, e.g. a button tap triggers an `Action`, that reads a two-way bound property, the work always sees a conflict resolved value that is preceded by all the writes of the enqueue. IOW, it can also be preceded by writes from other downstreams, but still the value would be conflict resolved.

This is exactly the guarantee which requires recursive writeback. In an one-to-one case, if the conflict resolved value is written back asynchronously (upstream self-enqueue), we cannot guarantee that the same downstream hasn't enqueued anything to the upstream that precedes the writeback block. IOW, it might see the unsolved value.

PR #221, opened 2017-01-12
Fix infinite feedback loop between Action states

### LINKED ISSUE #220: `Action` behavior: Changing `enabledIf` condition in a observer of `isExecuting.producer` leads to infinite loop
This is how the behavior happen.

```swift
let condition = MutableProperty<Int>(1)
let isValid = MutableProperty<Bool>(false)
isValid <~ condition.map { $0 == 0 }

let action = Action<Void, Void, NoError>(enabledIf: isValid) { _ -> SignalProducer<Void, NoError> in
    return .empty
}

action.isExecuting.producer
    .startWithValues { executing in
        print("executing: ", executing) // infinite printing
        condition.value = 10 // update condition
}
```

In RAC4, this behavior does not happen. (no infinite loop)
When using `isExecuting.signal`, this does not happen.

Is this a bug or am I using these incorrectly?

sharplet:
Fixes #220.

There is a potentially circular relationship between `Action.isEnabled` and `Action.isExecuting`. It's valid for users to observe properites on `Action` and feed the output back into the action state, and in turn influence `Action.isEnabled`.

This resolves an issue where:

1. The user supplies a condition as the action enabled state
2. Any property derived from the internal `Action.state` is observed
3. Synchronously, the observer updates the user-supplied state property
4. Which updates the internal `state` property
5. Which causes an externally-observable event to fire, returning to (3), indefinitely.

The reason for this manifesting as an infinite loop, rather than an infinite recursion, is the internal use of a `replayLazily` in `Property`: each time a side effect on the user-supplied property occurs, that queues up another event inside the replayed signal producer, and the observer is never attached because there is always a buffered event to consume.

The issue was that multiple events were firing synchronously when observing `Action.isExecuting`, even though the executing state had never changed. The fix is to skip repeats on `Action.isEnabled` and `Action.isExecuting`, so that observers are only ever notified if the value actually changes, instead of whenever something triggeres a side-effect on the internal state.

### DIFF
--- a/Sources/Action.swift
+++ b/Sources/Action.swift
@@ -102,8 +102,8 @@ public final class Action<Input, Output, Error: Swift.Error> {
 				}
 			}
 
-		self.isEnabled = state.map { $0.isEnabled }
-		self.isExecuting = state.map { $0.isExecuting }
+		self.isEnabled = state.map { $0.isEnabled }.skipRepeats()
+		self.isExecuting = state.map { $0.isExecuting }.skipRepeats()
 	}
 
 	/// Initializes an action that will be conditionally enabled, and creates a
--- a/Tests/ReactiveSwiftTests/ActionSpec.swift
+++ b/Tests/ReactiveSwiftTests/ActionSpec.swift
@@ -7,7 +7,7 @@
 //
 
 import Foundation
-
+import Dispatch
 import Result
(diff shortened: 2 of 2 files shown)


mdiep:
Thanks for fixing this!

Issue #163, opened 2016-12-18
`Signal` adaptivity.


andersio:
We usually have lots of `Signal`s sticking around. But not all of them is always in use at any point — by "in use", I mean a `Signal` graph as a whole is not observed by anyone. Some can be freed in this circumstance, e.g. `Notification` signals if not transformed, but generally they cannot.

Let's say `reactive.trigger(for: selector)` — it creates a `Signal` that would stick around forever until the object is gone. So it still costs a bit CPU time even if it gets no observers. Convenience `Signal`s, e.g. `Action.values`, `Action.errors`, also fall into this category.

Inspired by [CwlSignal](https://github.com/mattgallagher/CwlSignal)'s notion of activation, we might optimize this by having adaptive `Signal`s.

#### What does adaptivity mean to a `Signal`?
It means a `Signal` would adapt to the situations that the event delivery would not yield an substantial effect, and would deactivate accordingly. Deactivation means **all subsequent `value` events received are silently dropped** until reactivation.

Deactivated `Signal` can be reactivated simply by attaching an observer.

Currently, `value` events would be going through the entire event delivery routine — and all the derived `Signal`s — even if eventually it would reach no one.

#### Limited propagation

**Edit:** While `Signal` adaptivity is designed to cater a `Signal` graph, only stateless `Signal`s should propagate the deactivation. In other words, adaptivity is an opt-in.

For example, stateful operators like `combineLatest` rely on being always informed of the latest value, and therefore should not inform its upstream of its deactivation at all.

#### How would it work?
First, we would have to make a distinction between the types of `Signal` observers:
1. external observers.
2. event emitters of a `Signal`'s descendants.

With that, we would define that a `Signal` should deactivate if it has:

1. no external observer; and
1. its descendants has all declared inactivation

Similarly, we would define that a `Signal` should reactivate if:

1. is deactivated; AND
1. is attached with a external observer; or
1. is informed the reactivation of its descendants.

This requires a reversed channel to the event flow, similarly to producer interruption.

#### Is it breaking?
Apparently yes. But not for the general crowd of ReactiveSwift. It affects the same group of audience as [the new `Signal` lifetime semantic](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2959).

The introduction of `Signal` adaptivity would further restrict the `Signal` semantic:

* If a `Signal` graph does not have any external observer, injected side effects (e.g. `on`) would not be evaluated.

#### How does it affect the public API?
[...]

Issue #144, opened 2016-12-07
Goodbye `interrupted`...?


andersio:
Goodbye, our beloved `interrupted`.

![](http://i.giphy.com/5Hqt5EuWEzSLu.gif)

## TL;DR
By removing, we would have:  

1. **A more compact API**; and
2. **One less `Signal` to be created when starting a `SignalProducer`.**
    Composed properties would benefit significantly - in particular the use of `replayLazily` - in the same way as #140 on `SignalProducer`.

## Rationale
Back in ~~ReactiveSwift -2.0~~ ReactiveCocoa 3.0, `Signal` has a different lifetime semantic. That is, holding the event emitter of a `Signal` would indefinitely retain the `Signal` unless any terminal event is sent. So `interrupted` was introduced to help `SignalProducer` interrupt the leaves of its `Signal` graph (the ultimate upstreams) without hassle.

In ReactiveSwift 1.0, a new `Signal` lifetime semantic was introduced. Let's call it *Automatic Observer and Reference Counting* (AORC), which is pretty predictable in when a `Signal` would dispose of itself:

1. Nobody holds a strong reference to it; and
2. There is no active observer.

Alright. So why does it matter?

Look closer to the conditions, and the **event emitter** is off the list. That means for a `Signal` graph `tigerMom.map(a).map(b)`, the two transformed `Signal`s are not unilaterally forced to live as long as the ultimate upstream of theirs, i.e. `tigerMom`. They can be disposed of, as long as the two AORC conditions are met. At that point, they would also cease their observations to the upstream `Signal`.

Okay. But... how is it relevant to `interrupted`?

We just mentioned that a `Signal` can live on its own, and ceases its observation to an upstream when disposed. Doesn't this sound like what `interrupted` is designed for?

Now, the most important part is that the `Signal` graph created by a `SignalProducer` generally comprises of **only one-observer `Signal` nodes** (except the root and any multitcasting done by a custom operator through `startWithSignal`). That means if the root node is disposed of, the disposal would definitely propagate along the graph to the leaves.

_(Well, unless some of our operators being naughty.)_

It is exactly what `interrupted` does for us, isn't it? Interrupting one level by one level, from the root node to the leaves.

####  Note that the existing `startWithSignal`
[...]

PR #222, opened 2017-01-13
Introduce `SignalComposer` for signal operators and raw signal composition.

andersio:
_Part of the initiative of Signal adaptivity (#163), and refactoring of interrupt propagation (#144)._

_____

This PR proposes a common `Signal` builder interface for raw `Signal` composition — including primitive `Signal` operators — of _any arity_. 

The primary goal is to internalise both the creation of a composed `Signal`, and its observation to the upstream `Signal`(s). This would allow subsequent work to establish a back channel in the `Signal` graph, but without leaking implementation details to the public API.

#### API changes
The proposed API is addictive to ReactiveSwift 1.x. Existing custom operators would not be affected, but is encouraged to adopt the composer.

Subsequent work might cause it to be mandatory for lifted `Signal` operators.

_____
_A quote from #163._

[pitching] Signal back channel

Here is a prototype which constructs a back channel internally, by **requiring** all operators to adopt the ~~new `transform` operator~~ ~~new `SignalBuilder`~~ new `SignalComposer` that internalises the `Signal` creation.

```swift
public func map<U>(_ transform: @escaping (Value) -> U) -> Signal<U, Error> {
	return Signal.composer
		.observe(self) { event, observer in
			observer.action(event.map(transform))
		}
		.make()
}
```

https://github.com/ReactiveCocoa/ReactiveSwift/compare/transform-op?expand=1

~~The pattern can be extended to `SignalProducer` operators.~~

RFC if internalizing dangling details in this way would get thumbs up. Otherwise, the back channel would have to be publicly exposed like interruption propagation/cancel disposables in producer composition.

```swift
public func map<U>(_ transform: @escaping (Value) -> U) -> Signal<U, Error> {
	return Signal { observer, backChannel in
		return self.observe(backChannel) { event in
			observer.action(event.map(transform))
		}
	}
}
```

<hr>

The original idea was to associate a context with the `Observer`s to differentiate anonymous observers and event emitter, so that `Signal.observe` can differentiate them.

But in reality, the event emitter is usually invoked indirectly in an anonymous `Observer`, which means the context would not be caught implicitly.

### DIFF
--- a/Sources/Disposable.swift
+++ b/Sources/Disposable.swift
@@ -19,6 +19,10 @@ public protocol Disposable: class {
 	func dispose()
 }
 
+public protocol NestedDisposable: Disposable {
+	func setInner(_ disposable: Disposable?)
+}
+
 /// Represents the state of a disposable.
 private enum DisposableState: Int32 {
 	/// The disposable is active.
@@ -107,7 +111,7 @@ public final class ActionDisposable: Disposable {
 }
 
 /// A disposable that will dispose of any number of other disposables.
-public final class CompositeDisposable: Disposable {
+public final class CompositeDisposable: NestedDisposable {
 	private let disposables: Atomic<Bag<Disposable>?>
 	private var state: UnsafeAtomicState<DisposableState>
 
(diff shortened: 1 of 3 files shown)


andersio:
_(Prototyping an alternative approach)_

PR #223, opened 2017-01-13
Fix `map` documentation

### LINKED ISSUE #213: Documentation error
In `ReactiveSwift/Sources/Property.swift`, the documentation of `public func map<U>(_ transform: @escaping (Value) -> U) -> Property<U>` says the function returns an `AnyProperty` when it returns a `Property`.

dcaunt:
Fixes #213. This more closely matches the documentation for the other methods.


mdiep:
Thanks. ✨

PR #225, opened 2017-01-13
Context-aware `Observer`.

andersio:
_Part of the initiative of Signal adaptivity (#163), and refactoring of interrupt propagation (#144). Preceded by #222._
_____

This PR proposes to make `Observer` aware of its context internally. Two new variants of `observe` are introduced to enable propagation of the context, especially in the implementation of operators.

In other words, with the PR, all `Signal` compositions should attach the event emitter _directly_ to the upstream `Signal`. Currently, it is done by invoking the event emitter in an anonymous `Observer` — this breaks the implicit context propagation.

`ObserverContext` at this moment is only a placeholder. `ObserverContext.signalInput` would be used by future work to associate a _reversed back channel_ for ~~`Signal` adaptivity (#163) and~~ interrupt propagation (#144).

```swift
extension SignalProtocol {
	/// Attach an observer to `self`, with the given transform applied on every
	/// event being delivered to `observer`.
	///
	/// - parameters:
	///   - observer: The observer to attach to `self`.
	///   - transform: A closure that maps events of `self` to the event type
	///                which `observer` can accept.
	///
	/// - returns: An optional `Disposable` which can be used to stop the
	///            invocation of the callback. Disposing of the Disposable will
	///            have no effect on the Signal itself.
	@discardableResult
	public func observe<U, E: Swift.Error>(
		with observer: Observer<U, E>,
		_ transform: @escaping (Event<Value, Error>) -> Event<U, E>
	) -> Disposable?

	/// Attach an observer to `self`, with the given action invoked upon every
	/// event being delivered.
	///
	/// - parameters:
	///   - observer: The observer to attach to `self`.
	///   - action: A closure that is invoked for every event of `self`.
	///
	/// - returns: An optional `Disposable` which can be used to stop the
	///            invocation of the callback. Disposing of the Disposable will
	///            have no effect on the Signal itself.
	@discardableResult
	public func observe<U, E: Swift.Error>(
		with observer: Observer<U, E>,
		_ action: @escaping (Event<Value, Error>, Observer<U, E>) -> Void
	) -> Disposable?
}
```
While the observer argument in `(Event<Value, Error>, Observer<U, E>) -> Void` is not absolutely necessary, it mimics the signature of `withExtendedLifetime` and similar free functions, and accepts function references that would otherwise need to be reabstracted for capturing `observer`.

#### API changes
The proposed changes alone are addictive in ReactiveSwift 1.x.

If interrupt propagation is to be handled by the proposed _reversed back channel_, it would be at least mandatory for all lifted `Signal` operators, thus being a breaking change due to lifting being publicly available.

### DIFF
--- a/Sources/Flatten.swift
+++ b/Sources/Flatten.swift
@@ -393,7 +393,7 @@ extension SignalProtocol where Value: SignalProducerProtocol, Error == Value.Err
 			}
 		}
 		
-		return observe { event in
+		return self.observe(with: observer) { event, observer in
 			switch event {
 			case let .value(value):
 				state.modify { $0.queue.append(value.producer) }
@@ -522,7 +522,7 @@ extension SignalProtocol where Value: SignalProducerProtocol, Error == Value.Err
 			}
 		}
 
-		return self.observe { event in
+		return self.observe(with: observer) { event, observer in
 			switch event {
 			case let .value(producer):
 				producer.startWithSignal { innerSignal, innerDisposable in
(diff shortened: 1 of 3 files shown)


mdiep:
It's really hard to evaluate this without any code that actually makes use of it. Is it used in one of the related PRs? If so, which should I look at?

andersio:
WIP. Not pushed yet.

Issue #143, opened 2016-12-05
Reduce Disposable's sphere of influence


andersio:
Originally: https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2044

Currently, `Disposable` is used in two key areas in the public API:
1. **Resource collection and cleanup.**
    `Signal` generator, `SignalProducer` start side effect.

1.  **Work cancellation.**
    `SignalProducer` work cancellation, `Signal` observer detaching, `Scheduler` work cancellation.

While there is likely no sensible alternative for work cancellation tokens, the newly introduced `Lifetime` seems a nice contender to the use of `Disposable` for resource management in `Signal` and `SignalProducer`.

For example, we may have instead:
```swift
let producer = SignalProducer { observer, lifetime in
    // Attach a resource to the lifetime of the produced `Signal`.
    lifetime += signal.observe(observer)
}

let signal = Signal { observer, lifetime in
    // Attach a resource to the lifetime of the produced `Signal`.
    lifetime += upstream.observe(observer)
}
```

The key change here is that the "garbage collector" passed to the `SignalProducer` start handler **are no longer disposable**. `CompositeDisposable` as the public interface allows it to be disposed in the start handler, while it is supposed to be disposed only upon external interruption.

#### Edit: API changes
This would break sources that somehow relies on this though. I am not sure about the impact, but it feels like a fringe case.

The `SignalProducer` change would be breaking due to conflicting signature with existing initializer. Even if we replicate the CompositeDisposable API on `Lifetime`, it breaks whenever the second argument type is not inferred but explicitly specified.

~~The `Signal` change, however, is purely addictive to ReactiveSwift 1.0.~~ The `Signal` change would be breaking due to closure return type not being able to be inferred from multi-line body.

<hr>

As an extension, we may also introduce a new variant of `observe`, which associates an observation with the given `Lifetime` (or trigger signal).
```swift
property.signal.observe(during: reactive.lifetime) { event in
    // ...
}
```
The main advantage over `take(until:)` would be avoiding an extra `Signal` just to forward the events when used at the tail of a signal composition, at a cost of expanding the reach of the observation APIs.

mdiep:
I don't think it makes sense to add resource collection to `Lifetime`.

We have 3 jobs:

- Cancellation
- Cleanup
- Lifetime

I think cleanup is more closely related to cancellation than lifetime.

andersio:
[...]

PR #229, opened 2017-01-14
`Lifetime.observeEnded` and `take(during:)` fix.

andersio:
1. New methods and operator: ~~`attach`,~~ `observeEnded` ~~and `+=`~~.

   `ended` provides a full `Signal` interface that is overblown for the general use cases of `Lifetime`.

1. [Bugfix?] Safe observation ~~& deprecating `Lifetime.ended`.~~

   `ended.observeCompleted` does not handle the observations to a terminated `ended` Signal, and the action is simply ignored.

    [It should have been invoked regardless](https://github.com/ReactiveCocoa/ReactiveSwift/pull/229/files#diff-66075186b734efda335ddf4d9c209017R62), like how adding a `Disposable` to disposed `CompositeDisposable` would dispose of the disposable being added.

   ~~Composition of `Lifetime` is suggested to be done through lifetime operators instead, e.g. `and` and `or`, should the demand arise.~~

### DIFF
--- a/Tests/ReactiveSwiftTests/LifetimeSpec.swift
+++ b/Tests/ReactiveSwiftTests/LifetimeSpec.swift
@@ -1,6 +1,6 @@
 import Quick
 import Nimble
-@testable import ReactiveSwift
+import ReactiveSwift
 import Result
 
 final class LifetimeSpec: QuickSpec {
--- a/Tests/ReactiveSwiftTests/LifetimeSpec.swift
+++ b/Tests/ReactiveSwiftTests/LifetimeSpec.swift
@@ -1,6 +1,6 @@
 import Quick
 import Nimble
-import ReactiveSwift
+@testable import ReactiveSwift
 import Result
 
 final class LifetimeSpec: QuickSpec {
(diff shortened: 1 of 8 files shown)


sharplet:
I think I'm 👍  on `attach` and `observeEnded`, although I think `attach` could use a more descriptive name. But I think adding `+=` would be a step backwards in terms of clarity.

Tests/ReactiveSwiftTests/LifetimeSpec.swift
```diff
@@ -1,34 +1,97 @@
...
-import ReactiveSwift
+@testable import ReactiveSwift
 import Result
 
 final class LifetimeSpec: QuickSpec {
 	override func spec() {
 		describe("Lifetime") {
-			it("should complete its lifetime ended signal when the it deinitializes") {
-				let object = MutableReference(TestObject())
+			describe("ended") {
```
sharplet: It's a nitpick, but do you think we can make this change without reindenting the file? It makes the diff a little hard to follow, and there's no shared setup/teardown code.

mdiep:
I really like the simplicity of exposing `ended: Signal<(), NoError>`. Clear, unambiguous, and does very little to expand the API. I also think we should prefer `Signal`s to `Disposable`s.

If we're concerned about the safety issue you mentioned, adding the terminated events to `Observer` seems like a fine approach.

inamiy:
I'm +1 for not exposing `ended: Signal<(), NoError>`.
It's hard to grasp from the type that only `ended.observeCompleted { ... }` will work.

But I don't get why disposable is needed as an argument of `func attach`.
I think just plain `() -> ()` is sufficient.

Also, I don't think `Observer(terminated:)` is necessary because `end` only does `sendCompleted`.
(Is there any cases of `sendInterrupted` or others?)

P.S. I think `Signal<Never, NoError>` describes a bit better than `Signal<(), NoError>`.

andersio:
@inamiy When the lifetime has already terminated.

andersio:
@mdiep #143 might have a stake in this, if it is still on the table.

inamiy:
@andersio 
Ah, you are right. Lifetime may `sendInterrupted`.
And with #143, I got the idea of `func attach`.

mdiep:
I still prefer `ended: Signal<(), NoError>`.

I can be 👍 to `observeEnded` if people find `ended` confusing.

But I'd prefer not to add `attach` or `+=`. I think that muddles the job of `Lifetime`—which should be just to notify observers when the lifetime has ended. I don't think we should give `Lifetime` any knowledge of `Disposable`.

andersio:
I don't quite see the problem here. We can just treat a `Disposable` as a termination observer.

mdiep:
Yeah, so I think writing `lifetime.observeEnded(disposable.dispose)` is fine. I don't think it requires extra sugar that couples the two concepts.

andersio:
The deprecation of `ended` is reverted, while `observeEnded` is documented as the preferred API unless composition is needed.

Tests/ReactiveSwiftTests/LifetimeSpec.swift
```diff
@@ -1,6 +1,6 @@
 import Quick
 import Nimble
-import ReactiveSwift
+@testable import ReactiveSwift
```
sharplet: Does this still need to be `@testable`?

PR #232, opened 2017-01-16
Rex’s filterMap

RuiAAPeres:


Sources/Signal.swift
```diff
@@ -571,6 +571,31 @@ extension SignalProtocol {
...
 	}
+	
+	/// Applies `transform` to values from `signal` and forwards values with non `nil` results unwrapped.
+	/// - parameters:
+	///   - transform: A closure that accepts a value from the `value` event and
+	///                returns a new optional value.
+	///
+	/// - returns: A signal that will send new values, that are non `nil` after the transformation.
+	public func filterMap<U>(_ transform: @escaping (Value) -> U?) -> Signal<U, Error> {
+		return Signal { observer in
```
stephencelis: Seems like it could be useful utility operator! Could we chain `map` and `skipNil` here instead, or am I missing something?

andersio: Less indirection the better.

stephencelis: @andersio Performance-wise?

andersio: Every `Signal` means a couple more calls and locking, though we are gradually slashing a microsecond here and there.

It is not like a complex algorithm that does benefit from reusing implementations anyway...

andersio: Though it is hardly a problem when working with iOS. My obsession has started only because of the super high-speed scrolling in an NSTableView with many visible items.

🎃

stephencelis: Ah, I'd say reusing implementations is generally a nice benefit of FP 😄 Though I understand wanting to minimize locking, etc.

neilpa: It's been a couple years since I've tested it, but we noticed similar lag on larger iOS scroll views with deep signal chains. Hence why I went with the direct implementation here.

Sources/SignalProducer.swift
```diff
@@ -464,6 +464,16 @@ extension SignalProducerProtocol {
 		return lift { $0.filter(predicate) }
 	}
 
+	/// Applies `transform` to values from the producer and forwards values with non `nil` results unwrapped.
+	/// - parameters:
+	///   - transform: A closure that accepts a value from the `value` event and
+	///                returns a new optional value.
+	///
+	/// - returns: A producer that will send new values, that are non `nil` after the transformation.
+	public func filterMap<U>(_ transform: @escaping (Value) -> U?) -> SignalProducer<U, Error> {
```
mdiep: This could really just be called `flatMap`. (The Swift stdlib has this.) But maybe that's undesirable from a type inference perspective?

stephencelis: The `Sequence.flatMap` overload you're referring to (which removes `nil` values after transformation) is a weird one: it doesn't behave as `flatMap` should according to general FP laws/expectations, and it should probably be renamed `compactMap`. While I see creating parity with the standard library appealing, I'd rather keep things a bit more predictable within the functional expectations of FRP.

RuiAAPeres:
If no one opposes this PR, could we merge it?

mdiep:
I was hoping this would generate a little more discussion on the name.

This came from Rex, but I'm not sure the origins of the name and I'd like to get that right. Is there prior art for `filterMap` or for `compactMap` (which was mentioned above)?

RuiAAPeres:
Did a quick search, Rust seems to have it: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map

sharplet:
I tend to agree with @stephencelis (https://github.com/ReactiveCocoa/ReactiveSwift/pull/232#discussion_r96555572). I personally don't mind that the standard library reuses `flatMap` for this purpose, and for that reason I would lean slightly towards following the precedent it establishes. But on the other hand, every overload of `flatMap` that ReactiveSwift provides currently accepts a `FlattenStrategy`, and it's possible that this overload could increase the cognitive burden of understanding `flatMap`. For that reason my vote is for `filterMap`.

mdiep:
* Swift stdlib has `flatMap` (it also has `flatten :: [A?] -> [A]`)
* Rust has `filter_map`
* Ruby has `compactMap` (it also has `compact :: [A?] -> [A]`)
* Haskell has `mapMaybe`
* Elm has `List.filterMap`

I guess `filterMap` is the winner

.travis.yml
```diff
@@ -71,9 +71,7 @@ matrix:
       - pod lib lint ReactiveSwift.podspec
       env:
         - JOB=PODSPEC
-    - os: osx
-      language: generic
-      script:
+    - script:
```
NachoSoto: Can you separate these changes in another commit?

Sources/SignalProducer.swift
NachoSoto: But `flatMap` was never a correct name 😅 I think that would only lead to more confusion + ambiguity.

Tests/ReactiveSwiftTests/SignalSpec.swift
```diff
@@ -513,6 +513,28 @@ class SignalSpec: QuickSpec {
...
+				expect(lastValue).to(beNil())
+
+				observer.send(value: "0")
+				expect(lastValue) == 0
+				
+				observer.send(value: "1")
+				expect(lastValue) == 1
+				
+				observer.send(value: "A")
+				expect(lastValue) == 1
```
NachoSoto: Can you add tests to verify errors and completion?

Sources/SignalProducer.swift
stephencelis: Totally! That's what I was trying to communicate 😅

I wish Apple would rename the confusing standard library version to `compactMap`. Too late for an evolution proposal?

RuiAAPeres:
@NachoSoto 👍

Tests/ReactiveSwiftTests/SignalSpec.swift
```diff
@@ -513,6 +513,70 @@ class SignalSpec: QuickSpec {
...
+				
+				expect(lastValue).to(beNil())
+				
+				observer.send(value: "0")
+				expect(lastValue) == 0
+				
+				observer.sendCompleted()
+				
+				observer.send(value: "1")
+				expect(lastValue) == 0
```
mdiep: Can you test that the signal actually sends a completed event?

Tests/ReactiveSwiftTests/SignalSpec.swift
```diff
@@ -513,6 +513,70 @@ class SignalSpec: QuickSpec {
...
+				
+				expect(lastValue).to(beNil())
+				
+				observer.send(value: "0")
+				expect(lastValue) == 0
+				
+				observer.send(error: .default)
+				
+				observer.send(value: "1")
+				expect(lastValue) == 0
```
mdiep: Can you test that the signal sends the error?

Tests/ReactiveSwiftTests/SignalSpec.swift
```diff
@@ -576,6 +576,30 @@ class SignalSpec: QuickSpec {
...
+				
+				mappedSignal.observeCompleted { completed = true }
+				observer.sendCompleted()
+				
+				expect(completed) == true
+			}
+			
+			it("should send failure") {
+				let (signal, observer) = Signal<String, TestError>.pipe()
+				let mappedSignal: Signal<Int, TestError> = signal.filterMap { Int.init($0) }
```
mdiep: `.init` not needed here and above

mdiep:
Awesome! Sorry for the delay on this one.

Tests/ReactiveSwiftTests/SignalSpec.swift
NachoSoto: Unless you just do `signal.filterMap(Int.init)` to avoid the closure.

Issue #138, opened 2016-11-30
amb (race) as a new FlattenStrategy


inamiy:
Just got inspired from https://github.com/ReactiveCocoa/ReactiveObjC/pull/51, I think we can add `FlattenStrategy.first` where multiple producers race and winner forwards all the events after they get `flatten`ed.

Diagram: http://reactivex.io/documentation/operators/amb.html

I think this can be useful in some scenario (though I have never used it before).

NachoSoto:
Wouldn't that be equivalent to
```swift
nestedSignal
  .take(1)
  .flatten(.Merge) // or anything else
```

mdiep:
No, I think it's the first flattened signal to send an event.

```swift
let (s1, o1) = Signal.pipe()
let (s2 o2) = Signal.pipe()
let (s3, o3) = Signal.pipe()

Signal(values: s1, s2, s3)
    .flatten(.first)
   .observeValues { print($0) }

o3.send(value: 7) // s3 becomes the chosen signal and its values are used
```

inamiy:
One example would be `func timeout` that can be re-implemented like this:

```swift
public func timeout(after interval: TimeInterval, raising error: Error, on scheduler: DateSchedulerProtocol) -> SignalProducer<Value, Error> {
    return SignalProducer(values: [
        self, 
        SignalProducer(error: error).delayError(interval, on: scheduler)
    ]).flatten(.first)
}
// NOTE: `delayError` doesn't exist yet
```

RuiAAPeres:
@inamiy any progress on this? 😊

inamiy:
@RuiAAPeres Not yet, but I will tackle on this soon :)

andersio:
Implemented in #233.

PR #233, opened 2017-01-17
Add `FlattenStrategy.race`

inamiy:
Implemented https://github.com/ReactiveCocoa/ReactiveSwift/issues/138.

Ref: [ReactiveX - Amb operator](http://reactivex.io/documentation/operators/amb.html)

I'm still wondering whether `.first` is really a good naming since there's already a blocking `func first()`.
I personally feel `race` is a better name.

### DIFF
--- a/Sources/Flatten.swift
+++ b/Sources/Flatten.swift
@@ -827,26 +827,28 @@ extension Signal where Value: SignalProducerProtocol, Error == Value.Error {
 
 					innerSignal.observe { [unowned innerSignal] event in
 
-						let isWinningSignal: Bool = state.modify {
-							if $0.active == nil {
-								$0.active = innerSignal
+						let isWinningSignal: Bool = state.modify { state in
+							if state.active == nil {
+								state.active = innerSignal
 							}
-							return $0.active === innerSignal
+							return state.active === innerSignal
 						}
 
 						// Ignore non-winning signals.
 						guard isWinningSignal else { return }
 
 						// Dispose all running innerSignals except winning one.
-						disposableHandle.remove()
-						relayDisposable.dispose()
+						if !relayDisposable.isDisposed {
+							disposableHandle.remove()
+							relayDisposable.dispose()
+						}
 
 						switch event {
 						case .completed:
-							let shouldComplete: Bool = state.modify {
-								$0.active = nil
-								$0.innerSignalComplete = true
-								return $0.outerSignalComplete
+							let shouldComplete: Bool = state.modify { state in
+								state.active = nil
+								state.innerSignalComplete = true
+								return state.outerSignalComplete
 							}
 
 							if shouldComplete {
@@ -863,9 +865,9 @@ extension Signal where Value: SignalProducerProtocol, Error == Value.Error {
 				observer.send(error: error)
 
 			case .completed:
-				let shouldComplete: Bool = state.modify {
-					$0.outerSignalComplete = true
-					return $0.innerSignalComplete
+				let shouldComplete: Bool = state.modify { state in
+					state.outerSignalComplete = true
+					return state.innerSignalComplete
 				}
 
 				if shouldComplete {
--- a/Sources/Flatten.swift
+++ b/Sources/Flatten.swift
@@ -45,6 +45,16 @@ public enum FlattenStrategy: Equatable {
 	/// and the latest producer has completed.
 	case latest
 
+	/// Only the events from the "first input producer to send an event" (winning producer)
+	/// should be considered for the output.
+	/// Any other producers that already started (but not sending an event yet)
+	/// will be disposed.
+	///
+	/// The resulting producer will complete when:
+	/// 1. The producer-of-producers and the first "alive" producer has completed.
+	/// 2. The producer-of-producers has completed without inner producer being "alive".
+	case race
+
 	public static func ==(left: FlattenStrategy, right: FlattenStrategy) -> Bool {
 		switch (left, right) {
 		case (.latest, .latest):
@@ -78,6 +88,9 @@ extension Signal where Value: SignalProducerProtocol, Error == Value.Error {
 
 		case .latest:
 			return self.switchToLatest()
+
+		case .race:
+			return self.race()
 		}
 	}
 }
@@ -117,6 +130,9 @@ extension Signal where Value: SignalProducerProtocol, Error == NoError, Value.Er
 
 		case .latest:
 			return self.switchToLatest()
+
+		case .race:
+			return self.race()
 		}
 	}
 }
@@ -157,6 +173,9 @@ extension SignalProducer where Value: SignalProducerProtocol, Error == Value.Err
 
 		case .latest:
 			return self.switchToLatest()
+
+		case .race:
+			return self.race()
 		}
 	}
 }
@@ -196,6 +215,9 @@ extension SignalProducer where Value: SignalProducerProtocol, Error == NoError, 
 
 		case .latest:
 			return self.switchToLatest()
+
+		case .race:
+			return self.race()
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


RuiAAPeres:
> (I personally feel race is a better name)

`flatMap(.🚗, transformation:)`

andersio:
The context & use site of the strategy & the operator are different enough to be confusing IMO.

RuiAAPeres:
Could it be called simply `.amb`? From a discoverability point of view, it would be easier (searching for `Functional Reactive Programming amb` would yield the right sort of information). For people that already know the `.amb` operator the name would be fine as well.

sharplet:
What does "amb" actually mean? I personally feel it would be better to have a more intention-revealing name, and link back to the equivalent Rx version.

inamiy:
(Build failure seems unrelated)

I also don't know what `amb` actually means, so I'm +1 for either `.first` or `.race`.
`.race` is more descriptive (used in RxPHP?), but `.first` is also good name in contrast to `.latest`.

mdiep:
I'm guessing `amb` is short for _ambivalent_?

Sources/Flatten.swift
```diff
@@ -733,6 +754,124 @@ private struct LatestState<Value, Error: Swift.Error> {
 	var replacingInnerSignal: Bool = false
 }
 
+extension SignalProtocol where Value: SignalProducerProtocol, Error == Value.Error {
+	/// Returns a signal that forwards values from the first signal sent on
+	/// `self`, ignoring values sent on consecutive inner signals.
```
mdiep: Likewise, I don't think this is an accurate description.

Sources/Flatten.swift
```diff
@@ -733,6 +754,124 @@ private struct LatestState<Value, Error: Swift.Error> {
...
+				disposable += signalDisposable
+				disposable += signal.observeFirst(observer, relayDisposable)
+			}
+		}
+	}
+}
+
+private struct FirstState<Value, Error: Swift.Error> {
+	var outerSignalComplete: Bool = false
+	var innerSignalComplete: Bool = true
```
mdiep: It seems a little weird to default this to `true`. It might be better to default to `false` and check for `innerSignalComplete || active == nil` in `shouldComplete`?

Sources/Flatten.swift
```diff
@@ -31,6 +31,15 @@ public enum FlattenStrategy: Equatable {
 	/// The resulting producer will complete only when the producer-of-producers
 	/// and the latest producer has completed.
 	case latest
+
+	/// Only the events from the first input producer should be considered for
```
inamiy: Updated comments in cf6c4eb.

Sources/Flatten.swift
inamiy: Updated comments in cf6c4eb.

Sources/Flatten.swift
inamiy: Thanks for this catch! I could remove this redundant flag now in 819550f.

Sources/Flatten.swift
```diff
@@ -733,6 +755,120 @@ private struct LatestState<Value, Error: Swift.Error> {
...
+					}
+				}
+
+			case let .failed(error):
+				observer.send(error: error)
+
+			case .completed:
+				let shouldComplete: Bool = state.modify {
+					$0.outerSignalComplete = true
+					return $0.active == nil
```
mdiep: What should the behavior be if the outer signal receives a signal, completes, but the inner signal hasn't sent an event?

mdiep:
I think I like _race_.

_first_ conflicts with the instance method, but it's also a little ambiguous about whether it's the first signal to be sent or the first signal to send an event. I suppose you could also get confused with _race_ like that—but it seems a little more likely to direct you to the documentation.

Sources/Flatten.swift
inamiy: I haven't written the test yet, but I think the result should be the same as:
	
`it("should complete when the outer signal completes before sending any signals")`.
https://github.com/ReactiveCocoa/ReactiveSwift/pull/233/files#diff-9cd786f9e52e786771d96cf8c42d3befR1481

New test will be: `it("should complete when the outer signal completes before inner signals send any event")`.

inamiy: (I will also update ambiguous test titles too)

inamiy:
OK, I will rename to `.race`.

inamiy:
In 9630f9f, I had to fix `.completed` behavior because scenario like https://github.com/ReactiveCocoa/ReactiveSwift/issues/138#issuecomment-264353079 will complete immediately and this would be totally useless.

To fix that, I had to restore additional state `var innerSignalComplete: Bool = true`, and this flag basically checks the existence of at least one signal been started (doesn't care about value emission).

### New `.completed` behavior of `SignalProducer.flatten(.race)`

- If outer producer completes before any inner producer is created, complete immediately.
- If outer producer completes after inner producer is created, complete after inner producer is completed.
- If inner producer completes, complete after outer producer is completed.

mdiep:
# This is a breaking change because it adds a `case` to an existing `enum`.

This either needs to wait until we're ready for 2.0 on master or it needs to target a 2.0 branch.

andersio:
The PR should target `master-next` in this case, hmm?

mdiep:
@inamiy Would you mind resolving the conflicts here? I think this is ready to be merged to master for a 2.0 release.

NachoSoto:
Sorry if I'm late commenting on this, but does this really add value? Isn't this equivalent to:
```swift
signal
  .take(first: 1)
  .flatten(anyStrategy)
```

andersio:
No. This takes only one inner signal and flattens it. `race` accepts any arbitrary amount of inner signals, flattens the first one that sends events, and disposes of the rest.

mdiep:
> Sorry if I'm late commenting on this, but does this really add value? Isn't this equivalent to:

Discussion [here](https://github.com/ReactiveCocoa/ReactiveSwift/issues/138#issuecomment-264204773) last time you asked this. 😉

NachoSoto:
OMG lol, that's embarrassing @mdiep 😓

NachoSoto:
Ship it then :D

inamiy:
Rebase done in 76d457e 😉  (but CI alive?)

andersio:
@inamiy Try to do `git commit --amend --no-edit` and `git push -f`. Sometimes it stucks.

andersio:
Reopen to trigger CI.

Sources/Flatten.swift
```diff
@@ -764,6 +786,126 @@ private struct LatestState<Value, Error: Swift.Error> {
...
+								$0.active = innerSignal
+							}
+							return $0.active === innerSignal
+						}
+
+						// Ignore non-winning signals.
+						guard isWinningSignal else { return }
+
+						// Dispose all running innerSignals except winning one.
+						disposableHandle.remove()
```
andersio: Is it possible to make these not run for every event? They just need to be run once after all.

Sources/Flatten.swift
```diff
@@ -764,6 +786,126 @@ private struct LatestState<Value, Error: Swift.Error> {
...
+
+					state.modify {
+						$0.innerSignalComplete = false
+					}
+
+					let disposableHandle = relayDisposable.add(innerDisposable)
+
+					innerSignal.observe { [unowned innerSignal] event in
+
+						let isWinningSignal: Bool = state.modify {
```
andersio: Please name the arguments for multi-line closures.

Sources/Flatten.swift
```diff
@@ -764,6 +786,126 @@ private struct LatestState<Value, Error: Swift.Error> {
...
+						// Ignore non-winning signals.
+						guard isWinningSignal else { return }
+
+						// Dispose all running innerSignals except winning one.
+						disposableHandle.remove()
+						relayDisposable.dispose()
+
+						switch event {
+						case .completed:
+							let shouldComplete: Bool = state.modify {
```
andersio: Ditto.

Sources/Flatten.swift
```diff
@@ -764,6 +786,126 @@ private struct LatestState<Value, Error: Swift.Error> {
...
+							observer.action(event)
+						}
+					}
+				}
+
+			case let .failed(error):
+				observer.send(error: error)
+
+			case .completed:
+				let shouldComplete: Bool = state.modify {
```
andersio: Ditto.

Sources/Flatten.swift
inamiy: Done in 1b83292.

Sources/Flatten.swift
inamiy: Done in 1b83292.

Issue #239, opened 2017-01-25
The "Lazy Lens"


liscio:
I apologize for the length of this one, but I feel that this construct I built is very important to add to ReactiveSwift.

I built this extension for `SignalProducer` a while ago while working on the latest [Capo touch](http://capoapp.com/ios) update, and I feel it is an essential addition to the core library. Here's the code for it:

``` swift
extension SignalProducer {
    /// Initializes this SignalProducer with a closure that supplies its value
    /// during the .start event. This initializer is similar to `init(value:)`, 
    /// but it exists to avoid letting "deceptively simple" accessors from being
    /// evaluated immediately from within the body of a `.flatMap` call, e.g.:
    /// 
    /// ```
    /// let imageProducer = underlyingSignal.flatMap(.latest) {
    ///     return SignalProducer(value: $0.slowImageLoad)
    ///         .start(on: backgroundScheduler)
    /// }
    /// ```
    ///
    /// The above code will be evaluated immediately at the call site, and not
    /// on the `backgroundScheduler` as stated. Instead, this pattern can now
    /// work as expected:
    ///
    /// ```
    /// let imageProducer = underlyingSignal.flatMap(.latest) {
    ///     return SignalProducer(lazyValue: { $0.slowImageLoad })
    ///         .start(on: backgroundScheduler)
    /// }
    /// ```
    init(lazyValue: @escaping () -> Value) {
        self.init { observer, disposable in
            observer.send(value: lazyValue())
            observer.sendCompleted()
        }
    }
}

extension PropertyProtocol {
    func lazyLens<T>(getter: @escaping (Value) -> T) -> SignalProducer<T, NoError> {
        return producer.flatMap(.latest) { model in SignalProducer(lazyValue: { getter(model) }) }
    }

    func lazyLens<T>(on scheduler: SchedulerProtocol, getter: @escaping (Value) -> T) -> SignalProducer<T, NoError> {
        return producer.flatMap(.latest) { model in SignalProducer(lazyValue: { getter(model) }).start(on: scheduler) }
    }
}
```

And here's how I use it in practice:

``` swift
final class AlbumViewModel : ViewModelProtocol {
    let title: SignalProducer<String, NoError>
    let albumArtist: SignalProducer<String, NoError>
    let trackText: SignalProducer<String, NoError>
    let artwork: SignalProducer<UIImage?, NoError>
    let blurredArtwork: SignalProducer<UIImage?, NoError>

[...]

PR #240, opened 2017-01-26
Adding `lazyMap`

liscio:
~~~See #239 for an in-depth discussion of the idea here.~~~

Most of this discussion is out of sync with this PR now, which merely adds `lazyMap` to `Signal` and `SignalProducer`, and a lazy value initializer to `SignalProducer` to support the construct.

The below text is retained merely for historical purposes.

---

This PR Includes `init(lazyValue:)` for `SignalProducer`, `lens` for `PropertyProtocol`, `lens` for `MutableProperty`, and a handful of associated tests to demonstrate that everything works as advertised.

For what it's worth, I think this can go into a 1.1 release if it is accepted. It's completely additive for 1.0 and doesn't change any existing API.

### Anticipated Questions

#### Q: Why is `bindingTarget` in `MutableProperty` and not `MutablePropertyProtocol`?

I used the `modify` function to ensure that the value is safely written back into the property by the setter. I don't think this is a major limitation, generally speaking.

Note: This will probably change once #182 is merged.

#### Q: How is this even useful? Shouldn't our model objects just have a `MutableProperty` for each of its properties?

I don't really think that's a good idea, and that approach has broken down for me very quickly, in practice. Model collections can mean thousands and thousands of objects or structs, and you certainly don't want a `MutableProperty` and its associated machinery replicated M x N times (where M is the # of properties, N the # of model objects/structs).

Instead, your model objects are often some kind of simpler value type that should be easy to test, and cheap to duplicate en masse. The "reactive bits" of your model should only exist at the layers where changes are being made, and only at a "resolution" that makes sense for the manipulation that is afforded by your UI, or another source of changes. (By "resolution", I mean the difference between having a signal of changing `Person` values, versus a signal of changing `firstName` values, versus a signal for every changing character in the `firstName` string.)

#### Q: Why does `Property.lazyMap` return a `SignalProducer` and not a `Property`?

Doing so eliminates the benefits of evaluating lazily. `Property` requires an initial value from its producer when it is created via `lift` (and asserts when that's not the case), and that behavior doesn't really align with the concept here. We specifically don't want to start a `lazyMap`ped producer up prematurely, and we certainly don't want it spitting out values until its scheduler runs.

So for now, it's just syntactic sugar that prevents you from having to call `lazyMap` on the property's underlying producer.


Tests/ReactiveSwiftTests/PropertySpec.swift
```diff
@@ -819,6 +819,64 @@ class PropertySpec: QuickSpec {
...
+
+					var characters: [String] = []
+					theLens.startWithValues { character in
+						characters.append(character)
+					}
+
+					expect(getterEvaluated).to(beFalse())
+					expect(characters).to(beEmpty())
+
+					testScheduler.run()
```
sharplet: I'm planning to respond properly to this, but in the meantime I spotted a _theoretical_ false positive in this test. What do you think about this change? https://github.com/ReactiveCocoa/ReactiveSwift/commit/840bf6a2dbab3ff98f49ab47b69a93c594bd90ae

liscio: The test I added merely intended to demonstrate the following:

* Before the signal is started, the getter has not been evaluated
* When the signal is started, but the getter's queue hasn't scheduled the getter for execution, the getter has still not been evaluated
* Once the getter's queue is scheduled, the setter will have run, and the value will have been delivered

I'm probably missing the false positive here, but I think the synchronous execution of the scheduler is a benefit for the specific (simple!) test case I had in mind. Namely, "will the getter get scheduled at all, rather than executing immediately when the `SignalProducer` is started.."

Now, what you're proposing could help write another test case I wasn't feeling well-equipped to write, which is, "the getter should execute on the specific scheduler/thread that is specified".

Let me know what I might have misinterpreted.

liscio:
I have also just now added the writing side as well in the PR, because it was far too tempting.

Using lenses for reading and writing isn't appropriate for all situations, but it's certainly useful when transacting using value types, and you wish to offer a way to set individual properties on a struct. For example:

``` swift
let exportedImageSize: MutableProperty<CGSize>(.zero)

let width: BindingTarget<CGFloat> = exportedImageSize.lens { return CGSize(width: $1, height: $0.height) }
let height: BindingTarget<CGFloat> = exportedImageSize.lens { return CGSize(width: $0.width, height: $1) }

width <~ widthTextField.reactive.stringValues
height <~ heightTextField.reactive.stringValues
```

Personally, this would un-complicate a few spots in my own code where I achieve something similar to the above by taking two discrete width/height properties that I'm forced to combine into a size elsewhere. Now the `CGSize` value is the true source of truth, rather than a composed view of separate width/height properties.

Tests/ReactiveSwiftTests/PropertySpec.swift
sharplet: > "the getter should execute on the specific scheduler/thread that is specified"

Yeah, that's exactly what I was trying to verify. I agree that in this case having a simpler test is better, but I admin I got caught up in trying to solve the problem of verifying that execution occurs on the expected queue/scheduler. 😛

I believe that code does what it says on the tin (and is thread-safe), so if you think it's valuable to include it feel free to use it as inspiration. Some of those details deserve to be factored out into some kind of test helper anyway, so that the test itself can be higher level.

Sources/Property.swift
```diff
@@ -678,3 +759,73 @@ public final class MutableProperty<Value>: MutablePropertyProtocol {
 		observer.sendCompleted()
 	}
 }
+
+extension MutableProperty {
```
andersio: Perhaps this can wait until `ComposableMutablePropertyProtocol` in #182 lands. hmm?

liscio: It could, yes. But it's not really holding up the behavior right now.

If #182 goes in first, I'll definitely merge & apply to that. If #182 continues to take a while, I'm OK with a separate PR that moves this to that protocol.

Sources/Property.swift
```diff
@@ -50,6 +50,87 @@ extension MutablePropertyProtocol {
...
+	/// is modified.
+	///
+	/// - parameters:
+	///	  - getter: The closure used to obtain the returned value from this
+	///             property's underlying value.
+	///
+	/// - returns: A signal producer that returns values obtained using `getter`
+	///            each time this property's value changes.
+	public func lens<U>(getter: @escaping (Value) -> U) -> SignalProducer<U, NoError> {
+		return producer.flatMap(.latest) { model in SignalProducer(lazyValue: { getter(model) }) }
```
mdiep: How does this behave differently than:

```
return producer.flatMap(.latest) { model in SignalProducer(value: getter(model)) }
```

?

I don't see how `lazyValue:` makes any difference here. The returned producer will always be started immediately, so these should be equivalent.

Sources/SignalProducer.swift
```diff
@@ -65,6 +65,35 @@ public struct SignalProducer<Value, Error: Swift.Error> {
...
+	/// on the `backgroundScheduler` as stated. Instead, this pattern can now
+	/// work as expected:
+	///
+	/// ```
+	/// let imageProducer = underlyingSignal.flatMap(.latest) {
+	///     return SignalProducer(lazyValue: { $0.slowImageLoad })
+	///         .start(on: backgroundScheduler)
+	/// }
+	/// ```
+	public init(lazyValue: @escaping () -> Value) {
```
mdiep: This could probably be:

```
public init(_ block: @escaping () -> Value)
```

You can think of it as a value preserving type conversion.

I'm not sure whether that'd cause problems in practice. I think it's unlikely to cause conflicts.

Sources/Property.swift
liscio: @mdiep In the non-scheduled start variation, I don't think there's a difference as you've found.

Sources/SignalProducer.swift
liscio: It would certainly cut down on verbosity. I'll see how it goes.

liscio: The only potential conflict would be if the `init(value:)` variant went away, because then there'd be confusion between `() -> Value` and `Value` itself, forcing closures to always get evaluated (if the compiler wasn't going to balk at the similarity.)

Sources/Property.swift
```diff
@@ -50,6 +50,87 @@ extension MutablePropertyProtocol {
...
+	///             property's underlying value.
+	///
+	/// - returns: A signal producer that returns values obtained using `getter`
+	///            each time this property's value changes.
+	public func lens<U>(on scheduler: SchedulerProtocol, getter: @escaping (Value) -> U) -> SignalProducer<U, NoError> {
+		return producer.flatMap(.latest) { model in
+			return SignalProducer(lazyValue: { getter(model) })
+				.start(on: scheduler)
+		}
+	}
```
mdiep: This would should be the same as this?

```
return producer
    .observe(on: scheduler)
    .flatMap(.latest) { model in SignalProducer(value: getter(model) }
```

Sources/SignalProducer.swift
mdiep: I don't think we'd ever remove the label from `init(value: )`. So I kinda like the idea of a label-less initializer. (We could also add a variant that uses a completion handler, which would be handy.)

Sources/Property.swift
liscio: @mdiep I don't think that works in *quite* the same way, but does it make much of a difference?

According to the docs, what you've pasted above would deliver all the events from the inner signal on the specified scheduler. The `start(on:)` variant would only evaluate the getter on the scheduler. Because there's only one event getting sent, I'm not sure it really matters all that much.

liscio: @mdiep I don't think that works in *quite* the same way, but it passes the tests both ways.

According to the docs, what you've pasted above would deliver all the events from the inner signal on the specified scheduler. The `start(on:)` variant would only evaluate the getter on the scheduler. Because there's only one event getting sent, I'm not sure it really matters all that much. Thoughts?

mdiep: The events are delivered to the `flatMap` on `scheduler`.

They do behave differently though… because cancellation happens on a different scheduler. Your way is probably better.

mdiep:
_Waiting on @liscio's attempt at `lazy` or `lazyMap` as discussed in #239._

liscio:
@mdiep I've got them set up locally but things got a little strange when I was trying to build out some tests and things weren't working out as expected. Hopefully it won't be much longer 'til I get my confusion worked out.

mdiep:
No worries! I just wanted to leave a note (1) so that I remembered the status of this PR and (2) to make sure we were one the same page. ☺️

liscio:
I added to the anticipated questions above, and made a number of changes to the code in response to feedback.

I also added some additional tests, and cleaned up the ones that were there.

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
```diff
@@ -54,6 +54,265 @@ class SignalProducerLiftingSpec: QuickSpec {
...
+					expect(getterCounter) == 2
+
+					// ...but the destination still won't receive the value
+					expect(destination) == []
+
+					// Finally, pushing the target scheduler along will
+					// propagate only the first and last values
+					targetScheduler.advance()
+					expect(getterCounter) == 2
+					expect(destination) == ["🎃", "👻"]
```
liscio: @andersio this test result that was a little unexpected. I had hoped that advancing the scheduler would only have sent the final "ghost" value, but it's clear now that the binding target's queue has to batch up its incoming `setter` invocations.

A `.flatMap(.latest)` equivalent on the setter side (i.e. switching to the newest scheduled `setter` call) might be worth exploring since it only ever makes sense to accept the last-set value. Effectively, each `value` event arriving on the bound signal would have to wipe out all the scheduled setter blocks on its queue and replace with the latest one.

(Yes, I know—easier said than done.)

andersio: Well, it shouldn't be hard. A `SerialDisposable` would do the job.

mdiep: Typically I've done this by doing the equivalent of `throttle(0, on: UIScheduler())`.

Sources/Property.swift
```diff
@@ -50,6 +50,39 @@ extension MutablePropertyProtocol {
...
+	/// Returns a `SignalProducer` that sends a new value each time this
+	/// property's value changes, using the supplied `transform` to supply the
+	/// value being sent.
+	///
+	/// Consider the following example:
+	///
+	/// ```
+	/// let personProperty: MutableProperty<Person>
+	/// let nameProducer: SignalProducer<String, NoError>
+	/// nameProducer = personProperty.lazyLens(on: scheduler) { $0.name }
```
mdiep: s/lazyLens/lazyMap/

Sources/Property.swift
```diff
@@ -50,6 +50,39 @@ extension MutablePropertyProtocol {
 	}
 }
 
+extension PropertyProtocol {
+	/// Returns a `SignalProducer` that sends a new value each time this
+	/// property's value changes, using the supplied `transform` to supply the
+	/// value being sent.
```
mdiep: This reads like documentation for `map`, so it could use some cleanup.

Of note:

1. `that sends a new value each time this property's value changes` isn't true. The point of the operator is that it doesn't guarantee a 1-to-1 mapping of input to output.

2. It doesn't communicate what makes this lazy.

It looks like the other instances of `lazyMap` have better documentation, so you may just need to copy that here.

Sources/Property.swift
```diff
@@ -678,3 +711,71 @@ public final class MutableProperty<Value>: MutablePropertyProtocol {
...
+
+	/// Returns a `BindingTarget` that invokes the given setter to generate a
+	/// new value for this property.
+	///
+	/// Consider the following example:
+	///
+	/// ```
+	/// let personProperty = MutableProperty(Person(firstName: "Steve", lastName: "McQueen"))
+	/// let firstNameBinding = personProperty.bindingTarget {
+	///         return Person(firstName: $1, lastName: $0.lastName)
```
mdiep: Can you name these arguments? I think for documentation purposes that'll be a lot clearer.

Sources/Property.swift
```diff
@@ -678,3 +711,71 @@ public final class MutableProperty<Value>: MutablePropertyProtocol {
...
+	/// ```
+	///
+	/// - note: The lifetime of this `BindingTarget` is tied to the underlying
+	///         `MutableProperty` instance.
+	///
+	/// - parameters:
+	///	  - setter: A closure that takes a value of type `U` to produce a new
+	///             value for this property
+	///
+	public func bindingTarget<U>(setter: @escaping (Value, U) -> Value) -> BindingTarget<U> {
```
mdiep: This is pretty interesting, but it seems like we should consider this in a separate PR. I don't see any relation or dependency between the two? (Other than these were both originally part of your lens concept.)

Sources/Signal.swift
```diff
@@ -549,6 +549,25 @@ extension SignalProtocol {
...
 	}
 
+	/// Maps each value in the signal to a new value, lazily evaluating the
+	/// supplied transformation on the specified scheduler.
+	///
+	/// - note: The difference between `lazyMap` and `map` is that `lazyMap`
+	///         will not force the transformation to get invoked immediately
+	///			when a new value is sent. Furthermore, subsequent `.value`
+	///			events will not cause `transform` to repeatedly evaluate when
+	///			it has not been scheduled for execution by `scheduler`.
```
mdiep: Indentation here is a bit wonky

Sources/Signal.swift
```diff
@@ -549,6 +549,25 @@ extension SignalProtocol {
...
+	///
```
mdiep: Maybe add an _important_ block that points out that there's not a 1-to-1 mapping between inputs and outputs?

Sources/Signal.swift
```diff
@@ -549,6 +549,25 @@ extension SignalProtocol {
...
+	/// - parameters:
+	///	  - transform: The closure used to obtain the returned value from this
+	///                producer's underlying value.
+	///
+	/// - returns: A signal that sends values obtained using `transform` as this 
+	///            signal sends values.
+	public func lazyMap<U>(on scheduler: SchedulerProtocol, transform: @escaping (Value) -> U) -> Signal<U, Error> {
+		return flatMap(.latest) { model in SignalProducer({ transform(model) }).start(on: scheduler) }
```
mdiep: Maybe rename `model` to `value`?

This line is a little dense, so I'd also break it up onto multiple lines.

mdiep:
I like this direction a lot. ✨ 

Are you happy with it @liscio?

liscio:
@mdiep Yeah, I think that things are shaping up fairly well with this. I'll have to modify some of my own usage patterns to figure out how to mix and match these ideas nicely, because a "do-everything" `lens` method may not be possible.

I really did want to wrap up the `ViewModel` pattern in a way that was more uniform, where we craft the `lens`es we want in the view model without much fussing about the constructs used underneath to achieve the pattern. But with Swift 3.1 that need will completely go away!

Why? Because now we can do some really interesting things that get us closer to a clean pattern exposing our model object's properties in the viewModel. For example:

``` swift
extension Reactive where Base: MutableProperty<Album?>  {
    var titleValues: SignalProducer<String, NoError> {
        return base.producer.map { $0?.title ?? "" }
    }

    var albumArtistValues: SignalProducer<String, NoError> {
        return base.producer.map { $0?.albumArtist ?? "" }
    }
}
```

So now the need to repeat ourselves when "lensing" our model objects goes away in the view model. The "lens" can be defined once on a specialized extension of `MutableProperty` and then binding targets can be created in a similar way.

I guess this is a really long way of saying that I'm OK with dropping the `lens` as a function in favour of exposing effective tools for implementing `lens` as a pattern.

liscio:
@mdiep I addressed your comments, and in addition to pulling `bindingTarget` (to be opened in a separate PR) I have also pulled `lazyMap` from `PropertyProtocol` because it felt incongruent to the rest of the `Property` API that returned `Property` instances.

mdiep:
The tests don't compile on Linux. So this will need a little love.

liscio:
Oops. I fixed that Linux build issue earlier, but pushed the change over to the bindingTarget branch. Should be all good once this build completes.

Sources/SignalProducer.swift
```diff
@@ -65,6 +65,23 @@ public struct SignalProducer<Value, Error: Swift.Error> {
...
+	/// completes.
+	///
+	/// This initializer differs from `init(value:)` in that it will not 
+	/// evaluate the closure supplying its sole `value` event until the 
+	/// `SignalProducer` is started.
+	///
+	/// - parameters:
+	///   - block: A block that supplies a value to be sent by the `Signal` in
+	///            a `value` event.
+	public init(_ block: @escaping () -> Value) {
```
NachoSoto: Any reason not make this an `SignalProducer(value: )` overload that uses `@autoclosure` instead?

mdiep: How would that differentiate from the existing non-`@autoclosure` version?

You wouldn't want to always have a closure—that creates the same problem, just in reverse.

liscio: @NachoSoto In a nutshell, `@autoclosure` implies `@noescape`, and hence we lose out on the ability to defer execution.

NachoSoto: No, you can have an `@autoclosure @escaping` parameter. I guess it would have to have a different parameter name, you're right, but I think I would make the intention more clear.

liscio: @NachoSoto Interesting. I wonder, though, whether we'd still have an issue where the compiler wants to forcefully optimize a passed-in value as being evaluated immediately.

For example, consider the case of `SignalProducer(value: callAFunction())` versus the `SignalProducer({ callAFunction() })` construct. Would the former be turned into an immediately evaluated variant if the braces are omitted, versus being evaluated lazily? 

That seems a little subtle/delicate from a usage standpoint… But yeah—having the `lazyValue:` as I originally built it might be a good tiebreaker?

mdiep: I think the unnamed parameter best fits with the other unnamed lossless-type-converting functions that we already have.

I'd also really like to add this variant:

```
public init<Value, Error>(_ block: @escaping ((Result<Value, Error>) -> Void) -> Void)
```

PR #241, opened 2017-01-31
Add bindingTarget to MutableProperty

liscio:
This PR adds a helpful `bindingTarget` function to `MutableProperty` that allows for the easy creation of "lenses" on a property that contains a simple value type.

For example, you could use it like this:

``` swift
final class SomeViewModel {
  let size = MutableProperty<CGSize>(CGSize(width: 720, height: 480))
  
  let width: BindingTarget<CGFloat>
  let height: BindingTarget<CGFloat>

  init() {
    width = size.bindingTarget { newWidth, size in
      return CGSize(width: newWidth, height: size.height)
    }
    height = size.bindingTarget { newHeight, size in
      return CGSize(width: size.width, height: newHeight)
    }
  }
}
```

Then, your view controller could simply bind a control's values directly into the width/height binding targets, and the `size` property is the single source of truth.

The benefit of exposing `bindingTarget` is that the `modify` operation is used internally to ensure that the value is atomically set on the underlying property, and the returned `BindingTarget` instance follows the `lifetime` of its parent `MutableProperty`.

### DIFF
--- a/Sources/Property.swift
+++ b/Sources/Property.swift
@@ -678,3 +678,71 @@ public final class MutableProperty<Value>: MutablePropertyProtocol {
 		observer.sendCompleted()
 	}
 }
+
+extension MutableProperty {
+
+	/// Returns a `BindingTarget` that invokes the given setter to generate a
+	/// new value for this property.
+	///
+	/// Consider the following example:
+	///
+	/// ```
+	/// let personProperty = MutableProperty(Person(firstName: "Steve", lastName: "McQueen"))
+	/// let firstNameBinding = personProperty.bindingTarget { newFirstName, person in
+	///         return Person(firstName: newFirstName, lastName: person.lastName)
+	///     }
+	///
+	/// // Sets only the firstName value
+	/// firstNameBinding <~ SignalProducer(value: "Lightning")
+	///
+	/// // The personProperty now has a new Person value
+	/// ```
+	///
+	/// - note: The lifetime of this `BindingTarget` is tied to the underlying
+	///         `MutableProperty` instance.
+	///
+	/// - parameters:
+	///   - setter: A closure that takes a value of type `U` to produce a new
+	///             value for this property
+	///
+	public func bindingTarget<U>(setter: @escaping (Value, U) -> Value) -> BindingTarget<U> {
+		return BindingTarget(lifetime: self.lifetime, setter: { [weak self] (inputValue: U) in
+			self?.modify({ (theValue: inout Value) -> Void in
+				theValue = setter(theValue, inputValue)
+			})
+		})
+	}
+
+	/// Returns a `BindingTarget` that invokes the given setter to generate a
+	/// new value for this property.
+	///
+	/// Consider the following example:
+	///
+	/// ```
+	/// let personProperty = MutableProperty(Person(firstName: "Steve", lastName: "McQueen"))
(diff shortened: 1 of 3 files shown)


andersio:
Would it being a `BindingTarget` initializer be better for the task?

mdiep:
I kinda wonder if this could be a `map` method on `BindingTargetProtocol`. 🤔

liscio:
@mdiep That's what I was originally thinking when I was first driven to write this. It just might be the right way to solve the problem.

Even cooler would be a `.flatMap`-like operator. *What the hell even is that*, you might ask?

``` swift
// Imagine Person is a class that exposes a BindingTarget of its own
let changingPerson = MutableProperty<Person?>(/* ... */)

let target = changingPerson.flatMap { return $0?.reactive.firstName ?? .null }
target <~ firstNameTextField.reactive.text
```

In the above example, consider a user's selection in a master/detail interface where they can choose a `Person` from a list. The selection change effectively causes the target of the binding to change.

I can't even begin to imagine how it'd work under the hood, though.

mdiep:
Want to give this a try as `map`? The test suite (and your code) should provide a decent test of whether it's a problem inference-wise.

andersio:
I think the problem is that `BindingTargetProtocol` is write-only, and you do not get the chance to atomically edit the content. If you buff the protocol up, then it would become effectively `PropertyProtocol`.

inamiy:
I think `MutableProperty` should not conform to `BindingTargetProtocol` anymore because of new `bindTarget(setter:)` in this PR.

In addition, my rough local test proved that we don't need to use `if let target = target as? BindingTarget<Target.Value>` (which I found awkward) in `func <~`, i.e.

```diff
diff --git Sources/UnidirectionalBinding.swift Sources/UnidirectionalBinding.swift
index 8cc005cd..c43fcff6 100644
--- Sources/UnidirectionalBinding.swift
+++ Sources/UnidirectionalBinding.swift
@@ -91,19 +91,26 @@ public func <~
 	(target: Target, source: Source) -> Disposable?
 	where Source.Value == Target.Value, Source.Error == NoError
 {
 	// Alter the semantics of `BindingTarget` to not require it to be retained.
 	// This is done here--and not in a separate function--so that all variants
 	// of `<~` can get this behavior.
 	let observer: Observer<Target.Value, NoError>
-	if let target = target as? BindingTarget<Target.Value> {
-		observer = Observer(value: { [setter = target.setter] in setter($0) })
-	} else {
-		observer = Observer(value: { [weak target] in target?.consume($0) })
-	}
+	observer = Observer(value: { target.consume($0) }) // NOTE: strong capture!
 
 	return source.observe(observer, during: target.lifetime)
 }
 ```

Above code will require users to call `property.bindingTarget(setter: { $0.1 }) <~ producer` instead of `property <~ producer` though.
(Helper methods may solve this issue)

andersio:
Well, if we are committed to `BindingTarget`, we may alter `BindingTargetProtocol` to work with just a `BindingTarget` (and perhaps drop the class bound).

andersio:
This can be replaced by a more generic solution based on the [Smart Key Path](https://github.com/apple/swift-evolution/blob/master/proposals/0161-key-paths.md), perhaps.

andersio:
Closing due to inactivity, but we should definitely complete the suite with MutableProperty lenses. 😣

PR #243, opened 2017-02-02
Correct example with Property and Signal

ajjnix:
Missing `signal` from `currentTime: Property<TimeInterval>` before `observe`.


andersio:
Thanks for catching this.

PR #244, opened 2017-02-03
Try to clarify throttle/debounce

mdiep:
This came up in Slack today. I think this makes it a little clearer.

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -1678,21 +1678,24 @@ extension SignalProtocol {
 		}
 	}
 	
-	/// Don't send a value until at least `interval` seconds have passed since
-	/// *the returned signal* last sent a value, then forward the latest value
-	/// on `scheduler`.
+	/// Forward the latest value on `scheduler` after at least `interval`
+	/// seconds have passed since *the returned signal* last sent a value.
 	///
-	/// If the receiver sent values constantly, then the returned signal would
-	/// forward values every `interval` seconds.
+	/// If `self` always sends values more frequently than `interval` seconds,
+	/// then the returned signal will send a value every `interval` seconds.
+	///
+	/// To measure from when `self` last sent a value, see `debounce`.
+	///
+	/// - seealso: `debounce`
 	///
 	/// - note: If multiple values are received before the interval has elapsed,
 	///         the latest value is the one that will be passed on.
 	///
-	/// - note: If the input signal terminates while a value is being throttled,
-	///         that value will be discarded and the returned signal will 
-	///         terminate immediately.
+	/// - note: If `self` terminates while a value is being throttled, that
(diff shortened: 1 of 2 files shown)


Sources/SignalProducer.swift
```diff
@@ -1097,9 +1103,15 @@ extension SignalProducerProtocol {
 		return lift { $0.throttle(while: shouldThrottle, on: scheduler) }
 	}
 
-	/// Debounce values sent by the receiver, such that at least `interval`
-	/// seconds pass after the receiver has last sent a value, then
-	/// forward the latest value on the given scheduler.
+	/// Don't send a value until at least `interval` seconds have passed since
+    /// *the receiver* last sent a value, then forward the latest value on
```
andersio: Indentation is off here.

Sources/Signal.swift
```diff
@@ -1677,9 +1677,13 @@ extension SignalProtocol {
...
 	}
-
-	/// Throttle values sent by the receiver, so that at least `interval`
-	/// seconds pass between each, then forwards them on the given scheduler.
+	
+	/// Don't send a value until at least `interval` seconds have passed since
+	/// *the returned signal* last sent a value, then forward the latest value
+	/// on `scheduler`.
+	///
+	/// If the receiver sent values constantly, then the returned signal would
```
liscio: This is a tricky thing in general, but "the receiver" is difficult to use as a noun in concert with "sending and receiving" values.

"The receiver" is Objective-C terminology where an object "receives" a message, but now we're "calling functions on" objects. Had Swift instead used `mesg` rather than `func` then I'd be more OK with this old terminology. But I digress!

You might be able to safely use this wording instead:

```
/// Returns a signal that sends values exactly, or less frequently than, `interval` seconds 
/// apart, forwarding values on the given `scheduler`.
///
/// That is, the returned signal does not send a value until at least `interval` seconds have
/// passed since it last sent a value.
```

Sources/Signal.swift
```diff
@@ -1847,12 +1851,18 @@ extension SignalProtocol {
 			return disposable
 		}
 	}
-
-	/// Debounce values sent by the receiver, such that at least `interval`
-	/// seconds pass after the receiver has last sent a value, then forward the
-	/// latest value on the given scheduler.
+	
+	/// Don't send a value until at least `interval` seconds have passed since
```
liscio: Similar change to above:

```
/// Returns a signal that sends values only when `interval` seconds have elapsed 
/// since this signal has last sent a value.
///
/// If this signal always sends values more frequently than `interval` seconds, then the 
/// returned signal will never send any values. 
///
/// - seealso: `throttle`
```

If we ever get the Jazzy stuff online in html form, then we can specify a URL with the `seealso`.

PR #245, opened 2017-02-04
Remove `ActionProtocol`.

andersio:
Waiting for Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ branches:
     - master
     # Credit: @Omnikron13, https://github.com/mojombo/semver/issues/32
     - /^(\d+\.\d+\.\d+)(?:-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?(?:\+([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?$/
+    - master-next
 before_script:
   - git submodule update --init --recursive
 script:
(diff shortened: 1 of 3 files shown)


andersio:
It is part of the [ReactiveSwift 2.0 initiative of removing unnecessary protocols](https://github.com/ReactiveCocoa/ReactiveSwift#reactiveswift-20) that ultimately serves as _only_ a workaround to the lack of concrete same-type requirements. Since:

1. the major version would be bumped;
2. we cannot deprecate it in 1.x; and
3. this would not hit the majority of users;

I don't see a need of deprecation.

sharplet:
> we cannot deprecate it in 1.x

I don't believe this is true. I haven't experimented with it yet, but it seems to me it should be possible. I'll get back to you.

> this would not hit the majority of users, and even for those who define custom operators, it is just a matter of renaming

If we include `SignalProtocol` and `SignalProducerProtocol` the likelihood of affecting more users seems higher. It's also possible that users might be writing generic functions that use these protocols as constraints. Granted, the percentage is still probably small.

Still, I remain convinced that it's reasonable to provide deprecation warnings before removing an API. Aggressively breaking public API without first deprecating is a trend that I'd like to see the Swift community move away from.

mdiep:
1.x should be Swift 3.0 and Swift 3.1 compliant. So it seems like we _should_ be able to deprecate these things in a 1.x release.

sharplet:
I opened https://github.com/ReactiveCocoa/ReactiveSwift/pull/248 as a concrete example for discussion.

PR #246, opened 2017-02-04
Rename a parameter of a `SignalProducer` initializer.

andersio:
`block` is an Objective-C term.

### DIFF
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -68,16 +68,16 @@ public struct SignalProducer<Value, Error: Swift.Error> {
 	/// Creates a producer for a `Signal` that immediately sends one value, then
 	/// completes.
 	///
-	/// This initializer differs from `init(value:)` in that it will not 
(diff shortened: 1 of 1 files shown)


Sources/SignalProducer.swift
```diff
@@ -68,14 +68,14 @@ public struct SignalProducer<Value, Error: Swift.Error> {
...
 	///
 	/// - parameters:
-	///   - block: A block that supplies a value to be sent by the `Signal` in
-	///            a `value` event.
-	public init(_ block: @escaping () -> Value) {
+	///   - action: A action that yields a value to be sent by the `Signal` as
+	///             a `value` event.
+	public init(_ action: @escaping () -> Value) {
 		self.init { observer, disposable in
 			observer.send(value: block())
```
liscio: You should probably call `action()` here, right? ;)

andersio: Heh. Corrected it but forgot to push.

PR #248, opened 2017-02-05
Pitch: Deprecate ActionProtocol

sharplet:
...instead of removing it completely. See the discussion on #245 for context. I'm proposing this change as an alternative for discussion: there are some issues with this approach, so I thought it would be good to have something concrete to discuss.

Benefits:

- Mergeable into master now, suitable for inclusion in the next minor release.
- `ActionProtocol` is an implementation detail, but isn't clearly documented as such. There are possibly users making use of it in ways we didn't intend (I was once one of them).

**Update:** See https://github.com/ReactiveCocoa/ReactiveSwift/pull/248#issuecomment-277496797.

<strike>
Downsides:

- Because conditionally compiled code has to be syntactically valid, we essentially have to duplicate the extension. I experimented with using an internal typealias to work around this limitation, but it was unfortunately a dead-end.

I don't have much of a problem with the amount of duplication in this case. But it's pretty clear to me that this approach won't scale all that well to `SignalProtocol` and `SignalProducerProtocol`, whose extensions currently house all of the operator definitions. There would be a _lot_ of duplication. There are certainly techniques we could use to avoid duplicating the implementations (reducing the duplication to declarations only), but they might not be pretty.
</strike>

### DIFF
--- a/Sources/Action.swift
+++ b/Sources/Action.swift
@@ -207,6 +207,7 @@ private struct ActionState {
 }
 
 /// A protocol used to constraint `Action` initializers.
+@available(swift, deprecated: 3.1, message: "This protocol is no longer necessary and will be removed in a future version of ReactiveSwift. Use Action directly instead.")
 public protocol ActionProtocol: BindingTargetProtocol {
 	/// The type of argument to apply the action to.
 	associatedtype Input
--- a/Sources/Atomic.swift
+++ b/Sources/Atomic.swift
@@ -33,7 +33,7 @@ internal protocol AtomicStateProtocol {
 internal struct UnsafeAtomicState<State: RawRepresentable>: AtomicStateProtocol where State.RawValue == Int32 {
 	internal typealias Transition = (expected: State, next: State)
 #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
-	private let value: UnsafeMutablePointer<Int32>
+	internal let value: UnsafeMutablePointer<Int32>
 
 	/// Create a finite state machine with the specified initial state.
 	///


sharplet:
Although I just realised something: given that 3.1 is supposed to be source-compatible with 3.0, I think we should be able to just slap the deprecation warning on and not duplicate anything?? I'll take a look at how to get Swift 3.1 to compile in 3.0 mode...

sharplet:
I've just pushed a change which removes the duplication.

sharplet:
I now believe this should work just fine for our other `*Protocol`s, too. I'll do some more experimenting tomorrow.

sharplet:
> ... suitable for inclusion in the next minor release

Because the deprecation warning only shows for Swift >= 3.1, this could now go into a patch release.

mdiep:
Not sure what to do about the podspec error. I guess we should `--allow-warnings`?

sharplet:
Interesting, I didn't see that warning in my tests. I'll try with a different Xcode version and see if it comes up (I was using the 8.3 beta).

sharplet:
`@available(swift, ...)` is only, ahem, _available_, in Swift 3.1: https://github.com/apple/swift-evolution/blob/master/proposals/0141-available-by-swift-version.md. I'll have to nut out how to handle this properly under semver. Though if this is just going to emit a warning when compiling ReactiveSwift and not in user code, it might be fine to ignore it.

andersio:
This is not source compatible since the definition has changed. Even if it is kept intact, the real problem is that we cannot do this with `SignalProtocol` and `SignalProducerProtocol`. Once we transfer the default implementations to the concrete type, existing external extensions to these two protocols would lose accesses to all those operators, unless we make trampolines.

sharplet:
Yeah, we don't have to keep the incompatible changes, this should probably just be the deprecation. As the implementations have no longer moved, I'm not sure I see why we can't do this with the other protocols?

andersio:
If it is just marking it unavailable, it should be okay for a point release. Just that for other protocols, specifically `SignalProtocol` and `SignalProducerProtocol`, you would get tons of deprecation warnings on the extensions that cannot be suppressed.

sharplet:
I also just realised that it's not going to be possible to remove `SignalProtocol` and `SignalProducerProtocol` until we get recursive constraints in extensions (used for the `flatten` operator), which won't land until at least the Swift 4 timeframe. I don't think it would make sense to add a deprecation warning for these protocols until we have a version of that feature we can test with.

So, the current state of this PR:

- On Swift 3.1, this change emits deprecation warnings for all uses of `ActionProtocol`
- On Swift 3.0.x, this change emits an "invalid platform" warning _in ReactiveSwift_, but doesn't emit any warnings in calling code.

It's a shame that there's no way to suppress the deprecation warnings for uses internal to ReactiveSwift, and they're a little annoying. But I don't think I mind that for the sake of users. @ReactiveCocoa/reactiveswift I'd love to hear your thoughts on the matter of warnings internal to this codebase. It doesn't bother me, but I totally understand if others feel differently.

mdiep:
I don't mind internal deprecation warnings—so long as the next major release isn't far off.

If we really feel this is a pain, we could do it right before the last 1.x release.

andersio:
I think recursive constraint is not relevant here, but parametetized extensions. Sadly, it seems the 2.0 plan has to be scrapped then.

sharplet:
Maybe both? This definitely fails under Swift 3.1 with a recursive constraint error:

```swift
// Flatten.swift
extension Signal where Value == Signal ...
```

andersio:
Hmm, perhaps. I have a branch with producer operators converted, e.g.
```swift
extension SignalProducer where Value: _SignalProducerProtocol, Error == Value.Error {
```

But it got hit in compiling the tests with issues in overloading somehow specific to concrete type extensions, and it so far affects `attempt` and `then`.

SR: https://bugs.swift.org/browse/SR-3873

Sources/Action.swift
```diff
@@ -207,6 +207,7 @@ private struct ActionState {
 }
 
 /// A protocol used to constraint `Action` initializers.
+@available(swift, deprecated: 3.1, message: "This protocol is no longer necessary and will be removed in a future version of ReactiveSwift. Use Action directly instead.")
```
andersio: Since it cannot be suppressed anyway, we may deprecate it for 3.0.x too.

sharplet: I'm not so sure: currently it only generates a warning inside ReactiveSwift, so I like the idea that it won't generate any warnings for consumers.

andersio: This would happen when building through Carthage. https://travis-ci.org/ReactiveCocoa/ReactiveSwift/jobs/201869272#L821

sharplet:
Yeah, I feel like that's better than introducing deprecation warnings into consuming projects unnecessarily.

andersio:
@sharplet Oops, it still appears in the build log.
https://travis-ci.org/ReactiveCocoa/ReactiveSwift/jobs/199639595#L825

andersio:
Tried `-suppress-warnings` in https://github.com/ReactiveCocoa/ReactiveSwift/pull/254/commits/6eb9671d858c80c7458ee1f9917e6cd3e9e21ddb and it works.

andersio:
@sharplet May you please rebase it on `master`?

andersio:
@sharplet With #272 merged, could you please rebase it on master and strip the `swift(>=3.1)` constraint of the deprecation?

PR #254, opened 2017-02-13
Reformulate unidirectional binding protocols.

andersio:
Related: #241

Just the implementation. Should appear as the same to general end users.

1. Renaming: `BindingSource` and `BindingTargetProvider`.

1. `BindingTargetProvider` now requires all conforming type to return a concrete `BindingTarget`. Moreover, `BindingTargetProvider` is no longer class bound.

1. The special mechanic for `BindingTarget` has been removed with (2) in place.


inamiy:
LGTM! How about removing `func consume` and just use `public let consume: (Value) -> Void`?

andersio:
@inamiy @mdiep Rename it to `action`, perhaps?

inamiy:
@andersio
From your comment (with analogy to `Observer.action`), I found interesting that `Observer` can also conform to `BindingTargetProvider` with an inifinite lifetime given.
Maybe wanna try this idea in this PR or next?

And with above taken into consideration, I thought `func send(value:)` could be another option.
But, I also think `action` is OK.

andersio:
@inamiy It is possible to pack an `Observer<Value, NoError>` in `BindingTarget`. But I am not quite sure if it is useful in practice.

inamiy:
@andersio Hm, maybe it was too much unnecessary abstraction 😶

andersio:
The patch is rebased on `master`, and is now source compatible with 1.0.

andersio:
https://github.com/ReactiveCocoa/ReactiveSwift/pull/254/commits/6eb9671d858c80c7458ee1f9917e6cd3e9e21ddb suppresses the deprecation warnings when building with Carthage.

PR #255, opened 2017-02-13
Update combineLatest and combineLatest(with:) documentation.

calebd:


Sources/SignalProducer.swift
```diff
@@ -679,6 +682,9 @@ extension SignalProducerProtocol {
 	/// - note: If either input is interrupted, the returned producer will also
 	///         be interrupted.
 	///
+	/// - note: The returned producer will not complete until both inputs send
```
sharplet: How about "until both inputs complete"?

calebd: Sounds good to me.

mdiep:
Thanks! Really appreciate these sorts of PRs.

PR #257, opened 2017-02-14
Rename scheduler protocols.

andersio:
Since there isn't a name conflict, the protocols should drop the suffix per the Swift API Guidelines. The renaming is source compatible.


mdiep:
Good idea. 👍 

Love that we can do this and maintain source compatibility.

PR #260, opened 2017-02-15
Rename `Lifetime.makeLifetime` to `Lifetime.make`.

andersio:


sharplet:
On the grounds of "omit needless words" this seems like a good change. I was initially concerned about just using "make" because of ambiguity using leading dot syntax. But now that I think about it, leading dot syntax isn't possible in this case because the method returns a tuple, so 👍 from me.

The documentation will need to be updated too. Might be worth just grepping the project for any use of `makeLifetime`.

mdiep:
I don't love either option, but this seems like an approvement. 👍 

This will need to deprecate the old version before it can be merged.

andersio:
@mdiep `Lifetime.makeLifetime` was not part of any release.

sharplet:
@andersio Will you be able to update the docs and examples for this too?

andersio:
I already grepped the project. It doesn't seem there is anything else.

sharplet:
The README and Framework Overview both have references to `makeLifetime`.

andersio:
Whoops.

andersio:
Whoops.

PR #264, opened 2017-02-16
Rename the result enum cases of validating property.

andersio:
* `success` becomes `valid`.
* `substitution` becomes `coercedValid`.
* `failure` becomes `invalid`.

The names were taken directly from [Cocoa Bindings and Key Value Validation](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/CocoaBindings/Concepts/MessageFlow.html#//apple_ref/doc/uid/TP40002149-BCICADHC). The main concern from me which drives the change is that the word `substitution` is _ambiguous_, while it in fact mimics the "coerced valid value" in Key Value Validation.

### DIFF
--- a/Sources/ValidatingProperty.swift
+++ b/Sources/ValidatingProperty.swift
@@ -10,13 +10,13 @@ import Result
 /// ```
 /// let root = MutableProperty("Valid")
 /// let outer = MutableValidatingProperty(root) {
-///   $0 == "Valid" ? .success : .failure(.outerInvalid)
+///   $0 == "Valid" ? .valid : .invalid(.outerInvalid)
 /// }
 ///
-/// outer.result.value        // `.success("Valid")
+/// outer.result.value        // `.valid("Valid")
 ///
 /// root.value = "🎃"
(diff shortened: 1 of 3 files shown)


Sources/ValidatingProperty.swift
```diff
@@ -249,53 +255,53 @@ public final class ValidatingProperty<Value, ValidationError: Swift.Error>: Muta
...
-	case success
+	/// The proposed value is valid.
+	case valid
 
-	/// The value is invalid, but the validator can provide a substitution derived
-	/// from the invalid value that would be considered valid.
-	case substitution(Value, Error?)
+	/// The proposed value is invalid, but the validator coerces it into a
+	/// replacement which it deems valid.
+	case coercedValid(Value, Error?)
```
mdiep: This is a little awkward IMO. Maybe just name it `case coerced`?

andersio:
https://travis-ci.org/ReactiveCocoa/ReactiveSwift/builds/204064960

CI has passed, but somehow it wasn't shown here.

PR #265, opened 2017-02-16
Fixed a race condition in `MutableValidatingProperty`.

andersio:


Sources/ValidatingProperty.swift
```diff
@@ -78,15 +78,19 @@ public final class MutableValidatingProperty<Value, ValidationError: Swift.Error
...
-				}
-
-				if let value = writebackValue {
-					isSettingInnerValue = true
-					inner.value = value
-					isSettingInnerValue = false
+				// Acquire the lock of `inner` to ensure no modification happens until
+				// the validation logic here completes.
+				inner.withValue { _ in
+					let writebackValue: Value? = mutableResult.modify { result in
```
andersio: Since `mutableResult` might change as a result of value events from `inner`, we need to block modifications of `inner` until the validator is done.

mdiep:
Can we add a test for this?

andersio:
The [theoretical contention window](https://github.com/ReactiveCocoa/ReactiveSwift/pull/265/files#diff-42b514003f72de821971eab375eda51dL89) is too small that even brute forcing doesn't seem to be able to reveal it. It happens after the validating property has returned from the setter of `inner`, but before `isSettingInnerValue` is flipped back to `false`. During this window, if `inner` is updated, [the new value would not be evaluated by the validator](https://github.com/ReactiveCocoa/ReactiveSwift/pull/265/files#diff-42b514003f72de821971eab375eda51dL76) due to an asserted `isSettingInnerValue`.

PR #266, opened 2017-02-17
Drop the minor version in `.swift-version`.

andersio:
This is speculative. But apparently most of the Swift Pods that have a working CocoaDocs docset happen to specify only `3.0` for `.swift-version`.


andersio:
Heh. This is fun. The Linux build fails because of this.

ikesyo:
The file is originally for swiftenv, not for CocoaPods.

ikesyo:
The underlying problem is that valid `SWIFT_VERSION` build setting value is different per Xcode versions:

- Xcode 8.0: `3.0`
- Xcode 8.1: `3.0` or `3.0.1`
- Xcode 8.2: `3.0` only (can't use `3.0.2`)

I think this should be addressed on CocoaPods or CocoaDocs side.

PR #267, opened 2017-02-17
Fixed a race condition in `Signal` terminal event handling.

andersio:


Sources/Signal.swift
```diff
@@ -316,10 +316,15 @@ public final class Signal<Value, Error: Swift.Error> {
 					if case let .alive(snapshot) = s.state {
 						var observers = snapshot.observers
 						observers.remove(using: token)
-						s.state = .alive(AliveState(observers: observers,
-						                            retaining: observers.isEmpty ? nil : self))
+
+						withExtendedLifetime(snapshot.retaining) {
```
sharplet: What is `retaining` here? This line reads a little weirdly to me, I wonder if we can improve the name.

Sources/Signal.swift
```diff
@@ -316,10 +316,15 @@ public final class Signal<Value, Error: Swift.Error> {
...
+							s.state = .alive(AliveState(observers: observers,
+							                            retaining: observers.isEmpty ? nil : self))
```
sharplet: Thoughts on extracting this to a local `newState` so we can remove the awkward formatting over two lines?

Sources/Signal.swift
andersio:
```
	/// A self-retaining reference. It is set when there are one or more active
	/// observers.
```

andersio:
@sharplet Addressed both in code and in the commit. The implementation has been updated too.

Sources/Signal.swift
```diff
@@ -313,13 +313,24 @@ public final class Signal<Value, Error: Swift.Error> {
...
 						observers.remove(using: token)
-						s.state = .alive(AliveState(observers: observers,
-						                            retaining: observers.isEmpty ? nil : self))
+
+						// Ensure the old signal state snapshot does not deinitialize before
+						// `updateLock` is released. Otherwise, it might result in a
+						// deadlock in cases where a `Signal` legitimately receives terminal
+						// events recursively as a result of the deinitialization of the
+						// snapshot.
+						withExtendedLifetime(snapshot) {
```
andersio: This is not strictly necessary, since `snapshot` is guaranteed to live till the end of the scope. But it serves a documentation purpose, and would be optimised away by the compiler IIRC.

Tests/ReactiveSwiftTests/ActionSpec.swift
```diff
@@ -114,6 +114,21 @@ class ActionSpec: QuickSpec {
...
 
+			it("should not deadlock") {
+				final class ViewModel {
+					let action2 = Action<(), (), NoError> { SignalProducer(value: ()) }
+				}
+
+				let action1 = Action<(), ViewModel, NoError> { SignalProducer(value: ViewModel()) }
+
+				action1.values
+					.flatMap(.latest) { viewModel in viewModel.action2.values.map { _ in viewModel } }
```
andersio: The deadlock happens when the observer disposable releases the closure `{ _ in viewModel }` here. The mapped signal's `updateLock` is not released while the `viewModel` is released, [causing a deadlock when the terminal event is eventually propagated to the mapped signal](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/Signal.swift#L157).

PR #268, opened 2017-02-17
Fix scheduler's markup documentation

eimantas:
After changing some parameter names for scheduler instance methods markup reported "No description" for those parameters. This PR fixes the issue.

### DIFF
--- a/Sources/Scheduler.swift
+++ b/Sources/Scheduler.swift
@@ -20,7 +20,7 @@ public protocol SchedulerProtocol {
 	/// When the work is executed depends on the scheduler in use.
 	///
 	/// - parameters:
-	///   - action: Action to schedule.
+	///   - action: The action to be scheduled.
 	///
 	/// - returns: Optional `Disposable` that can be used to cancel the work
 	///            before it begins.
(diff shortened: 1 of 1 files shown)


Sources/Scheduler.swift
```diff
@@ -19,6 +19,9 @@ public protocol SchedulerProtocol {
 	///
 	/// When the work is executed depends on the scheduler in use.
 	///
+	/// - parameters:
+	///   - action: Action to schedule.
```
andersio: This could use an article, and perhaps should it be in passive voice?
s/Action/The action/.

Sources/Scheduler.swift
```diff
@@ -50,8 +53,8 @@ public protocol DateSchedulerProtocol: SchedulerProtocol {
 	///
 	/// - parameters:
 	///   - date: Starting time.
-	///   - repeatingEvery: Repetition interval.
-	///   - withLeeway: Some delta for repetition.
+	///   - interval: Repetition interval.
+	///   - leeway: Some delta for repetition.
```
andersio: Ditto

mdiep:
Thanks!

PR #272, opened 2017-02-18
Suppress all warnings when building with CocoaPods.

andersio:
Suppress all warnings when building with the Release configuration, so that the internal deprecation warnings would not be visible when ReactiveSwift is built using ~~Carthage and~~ CocoaPods.

SwiftPM does not provide a way to pass compiler flags, so it won't be covered by the patch.

Usage: #248, #254

### DIFF
--- a/ReactiveSwift.podspec
+++ b/ReactiveSwift.podspec
@@ -18,4 +18,6 @@ Pod::Spec.new do |s|
   # Directory glob for all Swift files
   s.source_files  = "Sources/*.{swift}"
   s.dependency 'Result', '~> 3.1'
+
+  s.pod_target_xcconfig = {"OTHER_SWIFT_FLAGS[config=Release]" => "-suppress-warnings" }
 end


mdiep:
I'm not sure that this is necessary.

As of Carthage 0.19, warnings are no longer passed through, so this should be a lot less necessary.

But I suppose this is probably okay for CocoaPods?

ReactiveSwift.xcodeproj/project.pbxproj
```diff
@@ -1299,6 +1299,7 @@
 				GCC_OPTIMIZATION_LEVEL = 0;
 				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
 				MACOSX_DEPLOYMENT_TARGET = 10.9;
+				OTHER_SWIFT_FLAGS = "-suppress-warnings";
```
NachoSoto: 👎 this would hide potential problems that might not be visible in debug.

ReactiveSwift.podspec
```diff
@@ -18,4 +18,6 @@ Pod::Spec.new do |s|
   # Directory glob for all Swift files
   s.source_files  = "Sources/*.{swift}"
   s.dependency 'Result', '~> 3.1'
+
+  s.pod_target_xcconfig = {"OTHER_SWIFT_FLAGS[config=Release]" => "-suppress-warnings" }
```
ikesyo: Warnings can be silenced by using `inhibit_all_warnings!` in `Podfile` at users side, so I'm not confident if this is really needed.

andersio: Wasn't aware of the option. Thanks!

andersio: The point is to hide warnings regardless of users' preferences, since ideally the warnings are relevant to us the maintainers only.

PR #276, opened 2017-02-23
Remove unnecessary `@inline`

ChristopherRogers:
Swift 3.1 beta complains that these methods are not qualified to have `@inline(__always)` because `value` is not public:

> Atomic.swift:65:42: Let 'value' is private and cannot be referenced from an '`@inline(__always)`' function

Since `@inline(__always)` is not officially supported, this needs to be fixed here, rather than in Swift. See [here](https://bugs.swift.org/browse/SR-4013?focusedCommentId=21981&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21981).

Related commit in Swift:
https://github.com/apple/swift/commit/50b5d01dd7a023b4d7c7a77019c1c26ee507da03

According to that commit message, it seems that `@inline` is now only for public, resilient API.

### DIFF
--- a/Sources/Atomic.swift
+++ b/Sources/Atomic.swift
@@ -58,7 +58,6 @@ internal struct UnsafeAtomicState<State: RawRepresentable>: AtomicStateProtocol 
 	/// - returns:
 	///   `true` if the current state matches the expected state. `false`
 	///   otherwise.
-	@inline(__always)
 	internal func `is`(_ expected: State) -> Bool {
 		return OSAtomicCompareAndSwap32Barrier(expected.rawValue,
 		                                       expected.rawValue,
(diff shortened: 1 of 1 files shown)


ikesyo:
Not `public`, but `@_versioned internal` properties should work.

ChristopherRogers:
Changing `value` to `internal` will allow it to build, since the access level is the same as the methods accessing it. However, it seems like it's more important to keep this property hidden to prevent being directly accessed by accident.

mdiep:
Thank you for raising this and opening a PR! Removing these does seem like the correct thing to do IMO.

andersio:
_Supposedly_ the compiler should be able to inline these without the attribute. However it wasn't the case back in Swift 3.0.

PR #281, opened 2017-02-25
Restructure the source layout to support multiple targets in SwiftPM.

andersio:
TL;DR: It moves `Sources/*` to `Sources/ReactiveSwift/*`. Otherwise, one can't add new targets in SwiftPM since it assumes the package being single module from the flattened layout.


mdiep:
Do you anticipate adding more targets?

andersio:
See #282.

andersio:
Close together with #282.

PR #282, opened 2017-02-25
Adopt `os_unfair_lock`.

andersio:
Preceded by #124.

`os_unfair_lock` is a replacement of `OSSpinlock` with the ability to resolve priority inversion.

Since our needs of mutual exclusion do not rely on thread wake order or fairness, `os_unfair_lock` suits us better than pthread mutex and [should perform better too](https://gist.github.com/steipete/36350a8a60693d440954b95ea6cbbafc).

Starvation could happen, but IMO practically only in high frequency concurrent event emission which is rare and resolvable with back pressure operators or solutions on the user's end.

#### Implementation

`PosixThreadMutex` is replaced by `UnsafeUnfairLock`, which is a wrapper of `os_unfair_lock` with a fallback path to `pthread_mutex_t` for unsupported deployment targets and non-Cocoa platforms.

All use of `NSLock` is replaced by `UnsafeUnfairLock` or its ref-counted wrapper `UnfairLock`.

`UnsafeUnfairLock` is mostly implemented in C to avoid the overhead of `#available` dynamic checks and [a current limitation in API availability propagation](https://bugs.swift.org/browse/SR-4079).

```
// Uncontended locking test, 1000000 iterations.
// https://gist.github.com/andersio/215d3f469e8c428b0b00bcf9706f8e52
avg: 42.2285819999544 ns
avg: 37.7035349999262 ns (no #available)
avg: 44.2758420005849 ns (pthread)
```

When using Swift Package Manager, the C portion is compiled separately as `OSLocking`, since SwiftPM does not support mixed-language targets.

~~TODO: Benchmark `UnsafeUnfairLock` and `pthread_mutex_t` in Swift.~~

Uncontended locking test, 10 million iterations, `-Owmo` and `-Os` [(Source)](https://gist.github.com/andersio/0e852ce9166c798e2ea3b4a3579803ab)
```
UnsafeUnfairLock: 42.5839338052302 ns
os_unfair_lock: 38.0752214928745 ns
pthread_mutex: 47.3852283026485 ns
```

Contested locking test [(Source)](https://gist.github.com/steipete/36350a8a60693d440954b95ea6cbbafc)
```
UnsafeUnfairLock: 0.438 s
os_unfair_lock: 0.376 s
pthread_mutex: 8.319 s
```


andersio:
Experimenting ways to move stuff into the Swift world. 🎃

Sources/ReactiveSwift/Atomic.swift
```diff
@@ -123,43 +127,135 @@ internal struct UnsafeAtomicState<State: RawRepresentable>: AtomicStateProtocol
...
+	internal func unlock() {
+		_lock.unlock()
+	}
+
+	internal func `try`() -> Bool {
+		return _lock.try()
+	}
+}
+
+@_fixed_layout
```
mdiep: Is this a required part of this change? We should stay away from unsupported attributes if at all possible.

.Package.test.swift
```diff
@@ -2,6 +2,10 @@ import PackageDescription
 
 let package = Package(
     name: "ReactiveSwift",
+    targets: [
+        Target(name: "ReactiveSwift", dependencies: ["OSLocking"]),
+        Target(name: "OSLocking")
+    ],
```
mdiep: Does this still require a separate target if it's in Swift?

andersio: Could try to move the rest of the stuff. Though as a side note, these C stuff are introduced to dodge [compiler enforced `@available` runtime checks in `lock()` and `unlock()`](https://bugs.swift.org/browse/SR-4079).

Sources/ReactiveSwift/Atomic.swift
andersio: No. Gonna remove it.

.Package.test.swift
andersio: I think it is not possible to move the rest of the C stuff to Swift though.

All uses of `_ras_*` function pointers in Swift are being resolved at link time. This is what the disassembly shows.

However, if written in Swift, the compiler would enforce availability, so have such pointers would trigger work at runtime. At best case, it would have to be a lazily initialised top-level (static) variable, which means extra overhead in obtaining the function pointer.

Until [SR-4079](https://bugs.swift.org/browse/SR-4079) is resolved, dropping down to C for performance is likely inevitable in this specific case.

andersio:
Since the C part is not going anywhere, it might as well be reverted back to a full-C implementation with annotated header for Swift. When I disassemble the `lock` method written in Swift, somehow the Swift compiler has injected a lot of instructions with unclear purpose between the `os_unfair_*` calls and the enum tag checking/unpacking. C is after all fairly predictable.

.travis.yml
```diff
@@ -18,6 +18,13 @@ xcode_workspace: ReactiveSwift.xcworkspace
 matrix:
   include:
     - xcode_scheme: ReactiveSwift-macOS
+      osx_image: xcode8
```
andersio: This is added to test how it compiles and runs on macOS 10.11.

andersio:
Close for now. It doesn't seem to carry a substantial benefit for the complexity it brought to the project. It would probably be better to wait until [the language can properly express it](https://bugs.swift.org/browse/SR-4079).

PR #285, opened 2017-02-28
Utilize `filterMap`

ikesyo:


andersio:
It seems Nimble causes the watchOS test to fail somehow.

ikesyo:
Hmm that is affected by https://github.com/Carthage/Carthage/issues/1792.

PR #292, opened 2017-03-10
Point the API reference link to CocoaDocs.

andersio:
Welp, less maintenance.


mdiep:
I guess either way is fine with me. 🤷🏼‍♂️

ikesyo:
😞 http://blog.cocoapods.org/CocoaDocs-Documentation-Sunsetting/

andersio:
Guess we should fix the existing script then.
🙈

Issue #297, opened 2017-03-25
Add a flatten strategy to limit concurrency


mdiep:
`concat` limits concurrency to 1 signal. `merge` allows unlimited concurrency. It'd be nice to have something in-between.

ReactiveObjC has this as [`flatten:`](https://github.com/ReactiveCocoa/ReactiveObjC/blob/a417e72712fa4eeea977daa8fdc46417db794aaf/ReactiveObjC/RACSignal%2BOperations.h#L211-L224).

This is useful for things like https://github.com/Carthage/Carthage/pull/1841.

NachoSoto:
Or change `.merge` to `.merge(maxConcurrency: UInt)` with `static let merge: FlattenStrategy = .Merge(UInt.max)`.

mdiep:
And `static let concat = .merge(1)`?

NachoSoto:
As long as it's not less performance, merging (no pun intended) the implementations would be amazing.

andersio:
I'd call it `bound`. 🙂

andersio:
Implemented in #298.

PR #298, opened 2017-03-28
Introduce `FlattenStrategy.concurrent`.

andersio:
Implements #297.

`concat` and `merge` are now aliases of `concurrent(limit: 1)` and `concurrent(limit: .max)`, respectively. The iterative process to start producers is preserved.

**WIP:**
- [x] Test cases for the new "merge up to N producers at any point of time" behaviour.

### DIFF
--- a/Sources/Flatten.swift
+++ b/Sources/Flatten.swift
@@ -376,34 +376,34 @@ extension SignalProtocol where Value: SignalProducerProtocol, Error == Value.Err
 				let producerState = UnsafeAtomicState<ProducerState>(.starting)
 				let deinitializer = ScopedDisposable(ActionDisposable(action: producerState.deinitialize))
 
-				withExtendedLifetime(deinitializer) {
-					producer.startWithSignal { signal, inner in
-						let handle = disposable.add(inner)
+				producer.startWithSignal { signal, inner in
+					let handle = disposable.add(inner)
 
-						signal.observe { event in
-							switch event {
-							case .completed, .interrupted:
-								handle.remove()
+					signal.observe { event in
+						switch event {
+						case .completed, .interrupted:
+							handle.remove()
 
-								let shouldComplete: Bool = state.modify { state in
-									state.activeCount -= 1
-									return state.shouldComplete
-								}
-
-								withExtendedLifetime(deinitializer) {
-									if shouldComplete {
-										observer.sendCompleted()
-									} else if producerState.is(.started) {
-										startNextIfNeeded()
-									}
-								}
-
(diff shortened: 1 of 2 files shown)


Sources/Flatten.swift
```diff
@@ -389,10 +395,11 @@ extension SignalProtocol where Value: SignalProducerProtocol, Error == Value.Err
 						}
 					}
 				}
-				state.modify { $0.isStarting = false }
+
+				isStarting = false
```
mdiep: Is assigning to a `Bool` an atomic operation? This write could happen concurrently with the read above. I think it needs to at least be in its own `Atomic`.

mdiep:
> Test cases for concurrent(limit: 0), which would queue all inner producers indefinitely until the outer producer terminates.

I'm not sure that makes sense. Maybe we just add a precondition that count must be >0?

Sources/Flatten.swift
```diff
@@ -341,46 +354,45 @@ extension SignalProducerProtocol where Value: PropertyProtocol {
...
+					signal.observe { [cleanup] event in
 						switch event {
 						case .completed, .interrupted:
-							handle?.remove()
-							
-							let shouldStart: Bool = state.modify {
-								$0.active = nil
-								return !$0.isStarting
+							handle.remove()
+							_ = cleanup
```
mdiep: Is this enough to guarantee that `producerState` remains initialized until after both the read and the write?

NachoSoto:
Do we have a way to test for regressions in performance/memory in `.concat`?

mdiep:
Looking at it, I'm not sure why this would cause any performance issues. The code is basically the same.

Sources/Flatten.swift
andersio: Yes. Its lifetime is limited by the closure or the `while` loop body scope, whichever is longer.

mdiep:
It looks like those tests don't compile on Linux.

Sources/Flatten.swift
```diff
@@ -341,58 +354,61 @@ extension SignalProducerProtocol where Value: PropertyProtocol {
...
-					signal.observe { event in
-						switch event {
-						case .completed, .interrupted:
-							handle?.remove()
-							
-							let shouldStart: Bool = state.modify {
-								$0.active = nil
-								return !$0.isStarting
-							}
+				withExtendedLifetime(deinitializer) {
```
mdiep: This could probably surround just the `producerState.setStarted()` below?

andersio: True,

Issue #291, opened 2017-03-07
Buffering Action


andersio:
It seems a buffering `Action` might be a useful addition. As compared to a regular `Action`, it

1. does not automatically disable during execution;
2. buffers inputs received when the `Action` is executing; and
3. has an optional queue coalescer.

Possible use cases:

1. Queues of alert/modal prompts.
1. Asynchronous worker requiring mutual exclusion in general, e.g. creating/restoring an `NSWindow` w.r.t. a given object, without the use of infinite `retry` to workaround disabled actions.

A possible implementation would be:

1. Adding a new wrapper `Buffered<Value>` with a constraining protocol `ActionBuffering`.
2. Specialization in `Action` when `Value: ActionBuffering`
3. An `apply` overload which takes `Value.Value` and returns a `SignalProducer` that emits `Error` instead of `ActionError<Error>`.

e.g.

```swift
let documentOpener: Action<Buffered<Token>, (), NoError>

documentOpener = Action(bufferingUpTo: .max) { token -> SignalProducer<(), NoError> in
    return retrieveDocument(for: token).flatMap(.concat, transform: openWindow(for:))
}

// No ambiguity here, since `Buffered<Value>` != `Value`.
let worker: SignalProducer<(), NoError> = documentOpener.apply(restoredToken)
worker.start()
```

mdiep:
Another possible implementation would be to add a configuration type.

```
extension Action {
  // These names are all terrible
  enum SomeName {
    case disabled // send disabled if already executing
    case joined // send completed when the current execution ends
    case buffered // execute after the current execution ends
  }
}
```

But I'm not sure we need this. Without strong demand for such behavior, I'd be very hesitant to add it.

sharplet:
This seems similar to how `FlattenStrategy` is used. It also reminds me of `RACCommand`'s concurrent execution mode. Thoughts on modelling this after `FlattenStrategy` and adding a new case called `reject` (or maybe`disable`, etc.) for the current behaviour?

sharplet:
FWIW this is something I've reached for in the past too (specifically the buffering/concat behaviour).

RuiAAPeres:
Hello. 👋  Thanks for opening this issue. Due to inactivity, we will soft close the issue. If you feel that it should remain open, please let us know. 😄

PR #301, opened 2017-03-30
Configurable execution strategy in `Action`.

andersio:
Implements #291.

`Action`s can now be configured to use different execution strategies other than the default "disable-when-executing" strategy.

~~Two new options have been added to `Action`: `strategy` and `disablesWhenExecuting`.~~

~~The current behaviour is equivalent to `disablesWhenExecuting` being true (the default), regardless of the strategy used.~~

~~So far `FlattenStrategy` and the flattening operator are reused in the implementation.~~ However, the queue coalescer ~~and the buffering cap~~ envisioned in #291 ~~are~~ is not included in the PR, and _might_ require non-trivial changes for maximum reuse.

~~P.S. The `merge` strategy allows `Action` to be concurrent.~~

---

WIP
- [ ] Test cases

### DIFF
--- a/Sources/Action.swift
+++ b/Sources/Action.swift
@@ -20,19 +20,64 @@ import Result
 public final class Action<Input, Output, Error: Swift.Error> {
 	private struct ActionState<Value> {
 		var isEnabled: Bool {
-			return isUserEnabled && !isExecuting
+			return isUserEnabled && executingCount < queueCapacity
 		}
 
-		var isUserEnabled: Bool
-		var isExecuting: Bool
+		var isUserEnabled: Bool = true
+		var executingCount: UInt = 0
+		let queueCapacity: UInt
 		var value: Value
+
+		init(queueCapacity: UInt, value: Value) {
+			self.queueCapacity = queueCapacity
+			self.value = value
+		}
 	}
 
-	private let execute: (Action<Input, Output, Error>, Input) -> SignalProducer<Output, ActionError<Error>>
-	private let eventsObserver: Signal<Signal<Output, Error>.Event, NoError>.Observer
-	private let disabledErrorsObserver: Signal<(), NoError>.Observer
+	public enum ExecutionStrategy {
+		/// Only one worker is allowed to be executing at a time. Any further execution
+		/// attempt during the execution is rejected with `ActionError.disabled`.
+		case exclusive
 
-	private let deinitToken: Lifetime.Token
+		/// Only one worker is allowed to be executing at a time. Any further execution
+		/// attempt during the execution would succeed immediately, and interrupt the
+		/// previous worker (if any).
+		case latest
+
+		/// Only one worker is allowed to be executing at a time. Any further execution
+		/// attempt during the execution would be queued.
+		case concat
+
+		fileprivate var queueCapacity: UInt {
+			switch self {
+			case .exclusive:
+				return 1
+
+			case .concat, .latest:
+				return .max
+			}
+		}
+
+		fileprivate var flattenStrategy: FlattenStrategy {
+			switch self {
+			case .exclusive, .concat:
+				return .concat
+
+			case .latest:
+				return .latest
+			}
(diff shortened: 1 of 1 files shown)


mdiep:
I think it makes more sense to create a custom strategy type for `Action` that encompasses the use of both `strategy` and `disablesWhenExecuting` here. That seems like it'd be the easiest to understand—I don't think all `FlattenStrategy`s will make sense here (what would `.race` do?).

Sources/Action.swift
```diff
@@ -10,15 +10,91 @@ import Result
 /// Actions enforce serial execution. Any attempt to execute an action multiple
 /// times concurrently will return an error.
 public final class Action<Input, Output, Error: Swift.Error> {
+	public enum ExecutionStrategy {
+		/// The default execution strategy. Only one worker is allowed to be
+		/// executing at a time. Any further start attempt during the execution
+		/// is rejected with `ActionError.disabled`.
+		case `default`
```
mdiep: I don't think `default` is a great name for this. We should find a name that describes the behavior.

Sources/Action.swift
```diff
@@ -10,15 +10,91 @@ import Result
...
+		///    been reached; or
+		/// 3. be rejected with `ActionError.disabled` if the capacity limit has
+		///    been reached.
+		///
+		/// - parameters:
+		///   - limit: The concurrent limit of the `Action`.
+		///   - capacity: The capacity limit of the `Action`.
+		///
+		/// - note: `capacity` is inclusive of `limit`.
+		case concurrent(limit: UInt, capacity: UInt)
```
mdiep: I'm not sure that opening it up to concurrency is a good idea. That adds quite a bit of complexity.

Maybe we should limit it to exclusive strategies to start?

Sources/Action.swift
```diff
@@ -10,15 +10,91 @@ import Result
...
 	private let deinitToken: Lifetime.Token
 
 	private let executeClosure: (_ state: Any, _ input: Input) -> SignalProducer<Output, Error>
-	private let eventsObserver: Signal<Event<Output, Error>, NoError>.Observer
 	private let disabledErrorsObserver: Signal<(), NoError>.Observer
 
 	/// The lifetime of the Action.
 	public let lifetime: Lifetime
 
+	private let dispatcher: (SignalProducer<Output, Error>) -> Void
```
mdiep: This could use documentation.

andersio:
Four strategies are left: `concat`, `exclusive`, `latest` and `replay`. That said I am not quite sure about the implementation for now... Meant to use `flatten` but found it a bit of a hack to fit in.

Sources/Action.swift
```diff
@@ -20,19 +20,64 @@ import Result
...
+		init(queueCapacity: UInt, value: Value) {
+			self.queueCapacity = queueCapacity
+			self.value = value
+		}
 	}
 
-	private let execute: (Action<Input, Output, Error>, Input) -> SignalProducer<Output, ActionError<Error>>
-	private let eventsObserver: Signal<Signal<Output, Error>.Event, NoError>.Observer
-	private let disabledErrorsObserver: Signal<(), NoError>.Observer
+	public enum ExecutionStrategy {
```
mdiep: We should make this a `struct` like we did for `FlattenStrategy`.

arnopoulos:
I'm working on a project where we're trying to prevent the user from spamming the server and this would be very helpful. Are there any updates on the progress of this?

andersio:
@arnopoulos I do not have time to move this forward in the foreseeable future. 😞

arnopoulos:
@andersio That's too bad because this definitely could be a useful feature. If I had a deeper understanding of the underlying architecture of ReactiveSwift, I would try to help out

ataibarkai:
is this feature addition still on hold for the foreseeable future?

andersio:
Not being actively worked on.

PR #302, opened 2017-03-30
Signal.replay

andersio:
Not sure if this is appreciated given the stance on `buffer(_:)`, since `Signal.replay` is essentially `buffer(_:)` minus _some_ of its semantic culprits (IMO). Say `Signal.replay` requires one to explicitly replay something hot.

Anyway, I rolled this from a feature parity standpoint, both in terms of the `Signal`-`SignalProducer` parity and the parity with Rx.

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -2431,3 +2431,210 @@ extension SignalProtocol where Error == AnyError {
 		}
 	}
 }
+
+extension SignalProtocol {
+	/// Creates a new `SignalProducer` that will multicast values emitted by
+	/// `self`, up to `capacity`. This means that all clients of the returned
+	/// `SignalProducer` will see the same version of the emitted values/errors.
+	///
+	/// If you find yourself needing *the current value* (the last buffered
+	/// value) you should consider using `PropertyType` instead, which, unlike
+	/// this operator, will guarantee at compile time that there's always a
+	/// buffered value. This operator is not recommended in most cases, as it
+	/// will introduce an implicit relationship between the original client and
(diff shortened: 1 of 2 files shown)


mdiep:
Yeah, I don't think we want this. I think it's best to minimize the exposure of replaying/buffering methods. Unless someone is asking for behavior like this with a really compelling use case, I'm 👎 to adding it.

AlexanderKaraberov:
@mdiep @andersio 

>  I think it's best to minimize the exposure of replaying/buffering methods
Could you please elaborate on this? Or if this was already explained/discussed somewhere point me to that resource. I tried to search in the repo issues and comments any explanation of why `SignalProducer.buffer` was removed and why replay/buffering is discouraged but with no luck...
I don't really see any drawbacks of those but maybe I'm not able to see the whole picture or lack some experience (e.g. RxJava contains `buffer`).

andersio:
Many “abuse” `buffer(0)` when a `Signal` is the more appropriate choice, while some use cases of `buffer(1)` have already been covered by properties.

That’s said I personally found buffering/replaying can be useful in some circumstances, e.g. deferring push notifications until the async re-authentication has completed.

mdiep:
It also erases the distinction between hot (`Signal`s) and cold (`SignalProducer`s) observables. These types of observables are often called _warm_. They make it difficult to reason about what will cause work to be performed. Experience has shown that it's usually best to model buffering explicitly in some form.

There's a lot of background in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2639.

PR #304, opened 2017-03-30
Move Signal and SignalProducer operators to the concrete types.

andersio:
`SignalProducer` heavily intertwines with `Signal` at its core, while both `SignalProducer` and `Signal` are already pretty generic with refinements & specialisation enabled through closures & type erasure, IMO mitigating the need of another hot signal type. Moreover, `Property` is not going to conform to either of them due to clashes in base names, even assuming the presence of higher-kinded types in Swift.

So with all these, I do not envision any value of semantic abstraction from these protocols.

The PR does not remove `SignalProtocol` and `SignalProducerProtocol` however, since they are still necessary for constraining flattening operators until we have parameterised extensions.

### DIFF
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -1259,22 +1259,6 @@ extension SignalProducer where Error == NoError {
 		return lift { $0.timeout(after: interval, raising: error, on: scheduler) }
 	}
 
-	/// Wait for completion of `self`, *then* forward all events from
-	/// `replacement`.
-	///
-	/// - note: All values sent from `self` are ignored.
-	///
-	/// - parameters:
-	///   - replacement: A producer to start when `self` completes.
-	///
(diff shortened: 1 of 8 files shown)


Sources/SignalProducer.swift
```diff
@@ -1329,6 +1306,12 @@ extension SignalProducer {
 	}
 }
 
+// FIXME: SWIFT_COMPILER_ISSUE
```
andersio: Yep.

https://bugs.swift.org/browse/SR-4436
https://bugs.swift.org/browse/SR-3873?filter=-2

andersio: For the ambiguity of `then`, my wild guess is that in concrete extensions, the compiler picks overloads with signatures with parameters resolved but no information about their original form. So the more specific overloads are ranked the same as the most generic overload.

Issue #74, opened 2016-11-01
Deadlock


codyrobb:
Hi, 

I am running into a deadlock issue. I’m not entirely sure why it is happening yet. Here is my setup (dumbed down as this is the only part breaking, can provide more if needed).

```swift
class ActionController {
	var delay: NSTimeInterval = 0.0
	let undoable = MutableProperty<Bool>(false)

	func start() {
		if delay > 0.0 {
			undoable.value = true
		}
		
		// More stuff
		// Eventually will commit once delay is hit
		// Can provide more if needed.
	}
	
	func undo() {
		guard undoable.value else {
			return
		}
		
		// Do some stuff
		
		/***** DEADLOCK HAPPENS ON LINE BELOW *******/
		undoable.value = false
		/********************************************/
	}
}
```

```swift
class MyViewModel {

	private let currentAction = MutableProperty<ActionController?>(nil)
	
	let undoEnabled = MutableProperty<Bool>(false)
	
	init() {
		undoEnabled <~ currentAction.producer
			.flatMap(.Concat) { action in
				guard let action = action else {
					return SignalProducer<Bool, NoError>(value: false)
				}
				
				return action.undoable.producer
			}
			.skipRepeats()
	}
	
	func commitAction() {
		let action = ActionController()
		action.delay = 10
		action.start()
	}
	
	func undoCurrentAction() {
		currentAction.value?.undo()
	}

	 func onActionCompleted() {
		// Method gets called when ActionController completes    
		currentAction.value = nil
	}

}
```

Both `commitAction()` and `undoCurrentAction()` get called from a `UIViewController` that the `MyViewModel` is powering and has subscribers to `undoEnabled`.

Example for testing:

```swift
        viewModel.undoEnabled.producer.startWithNext { [unowned self] enabled in
            if enabled {
                self.viewModel.undoCurrentAction()
            }
        }
```

The deadlock seems to be happening when calling `undo` and setting `undoable`’s `value` to `false`.

codyrobb:
After setting a breakpoint on `_NSLockError()`, it looks like it is line 66 in `Signal.swift` causing the lock and eventual breakdown.

This is one of my first bigger stabs at RAC4 after coming from RAC2.X. I might be misusing some things. Please let me know!

codyrobb:
Another note: I am using ReactiveCocoa 4.2.2, but this seemed to be a better place to post the issue as it appeared you guys were trying to get the Swift API oriented issues over here.

[...]

PR #308, opened 2017-03-31
Add `retryWhen`

inamiy:
This is an attempt to implement [`retryWhen`](http://reactivex.io/documentation/operators/retry.html) in ReactiveSwift.
I wrote simple version & Rx original version here:

- (1)  8754889: Simpler version with type: 
    - `(FlattenStrategy, (Error) -> SignalProducer<(), Error>) -> SignalProducer<Value, Error>`
- (2) 2f829f8: Rx original version with type:
    - `(FlattenStrategy, (Signal<Error, NoError>) -> Signal<(), Error>) -> SignalProducer<Value, Error>`

### Use case: Delay retry after error

```swift
// Using (1) Simple version
producer.retry(.merge) { error in
    return timer(3, on: mainScheduler).take(1).map { _ in () } // retry after 3 sec
}

// Using (2) Rx original version
producer.retry(.merge) { errors in
    return errors.flatMap { 
        return timer(3, on: mainScheduler).take(1).map { _ in () } // retry after 3 sec
    }.take(5)  // up to 5 times
}
```

(1) is obviously simpler and has a similar type as `flatMap` (which I think is clean), but it does not have error counting logic so (2) might be a better option.

But to implement (2), I had to change `NSLock` to `NSRecursiveLock` in `Signal.swift`.
(BTW I personally think using recursive-lock is a better choice in general)


inamiy:
```
FlattenStrategy_concurrent__should_asynchronously_merge_up_to_the_stated_limit__buffer_any_subsequent_producers_and_dequeue_them_in_the_submission_order, expected to equal <[0, 1, 2, 3]>, got <[0]>
```

CI error seems unrelated to this PR.

mdiep:
A few notes:

1. As an additive API, this doesn't need to hold up a 2.0 release.

2. I think the simpler `(Error) -> SignalProducer<(), Error>` variant is preferable.

3. I don't think this needs to take a flatten strategy. It seems like the returned signal should have a `take(first: 1)` that prevents multiple retries from the same error.

4. I don't think a recursive lock is going to be acceptable.

Do you have a particular use case in mind for this? Or are you suggesting it because you find it interesting?

inamiy:
@mdiep 
Use cases will be:
1. Delaying retry (as I just showed example on top comment)
2. Conditional retry based on `error` (e.g. retry only when error has status code XXX...)

> I don't think this needs to take a flatten strategy

Sounds simpler! I will just use `.merge` then 😉 

If we go with (1), I also need error counting to stop retrying forever, so below code might be better: 

```swift
public func retry(upTo count: Int, when: @escaping (Error) -> SignalProducer<(), Error>) -> SignalProducer<Value, Error> {
	precondition(count >= 0)

	if count == 0 {
		return producer
	} else {
		return flatMapError { error in
			return when(error).flatMap(.merge) {
				return self.producer.retry(upTo: count - 1, when: when)
			}
		}
	}
}
```

mdiep:
I wonder if we'd be better adding variants that cover those two cases. Those seem like the most common uses.

The first case would actually be really well suited to have some sort of timing function—an enum that covered both a constant time and an exponential backoff with some sort of upper limit.

inamiy:
@mdiep 
Adding variants sounds friendly, but for exponential backoff, I guess it will require additional state inside which will make (1)'s implementation more complicated.

On the other hand, if we go with (2), we can use `scan` to count errors and handle exponential delay calculation (and probably many more).

So... Though I wasn't aware of (2)'s powerfulness on the first place, I'm leaning toward (2) now 🤔

inamiy:
Updated the code to track current master branch.

Signature became simpler as follows, but still, `PthreadLock(recursive: true)` is required to avoid the crash in test cases.

```swift
func retry(when: ((Signal<Error, NoError>) -> Signal<(), Error>) -> SignalProducer<Value, Error>
```

Sources/Signal.swift
```diff
@@ -67,7 +67,7 @@ public final class Signal<Value, Error: Swift.Error> {
 			state = .alive(Bag(), hasDeinitialized: false)
 
 			stateLock = Lock.make()
-			sendLock = Lock.make()
+			sendLock = Lock.PthreadLock(recursive: true)
```
chuganzy: Agree with this change. Ref: https://github.com/ReactiveCocoa/ReactiveSwift/issues/74#issuecomment-472686906

andersio:
We should look at improving options for retrying, though use of recursive locks is a non-goal.

Closing due to inactivity.

PR #311, opened 2017-04-03
SwiftPM 3.1

ikesyo:
We can handle environment variables within `Package.swift` in Swift 3.1.


.travis.yml
```diff
@@ -77,9 +76,8 @@ matrix:
       before_install:
         - eval "$(curl -sL https://gist.githubusercontent.com/kylef/5c0475ff02b7c7671d2a/raw/9f442512a46d7a2af7b850d65a7e9bd31edfb09b/swiftenv-install.sh)"
       script:
-        - mv .Package.test.swift Package.swift
         - swift build
-        - swift test
+        - SWIFTPM_TEST_ReactiveSwift=YES swift test
```
sharplet: Could this be moved to `env`?

Package.swift
```diff
@@ -1,8 +1,18 @@
...
 let package = Package(
     name: "ReactiveSwift",
-    dependencies: [
-        .Package(url: "https://github.com/antitypical/Result.git", versions: Version(3, 2, 1)..<Version(3, .max, .max)),
-    ]
+    dependencies: {
+        var deps: [Package.Dependency] = [
+            .Package(url: "https://github.com/antitypical/Result.git", versions: Version(3, 2, 1)..<Version(3, .max, .max)),
+        ]
+        if ProcessInfo.processInfo.environment["SWIFTPM_TEST_ReactiveSwift"] == "YES" {
```
sharplet: What do you think about extracting a helper at the top level for this, say an `isSwiftPackagerManagerTest` computed property?

.travis.yml
ikesyo: I'd rather run `swift build` without the test dependencies so I added the env variable to explicitly `swift test`.

PR #313, opened 2017-04-05
Signal availability

andersio:
Implements #163.

Enable and disable the input side effect — that pipes values into a given `Signal` — in response to the number of active observers in the `Signal`.

It is designed for operators that can be proved to be memoryless in systems theory, e.g. `map`. Operators which transform the latest `value` event based on histories should not use the feature.

### Summary
1. Normal observers attached through `observe(_:)` and shorthands are always active.

1. Any `Signal` must handle the activation and deactivation of its observers.

1. Only if a `Signal` is declared stateless when initialized, it may notify to its upstreams its own activation and deactivation to enable or disable its input side effect. The notification channel is established by observing its upstreams with its `SignalChannel` (i.e. the new two-argument `observe(_:_:)`).

### Breaking change
If a `Signal` constructed by stateless operators, the side effect would not be evaluated when it has no observers even if the `Signal` is retained.

It is a further restriction on top of the current rule of automatic disposal.

```swift
var count = 0

// A mapped signal that is retained but without observer.
let mapped = signal.map { count += $0; return $0 + 1 }

observe.send(value: 10)

// Currently: Pass
// This PR: Assertion failure - count is zero.
assert(count == 10)

// Adding an observer activates the signal.
mapped.observe { _ in }
observe.send(value: 5)

// Currently: Pass
// This PR: Assertion failure - count is 5.
assert(count == 15)
```

### Use cases
* `Action`: all `map`-derived `Signal`s no longer occupy any CPU time, unless they are really being observed.

### DIFF
--- a/Sources/Bag.swift
+++ b/Sources/Bag.swift
@@ -46,9 +46,13 @@ public struct Bag<Element> {
 			}
 		}
 	}
+
+	public func index(of token: RemovalToken) -> Index? {
+		return elements.index(where: { $0.token === token })
+	}
 }
 
-extension Bag: Collection {
+extension Bag: MutableCollection {
 	public typealias Index = Array<Element>.Index
 
 	public var startIndex: Index {
@@ -60,7 +64,12 @@ extension Bag: Collection {
 	}
 
 	public subscript(index: Index) -> Element {
-		return elements[index].value
+		get {
+			return elements[index].value
+		}
+		set {
+			elements[index].value = newValue
+		}
 	}
 
 	public func index(after i: Index) -> Index {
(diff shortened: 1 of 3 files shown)


mdiep:
I'm 👎 on this.

1. It makes the `Signal` contract more complex, which is a problem for users. A simple contract is easier to understand and keep in mind.

2. It adds quite a bit of complexity to the implementation. I feel pretty strongly that we need to fight to maintain as simple an implementation as possible.

3. I expect there to be very modest gains to be made from this. I expect that unobserved `Signal`s made of stateless operators are relatively rare. I also expect that the vast majority of those that do exist don't consume much CPU time.

I don't think this meets the complexity-to-performance ratio necessary for acceptance.

andersio:
I agree that some problems meant to be addressed by this can be done less complicated with recent changes.

PR #320, opened 2017-04-11
Revert `Observer` to be a `struct`.

andersio:
1. While closures work like a reference type (in regard to their contexts), neither we rely on it with irreplaceable use cases, nor should we assume closures to have an identity as the language does not expose it.

2. Using `struct` allows the closure reference to be stored in place, instead of requiring an extra level of pointer indirection. This would reduce the memory footprint considerably by 32 bytes less per attached observer (isa, refcount, funcRef & contextRef).

### DIFF
--- a/Sources/Deprecations+Removals.swift
+++ b/Sources/Deprecations+Removals.swift
@@ -209,7 +209,7 @@ extension CompositeDisposable {
 
 extension Observer {
 	@available(*, unavailable, renamed: "init(value:failed:completed:interrupted:)")
-	public convenience init(
+	public init(
 		next: ((Value) -> Void)? = nil,
 		failed: ((Error) -> Void)? = nil,
 		completed: (() -> Void)? = nil,
--- a/Sources/Observer.swift
+++ b/Sources/Observer.swift
@@ -8,7 +8,7 @@
 
 /// An Observer is a simple wrapper around a function which can receive Events
 /// (typically from a Signal).
-public final class Observer<Value, Error: Swift.Error> {
+public struct Observer<Value, Error: Swift.Error> {
 	public typealias Action = (Event<Value, Error>) -> Void
 
 	/// An action that will be performed upon arrival of the event.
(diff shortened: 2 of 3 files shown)


mdiep:
But conceptually, observers are reference types. So I think it makes sense for them to be a `class`.

NachoSoto:
+1, semantics are explicit as a class.

andersio:
Does the reference semantics matter though? We can even replace it today with a generic typealias `Event<Value, Error> -> Void` without breaking anything semantically, only syntactically.

mdiep:
> Does the reference semantics matter though? We can even replace it today with a generic typealias Event<Value, Error> -> Void without breaking anything semantically, only syntactically.

That would still be reference semantics.

Reference semantics are not required by the `Observer` type, but it's still using reference semantics, so it makes sense to make the type itself have reference semantics. It communicates the intent more clearly to contributors and to users.

andersio:
Though my point has always been that the "reference semantic" of closure does not matter. At its core, it is just an immutable collection of captures plus a function pointer, which makes no difference from an immutable collection of references.

Yes, the _context_ is reference counted behind the scene, but so as value type collections. The common theme between these two is that the identity is not the concern, and the reference semantic is just an _optimisation_ exploring the immutability.

mdiep:
By _reference semantics_ I mean _has identity_. (cf. the object/entity/value distinction in _Domain-Driven Design_.)

A value type collection doesn't have identity, but its CoW buffer _does_. So it's consistent to use a `class` inside a `struct` in that case. But in this case, the CoW buffer is purely an implementation detail. The consumer has no knowledge of it because it's not exposed. The entirety of the public interface has value semantics.

But an `Observer` _does_ have identity—at least conceptually. I don't think of an observer as a value type that holds on to closures. I think of it as an _object_ that responds to events.

andersio:
> A value type collection doesn't have identity, but its CoW buffer does.

Not just the CoW buffer, but also the fact that it can contain references.

andersio:
Something fun happened. When I did a synthetic benchmark on this, `class`-based observers are certainly faster, likely due to less copying (one vs two pointers).

PR #323, opened 2017-04-13
Adopt nested generics.

andersio:
Nest signal and producer helper types into `Signal` and `SignalProducer`, respectively.


NachoSoto:
Can you check if this compiles with optimizations? I've seen several compiler bugs regarding nested generics with `-O`.
Unless we already run tests in release mode (which IMO we should).

andersio:
It does. Tests passed too.

mdiep:
I think we need to hold off on this until Linux can compile it.

PR #325, opened 2017-04-13
Action documentation rewrite and initializer renaming.

andersio:
The documentation rewrite in this PR attempts to explain the mechanics around the concept of:

1. Action representing a repeatable work; and
2. the `apply()` producer attempting to create and start a unit of work.

Consequently, `Action(input:_:)` has been renamed to `Action(state:_:)`, since _input_ refers to the varying input associated with execution attempts.


Sources/Action.swift
```diff
@@ -213,43 +223,47 @@ extension Action: BindingTargetProvider {
...
-	///   - input: A property whose current value is used as input
-	///            whenever the action is executed.
-	///   - execute: A closure to return a new `SignalProducer` based on the
-	///              current value of `input`.
-	public convenience init<P: PropertyProtocol, T>(input: P, _ execute: @escaping (T) -> SignalProducer<Output, Error>) where P.Value == T {
-		self.init(input: input.map(Optional.some), execute)
+	///   - state: A property to be the state of the `Action`.
+	///   - workProducer: A closure that produces a unit of work, as `SignalProducer`, to
+	///                   be executed by the `Action`.
+	public convenience init<P: PropertyProtocol, T>(state: P, _ workProducer: @escaping (T) -> SignalProducer<Output, Error>) where P.Value == T {
```
mdiep: I don't think these names are an improvement over the current names.

1. `state` seems less accurate than `input`. I wouldn't say that this property describes the _state_ of the `Action`. It's the inputs that are used—consumed in a reactive way.

2. `workProducer` is a bit misleading because it's not a producer itself—it's a closure that returns a producer. So I think `execute` is a preferable name here as well.

andersio: > I wouldn't say that this property describes the state of the `Action`. It's the inputs that are used—consumed in a reactive way.

The point is though the term _input_ is already bound to the caller-facing `apply()`, which does not take any input in this case. So calling this `input` leads to two interpretations of `input`, which do not seem nice for documentation purpose.

Moreover, it is no difference from `init(state:enabledIf:_:)` after all, just that it has a default `enabledIf` closure that asserts when the value is `nil`.

> `workProducer` is a bit misleading because it's not a producer itself—it's a closure that returns a producer. So I think `execute` is a preferable name here as well.

I can't defend `workProducer` in this regard. But `execute` doesn't seem right, since it doesn't really execute but returns something to be executed. `workFactory`, `producerFactory` or `workCustomizer` are possible alternatives.

mdiep: I'm 👍 with `state`.

> I can't defend workProducer in this regard. But execute doesn't seem right either, since it doesn't really execute but returns something to be executed. workFactory, producerFactory or workCustomizer are possible alternatives.

True, but I think most things that return `SignalProducer`s are named this way. Look at the methods in Carthage or any of its dependencies, for example.

PR #326, opened 2017-04-13
Move `timer` to `SignalProducer`.

andersio:
The very last two free functions of ReactiveSwift find their new home. Renaming fix-its work properly despite `self` being the second argument.

### DIFF
--- a/Sources/Deprecations+Removals.swift
+++ b/Sources/Deprecations+Removals.swift
@@ -1,5 +1,13 @@
 import Foundation
+import Dispatch
 import enum Result.NoError
+
+// MARK: Unavailable methods in ReactiveSwift 2.0.
+@available(*, unavailable, renamed:"SignalProducer.timer")
+public func timer(interval: DispatchTimeInterval, on scheduler: DateScheduler) -> SignalProducer<Date, NoError> { fatalError() }
+
+@available(*, unavailable, renamed:"SignalProducer.timer")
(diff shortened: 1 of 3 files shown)


andersio:
Type inference do not work if it is made as an extension to `SignalProducer`. If the concern of discovery is primarily in autocompletion, rather than documentation, these would not show up in the list.

```swift
extension SignalProducer where Value == Date, Error == NoError {
	public init(interval: DispatchTimeInterval, on scheduler: DateScheduler) {
		self = timer(interval: interval, on: scheduler, leeway: interval * 0.1)
	}
}

extension SignalProducer {
	public static func timer(interval: DispatchTimeInterval, on scheduler: DateScheduler) -> SignalProducer<Date, NoError> {
		return ReactiveSwift.timer(interval: interval, on: scheduler, leeway: interval * 0.1)
	}
}

func test() {
	let scheduler: DateScheduler! = nil
	_ = SignalProducer<Date, NoError>(interval: .seconds(1), on: scheduler)
	_ = SignalProducer<Any, NoError>.timer(interval: .seconds(1), on: scheduler)
}
```

mdiep:
This works:

```swift
extension SignalProducer where Value == Date, Error == NoError {
	public static func timer(interval: DispatchTimeInterval, on scheduler: DateScheduler) -> SignalProducer<Value, Error> {
		return ReactiveSwift.timer(interval: interval, on: scheduler, leeway: interval * 0.1)
	}
}

_ = SignalProducer.timer(interval: .seconds(5), on: QueueScheduler.main)
```

NachoSoto:
👌🏻

PR #337, opened 2017-04-20
Rename closure parameter names of operators, and update documentations.

andersio:
1. Follow the Swift API Guidelines and the Standard Library in naming closure parameters according to their assertions or roles, instead of generic names like `predicate`. e.g.
    ```
    -	public func filter(_ predicate: @escaping (Value) -> Bool) -> Signal<Value, Error> {
    +	public func filter(_ isIncluded: @escaping (Value) -> Bool) -> Signal<Value, Error> {
    ```

1. Remove the label of a `SignalProducer.attempt` overload, which is supposed to be a parameter name.

1. Condense a few pieces of documentations by taking advantage of the new parameter names, while keeping another few consistent among the `Signal` and the lifted versions.

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -425,7 +425,7 @@ extension Signal: SignalProtocol {
 }
 
 extension Signal {
-	/// Attach a closure as an observer to all events from `self`.
+	/// Observe `self` for all events being emitted.
 	///
 	/// - note: If `self` has terminated, the closure would be invoked with an
 	///         `interrupted` event immediately.
@@ -440,8 +440,7 @@ extension Signal {
 		return observe(Observer(action))
 	}
 
-	/// Attach a closure as an observer to the `value` events and the `failed` event from
-	/// `self`.
+	/// Observe `self` for all values being emitted, and if any, the failure.
 	///
 	/// - parameters:
 	///   - action: A closure to be invoked with values from `self`, or the propagated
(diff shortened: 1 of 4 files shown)


Sources/Signal.swift
```diff
@@ -1459,37 +1442,33 @@ extension Signal {
 }
 
 extension Signal where Value: Equatable {
-	/// Forward only those values from `self` which are not duplicates of the
-	/// immedately preceding value. 
+	/// Forward only values from `self` that are not equal to its consecutive predecessor.
```
mdiep: I think _the immediately preceding value_ is clearer here. I'm not sure _consecutive predecessor_ makes sense.

Sources/Signal.swift
```diff
@@ -2360,69 +2338,63 @@ extension Signal {
 }
 
 extension Signal where Error == NoError {
-	/// Apply a failable `operation` to values from `self` with successful
-	/// results forwarded on the returned signal and thrown errors sent as
-	/// failed events.
+	/// Invoke a throwable action with every value from `self`, and forward the values
```
mdiep: I think _Apply a failable `operation`_ is clearer here.

_Invoke a throwable action_ sounds way too imperative.

c.f. the description of `map` from Swift:

> Returns an array containing the results of mapping the given closure over the sequence's elements.

Sources/Signal.swift
```diff
@@ -1404,15 +1388,17 @@ extension Signal {
 	}
 
 
-	/// Send only the final value and then immediately completes.
+	/// Accumuate all values from `self`, and forward only the final accumuated result.
```
mdiep: I don't think `accumulate` (typoed) is the best description. The return value is the accumulator, but I don't think _accumulate_ describes the operation itself.

C.f. the Swift `reduce` description:

> Returns the result of combining the elements of the sequence using the given closure

PR #339, opened 2017-04-20
Rename `flatMap(_:transform:)` to `flatMap(_:_:)`.

andersio:
While the API guidelines says all argument labels should be named unless otherwise specified, the transforms of FP-borrowed operators usually do not have a label given the "terms of art" treatment. So the PR proposes to strip the label from `flatMap`.


mdiep:
Something like `return self.flatMap(.merge, SignalProducer.init)` seems a little weird to me, but I think I'm 👍 on this. 😕 

On the one hand, Swift's `flatMap` has an unnamed argument (and so does our `flatMapError`). On the other hand, this `flatMap` has 2 arguments.

Since I don't think this is obviously the better option, I'd like to give @ReactiveCocoa/reactiveswift a chance to speak up in case anyone has an objection.

mdiep:
No objects, so I think this is good to go. Want to fix up the conflicts?

I'm thinking that we should really make our deprecations file based on the version. `So Deprecations+Removals-1.swift`, `Deprecations+Removals-2.swift`, etc.

andersio:
Fixed.

PR #340, opened 2017-04-21
Streamlined the mechanics behind composed properties.

andersio:
1. Replaced `replayLazily(upTo:)` with a custom implementation.

    Composed properties now cache only one copy of the latest value, down from 1-2 copies currently. Moreover, it now stops caching after both the `Property` and all the producer instances has deinitialised.


Sources/Property.swift
```diff
@@ -687,14 +687,34 @@ public final class MutableProperty<Value>: ComposableMutablePropertyProtocol {
...
+/// The requirement of a `Value?` storage from composed properties prevents further
+/// implementation sharing with `MutableProperty`.
+private final class PropertyBox<Value> {
+	let lock = NSRecursiveLock()
+	var value: Value
+
+	init(_ value: Value) {
+		self.value = value
+	}
+}
```
mdiep: I'd like to see this use the same approach as `Atomic` to prevent unsafe accesses—hiding `value` and `lock` behind locking APIs.

Sources/Property.swift
```diff
@@ -687,14 +687,34 @@ public final class MutableProperty<Value>: ComposableMutablePropertyProtocol {
...
+
+extension NSRecursiveLock {
+	fileprivate func sync<Result>(_ action: () throws -> Result) rethrows -> Result {
```
mdiep: I think this can go away of appropriate methods are added to `PropertyBox`.

Sources/Property.swift
```diff
@@ -700,18 +701,21 @@ public final class MutableProperty<Value>: ComposableMutablePropertyProtocol {
...
+	func modify<Result>(_ action: (inout Value) throws -> Result) rethrows -> Result {
+		return try modifyDidSet(action, didSet: nil)
 	}
-}
 
-extension NSRecursiveLock {
-	fileprivate func sync<Result>(_ action: () throws -> Result) rethrows -> Result {
-		lock(); defer { unlock() }
-		return try action()
+	func modifyDidSet<Result>(_ action: (inout Value) throws -> Result, didSet: (() -> Void)?) rethrows -> Result {
```
mdiep: Just make this:

```swift
func modify<Result>(
    _ action: (inout Value) throws -> Result,
    didSet (() -> Void)? = nil
) rethrows -> Result {
```

?

Sources/Property.swift
```diff
@@ -571,27 +571,28 @@ public final class Property<Value>: PropertyProtocol {
...
 
-				box.lock.sync {
-					if let value = event.value {
-						box.value = value
+				box.modifyDidSet({ value in
+					if let newValue = event.value {
+						value = newValue
 					}
+				}, didSet: {
 					observer.action(event)
```
mdiep: Can't `observer.action` be in the `action` block? It still happens inside the lock either way.

Sources/Property.swift
andersio: Ah, I thought the trailing closure would be in conflict. But I just realize `didSet` can be moved as the first. 😛

Sources/Property.swift
andersio: It violates [access exclusivity](https://github.com/apple/swift-evolution/blob/master/proposals/0176-enforce-exclusive-access-to-memory.md) and the default copy-in, copy-out semantic of `inout`, which updates `_value` only upon the return of the closure. In-place mutation is an optimisation on top of these two assumptions.

andersio: Ah, `MutableProperty.modify` needs to be changed too apparently.

PR #342, opened 2017-04-22
Unfair Lock, take 2

andersio:
As mentioned in #282, `os_unfair_lock` is generally faster all-around (slightly when uncontended, significantly when contended), since it does not need to maintain fairness among waiting threads.

~~This branch is a transitional implementation purely in Swift, until eventually Swift supports availability propagation of enum cases to switch case scopes.~~

~~Related: [SR-4079](https://bugs.swift.org/browse/SR-4079)~~

`Lock` is an abstract base class of two implementations: `UnfairLock` and `PthreadLock`. When constructing a lock through `Lock.Type.make`, it picks the implementation with `#available` runtime checks.

It turns out that virtual dispatch (subclassing) is still pretty fast, and is certainly faster than the implementation of the #282 trick in pure Swift, which requires per-instance thunks.

[Benchmark Source](https://gist.github.com/andersio/c4e7c2425ce2725cdf63edc3b032d024).
```
5_000_000 iterations
@Lock: 40.4060875994435 ns
@NSLock: 67.0863552018613 ns
@PthreadLock: 47.6224342023663 ns
@UnfairLock: 40.1545092006554 ns
```
Existentials are rather slow, by the way.

Inspecting the generated object code with `-Owmo` reveals that Swift has inlined the vtable. So I guess even if [SR-4079](https://bugs.swift.org/browse/SR-4079) got fixed someday, it would hardly be faster than `Lock`.

```
    0x1052c5566 <+38>:   cmpq   0x7bacb(%rip), %rax       ; (void *)0x0000000105350690: type metadata for ReactiveSwift.Lock
    0x1052c556d <+45>:   je     0x1052c58dc               ; <+924> [inlined] function signature specialization <Arg[0] = Dead> of ReactiveSwift.Lock.lock () -> () at Signal.swift:189
    0x1052c5573 <+51>:   cmpq   0x7bac6(%rip), %rax       ; (void *)0x0000000105350738: type metadata for ReactiveSwift.Lock.UnfairLock
    0x1052c557a <+58>:   je     0x1052c558a               ; <+74> [inlined] ReactiveSwift.Lock.UnfairLock.lock () -> () at Signal.swift:302
    0x1052c557c <+60>:   cmpq   0x7bac5(%rip), %rax       ; (void *)0x00000001053507f0: type metadata for ReactiveSwift.Lock.PthreadLock
    0x1052c5583 <+67>:   je     0x1052c559c               ; <+92> [inlined] ReactiveSwift.Lock.PthreadLock.lock () -> () at Signal.swift:302
    0x1052c5585 <+69>:   callq  *0x58(%rax)
    0x1052c5588 <+72>:   jmp    0x1052c55ac               ; <+108> at Signal.swift:303
    0x1052c558a <+74>:   incq   0x8e077(%rip)             ; __profc_/Users/anders/Repo/RAS/Sources/Scheduler.swift:_TFCC13ReactiveSwift13TestSchedulerP33_B7AFF4CC5431FDC431F4B042BAEABD8F15ScheduledActionCfT4dateV10Foundation4Date6actionFT_T__S1_ + 5256
->  0x1052c5591 <+81>:   addq   $0x10, %rdi
    0x1052c5595 <+85>:   callq  0x105329fa2               ; symbol stub for: os_unfair_lock_lock
```

### DIFF
--- a/Sources/Atomic.swift
+++ b/Sources/Atomic.swift
@@ -104,33 +104,97 @@ internal struct UnsafeAtomicState<State: RawRepresentable> where State.RawValue 
 #endif
 }
 
-final class PosixThreadMutex: NSLocking {
-	private var mutex = pthread_mutex_t()
+/// `Lock` exposes `os_unfair_lock` on supported platforms, with pthread mutex as the
+// fallback.
+internal class Lock {
+	#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
+	@available(iOS 10.0, *)
+	@available(macOS 10.12, *)
+	@available(tvOS 10.0, *)
+	@available(watchOS 3.0, *)
+	internal final class UnfairLock: Lock {
+		private var _lock: os_unfair_lock
 
-	init() {
-		let result = pthread_mutex_init(&mutex, nil)
-		precondition(result == 0, "Failed to initialize mutex with error \(result).")
+		override init() {
+			_lock = os_unfair_lock()
+			super.init()
+		}
+
+		override func lock() {
+			withUnsafeMutablePointer(to: &_lock, os_unfair_lock_lock)
+		}
+
+		override func unlock() {
+			withUnsafeMutablePointer(to: &_lock, os_unfair_lock_unlock)
+		}
+
+		override func `try`() -> Bool {
+			return withUnsafeMutablePointer(to: &_lock, os_unfair_lock_trylock)
+		}
+	}
+	#endif
+
+	internal final class PthreadLock: Lock {
+		private var _lock: pthread_mutex_t
+
+		override init() {
+			_lock = pthread_mutex_t()
+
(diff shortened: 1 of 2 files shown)


Sources/Atomic.swift
```diff
@@ -104,33 +104,97 @@ internal struct UnsafeAtomicState<State: RawRepresentable> where State.RawValue
 #endif
 }
 
-final class PosixThreadMutex: NSLocking {
-	private var mutex = pthread_mutex_t()
+/// `Lock` exposes `os_unfair_lock` on supported platforms, with pthread mutex as the
+// fallback.
+internal class Lock {
+	#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
```
sharplet: It's a bummer that this uses both `#if os` and `@available` :(

PR #343, opened 2017-04-23
Move `SignalProducerProtocol.then` to the concrete type.

andersio:
Follow-up to #304.

`then(_:)` is disambiguated by introducing two more specific variants.

### DIFF
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -1733,7 +1733,23 @@ extension SignalProducer {
 	/// Wait for completion of `self`, *then* forward all events from
 	/// `replacement`. Any failure or interruption sent from `self` is
 	/// forwarded immediately, in which case `replacement` will not be started,
-	/// and none of its events will be be forwarded. 
+	/// and none of its events will be be forwarded.
+	///
+	/// - note: All values sent from `self` are ignored.
+	///
+	/// - parameters:
+	///   - replacement: A producer to start when `self` completes.
+	///
+	/// - returns: A producer that sends events from `self` and then from
+	///            `replacement` when `self` completes.
+	public func then<U>(_ replacement: SignalProducer<U, NoError>) -> SignalProducer<U, Error> {
+		return _then(replacement.promoteErrors(Error.self))
+	}
+
+	/// Wait for completion of `self`, *then* forward all events from
+	/// `replacement`. Any failure or interruption sent from `self` is
+	/// forwarded immediately, in which case `replacement` will not be started,
+	/// and none of its events will be be forwarded.
 	///
 	/// - note: All values sent from `self` are ignored.
 	///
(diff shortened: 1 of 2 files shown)


Sources/SignalProducer.swift
```diff
@@ -1743,6 +1760,10 @@ extension SignalProducer {
 	/// - returns: A producer that sends events from `self` and then from
 	///            `replacement` when `self` completes.
 	public func then<U>(_ replacement: SignalProducer<U, Error>) -> SignalProducer<U, Error> {
+		return _then(replacement)
+	}
+
+	internal func _then<U>(_ replacement: SignalProducer<U, Error>) -> SignalProducer<U, Error> {
```
NachoSoto: Why have this?

Tests/ReactiveSwiftTests/SignalProducerSpec.swift
```diff
@@ -2696,6 +2721,6 @@ extension SignalProducer {
 			}
 		}
 
-		return SignalProducer.attempt(operation)
+		return SignalProducer(operation)
```
NachoSoto: Maybe add `attempt` as a label? Having so many unnamed initializers leads to ambiguity.

Sources/SignalProducer.swift
andersio: Not having this leads to infinite self recursion, given that Swift selects the most specific overload for a given name.

Tests/ReactiveSwiftTests/SignalProducerSpec.swift
andersio: [A test case](https://github.com/ReactiveCocoa/ReactiveSwift/pull/343/files#diff-9cd786f9e52e786771d96cf8c42d3befR207) was added specifically for this, and this set of overloads doesn't appear to be ambiguous.

Moreover, it is expected to be called with trailing closures anyway. If it is ambiguous as trailing closures, adding labels do not help.

Sources/SignalProducer.swift
mdiep: Can you add a comment that explains that?

Sources/SignalProducer.swift
```diff
@@ -1733,7 +1718,39 @@ extension SignalProducer {
...
+	/// and none of its events will be be forwarded.
+	///
+	/// - note: All values sent from `self` are ignored.
+	///
+	/// - parameters:
+	///   - replacement: A producer to start when `self` completes.
+	///
+	/// - returns: A producer that sends events from `self` and then from
+	///            `replacement` when `self` completes.
+	public func then(_ replacement: SignalProducer<Value, Error>) -> SignalProducer<Value, Error> {
```
mdiep: This was added to provide a more specific variant for type inference? If so, can you add a `//` comment that documents that?

Is this inference covered by a test case?

Tests/ReactiveSwiftTests/SignalProducerSpec.swift
mdiep: I prefer the `static func attempt` name for that reason. It also lends itself to RACformatting.

```swift
SignalProducer
    .attempt {
        …
    }
    .flatMap(.concat) {
        …
    }
```

Is renaming this an essential part of the PR? If not, could we move it off the protocol here and discuss the renaming in a separate PR?

andersio: It is. I will open a separate PR for it then.

> This coincidentally dodged the return type ambiguity of `SignalProducer.Type.attempt`, which somehow managed to get through as a protocol extension.

andersio:
Addressed all comments.

PR #345, opened 2017-04-23
Streamline high-arity `combineLatest` and `zip`.

### LINKED ISSUE #156: Stack overflow with super-high-arity `combineLatest` and `zip`.
Related Bug: https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2384

#### Possible Solution
Flatten `combineLatest` and `zip` implementations using gyb, which would also allow lock sharing.

andersio:
Preceded by https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2521.
Fixes #156.

1. One lock and one `Signal` regardless of the arity or the number of signals.
1. Stack friendly.
1. Human maintainable IMO.

| Test item, 50000 iterations | `master` | This PR | Delta | Speedup |
| ---- | ---- | ---- | --- | --- |
| Tenary `combineLatest`, `Signal` | 0.370 sec | 0.205 sec | -18% | 1.80x |
| Ternary `combineLatest`, `SignalProducer` | 0.589 sec | 0.313 sec | -47% | 1.88x |
| 6-ary `combineLatest`, `Signal` | 1.733 sec | 0.440 sec | -75% | 3.94x |
| 6-ary `combineLatest`, `SignalProducer` | 2.541 sec | 0.651 sec | -74% | 3.90x |
| Blowing Main Stack (1 iteration) | 4950 | 7450 | N/A | N/A |

### Note
`combineLatest` and `zip` adds signals in a chain to a `Singal.AggregateBuilder`, which constructs a `Signal` when the chain ends with `make()`. `Singal.AggregateBuilder` encapsulates all the type erasure & recovery.

In https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2521, a primary dissent was that it introduced risks that the compiler cannot statically validate. This PR does not, and cannot mitigate, such risk with the performance goal due to the lack of variadic generics.

Note: Xcode 8.3.2, i5 6360U, Whole Module Optimisation. [Benchmark Source (Gist)](https://gist.github.com/andersio/5eaef64adccc4af9f687abc8894cc5de).


Sources/Signal.swift
```diff
@@ -2015,177 +1876,352 @@ private enum ThrottleWhileState<Value> {
...
+		private var isInitialized: Bool {
+			mutating get {
+				if _isInitialized {
+					return true
+				}
+
+				_isInitialized = values.reduce(true, { $0 && !($1 is Placeholder) })
+				return _isInitialized
+			}
+		}
```
mdiep: I don't love using the term _initialized_. How about flipping the meaning and naming it `hasEmptySignal`?

Sources/Signal.swift
```diff
@@ -2015,177 +1876,352 @@ private enum ThrottleWhileState<Value> {
...
+	init(count: Int, action: @escaping (ContiguousArray<Any>) -> Void)
+}
+
 extension Signal {
+	private struct CombineLatestStrategy: SignalAggregateStrategy {
+		private struct Placeholder {}
+
+		private var values: ContiguousArray<Any>
+		private var completionCount: Int
+		private var _isInitialized: Bool
```
mdiep: It's very minor, but in cases like this, I'd put the cached property immediately before the uncached version.

```swift
private let action: (ContiguousArray<Any>) -> Void

private var _isInitialized: Bool
private var isInitialized: Bool { … }
```

Sources/Signal.swift
```diff
@@ -2015,177 +1876,352 @@ private enum ThrottleWhileState<Value> {
...
+		private let action: (ContiguousArray<Any>) -> Void
+
+		private var isInitialized: Bool {
+			mutating get {
+				if _isInitialized {
+					return true
+				}
+
+				_isInitialized = values.reduce(true, { $0 && !($1 is Placeholder) })
```
mdiep: Trailing closure syntax?

Sources/Signal.swift
```diff
@@ -2015,177 +1876,352 @@ private enum ThrottleWhileState<Value> {
...
+	mutating func complete(at position: Int) -> Bool
+
+	init(count: Int, action: @escaping (ContiguousArray<Any>) -> Void)
+}
+
 extension Signal {
+	private struct CombineLatestStrategy: SignalAggregateStrategy {
+		private struct Placeholder {}
+
+		private var values: ContiguousArray<Any>
```
mdiep: Would it be better to use an `Optional`?

```swift
private var values: ContiguousArray<Any?>
```

Sources/Signal.swift
```diff
@@ -2015,177 +1876,352 @@ private enum ThrottleWhileState<Value> {
...
+
+	private struct ZipStrategy: SignalAggregateStrategy {
+		private var values: ContiguousArray<[Any]>
+		private var isCompleted: ContiguousArray<Bool>
+		private let action: (ContiguousArray<Any>) -> Void
+
+		mutating func update(_ value: Any, at position: Int) -> Bool {
+			values[position].append(value)
+
+			if values.reduce(true, { $0 && !$1.isEmpty }) {
```
mdiep: Maybe add a computed property for this to aid readability? Or add a `mutating func` that returns a `ContiguousArray<Any>?` of the values to send?

Sources/Signal.swift
```diff
@@ -2015,177 +1876,352 @@ private enum ThrottleWhileState<Value> {
...
+				var buffer = ContiguousArray<Any>()
+				buffer.reserveCapacity(values.count)
+
+				for index in values.startIndex ..< values.endIndex {
```
mdiep: `values.indices`?

Sources/Signal.swift
```diff
@@ -2015,177 +1876,352 @@ private enum ThrottleWhileState<Value> {
...
+					buffer.append(values[index].removeFirst())
+				}
+
+				action(buffer)
+
+				if Swift.zip(values, isCompleted).contains(where: { $0.isEmpty && $1 }) {
```
mdiep: This conditional is repeated in `complete` below. I think it'd be good to make this into a computed property.

Sources/Signal.swift
```diff
@@ -2015,177 +1876,352 @@ private enum ThrottleWhileState<Value> {
...
+	private final class AggregateBuilder<Strategy: SignalAggregateStrategy> {
+		private var startHandlers: [(Int, CompositeDisposable, Atomic<Strategy>, @escaping (Event<Never, Error>) -> Void) -> Void]?
+
+		init() {
+			self.startHandlers = []
+		}
+
+		@discardableResult
+		func add<U>(_ signal: Signal<U, Error>) -> Self {
+			precondition(startHandlers != nil, "Signals cannot be registered after the state has started.")
```
mdiep: Can we do this as part of `init` instead?

I think each `make` below could become an `init` that calls a top-level `make` with the signals that it wants to start.

Sources/Signal.swift
```diff
@@ -2015,177 +1876,352 @@ private enum ThrottleWhileState<Value> {
...
+
+		init(count: Int, action: @escaping (ContiguousArray<Any>) -> Void) {
+			values = ContiguousArray(repeating: [], count: count)
+			isCompleted = ContiguousArray(repeating: false, count: count)
+			self.action = action
+		}
+	}
+
+	private final class AggregateBuilder<Strategy: SignalAggregateStrategy> {
+		private var startHandlers: [(Int, CompositeDisposable, Atomic<Strategy>, @escaping (Event<Never, Error>) -> Void) -> Void]?
```
mdiep: This seems like it would benefit from a `struct`. It's a bit large as a type.

Sources/Signal.swift
```diff
@@ -2015,177 +1876,352 @@ private enum ThrottleWhileState<Value> {
...
+					case let .failed(error):
+						action(.failed(error))
+					}
+				}
+			}
+
+			return self
+		}
+
+		private func make<Composite>(count: Int, _ transform: @escaping (ContiguousArray<Any>) -> Composite) -> Signal<Composite, Error> {
```
NachoSoto: Yikes :( it makes me sad that we lose type information here.

mdiep: Yeah. 😞 But until the type system is powerful enough, I don't mind some internal type erasure.

Sources/Signal.swift
andersio: The shared aggregator interface uses `(ContiguousArray<Any>) -> Void`, since zip does not need the optionality.

Sources/Signal.swift
andersio: `values.indices` holds a strong reference to the collection, and is not recommended if you gonna do in-place mutation.

andersio:
@mdiep https://github.com/ReactiveCocoa/ReactiveSwift/pull/345/commits/6ee1fe723f77e0f3c4978585c172ead1ef50045b probably still doesn't do what you want. But I don't think nested calls would make it any more manageable.

andersio:
This seems working, but we would have to pull back `TupleExtensions.swift`. It also takes `2n - 3` calls for a n-arity operator to pack its results.

```swift
struct Builder<Composite, E> {
    func append<U, NewComposite>(_ signal: S<U, E>, _ pack: @escaping (Composite, Any) -> NewComposite) -> Builder<NewComposite, E> {
        let transform = self.transform
        let newTransform = { pack(firstTransform($0), $0[0]) }
    }
}

let _: Builder<(A, B, C, D, E), E> = Builder<(A, B), E>(a, b)
    .append(c, repack)
    .append(d, repack)
    .append(e, repack)
```

mdiep:
This is what I was thinking:

```swift
extension Signal {
  func erased() -> Signal<Any, Error> { … }
}

class Combiner<Strategy> {
  var startHandlers: […]

  init<Composite>(
    signals: [Signal<Any, Error>],
    _ transform: @escaping (ContiguousArray<Any>) -> Composite
  ) -> Signal<Composite, Error> {
    // the code that was in make(count:_:), but also the code from add(_:)
  }

  convenience init<A, B>(
    _ a: Signal<A, Error>,
    _ b: Signal<B, Error>
  ) -> Signal<(A, B>), Error> {
    self.init([a.erased(), b.erased()]) { ($0[0] as! A, $0[1] as! B) }
  }

  …
}
```

andersio:
I feel like https://github.com/ReactiveCocoa/ReactiveSwift/pull/345/commits/6ee1fe723f77e0f3c4978585c172ead1ef50045b is pretty close, except for not needing an erased version of signals.

Sources/SignalProducer.swift
```diff
@@ -2200,3 +2212,66 @@ extension SignalProducer where Value == Date, Error == NoError {
...
+extension SignalProducer {
+	fileprivate func startWithSignal(interruptingBy disposable: CompositeDisposable, setup: (Signal<Value, Error>) -> Void) {
+		startWithSignal { signal, interruptHandle in
+			disposable += interruptHandle
+			setup(signal)
+		}
+	}
+}
+
+private func flattenStart<A, B, Error>(_ disposable: CompositeDisposable, _ a: SignalProducer<A, Error>, _ b: SignalProducer<B, Error>, _ start: (Signal<A, Error>, Signal<B, Error>) -> Void) {
```
mdiep: Can you add some documentation to these?

Sources/SignalProducer.swift
```diff
@@ -2200,3 +2212,66 @@ extension SignalProducer where Value == Date, Error == NoError {
 		}
 	}
 }
+
+extension SignalProducer {
+	fileprivate func startWithSignal(interruptingBy disposable: CompositeDisposable, setup: (Signal<Value, Error>) -> Void) {
```
mdiep: Can you add some documentation for this?

Sources/SignalProducer.swift
mdiep: Could these be structured after `lift` instead?

Sources/Signal.swift
```diff
@@ -2015,177 +1876,346 @@ private enum ThrottleWhileState<Value> {
...
+		private enum Placeholder {
+			case none
+		}
+
+		private var values: ContiguousArray<Any>
+		private var completionCount: Int
+		private let action: (ContiguousArray<Any>) -> Void
+
+		private var _hasAllSentInitial: Bool
+		private var hasAllSentInitial: Bool {
```
mdiep: This reads a little funny. Maybe rename to `haveAllSentInitial` or `hasAllInitialValues`?

Sources/Signal.swift
```diff
@@ -2015,177 +1876,346 @@ private enum ThrottleWhileState<Value> {
...
+	}
+
+	private struct ZipStrategy: SignalAggregateStrategy {
+		private var values: ContiguousArray<[Any]>
+		private var isCompleted: ContiguousArray<Bool>
+		private let action: (ContiguousArray<Any>) -> Void
+
+		private var hasCompletedAndEmptiedSignal: Bool { return Swift.zip(values, isCompleted).contains(where: { $0.isEmpty && $1 }) }
+		private var canEmit: Bool { return values.reduce(true) { $0 && !$1.isEmpty } }
+		private var isAllCompleted: Bool { return isCompleted.reduce(true) { $0 && $1 } }
```
mdiep: Maybe rename this to `areAllCompleted` so it reads nicely in English?

Sources/Signal.swift
```diff
@@ -2015,177 +1876,346 @@ private enum ThrottleWhileState<Value> {
...
+			self.action = action
+		}
+	}
+
+	private struct ZipStrategy: SignalAggregateStrategy {
+		private var values: ContiguousArray<[Any]>
+		private var isCompleted: ContiguousArray<Bool>
+		private let action: (ContiguousArray<Any>) -> Void
+
+		private var hasCompletedAndEmptiedSignal: Bool { return Swift.zip(values, isCompleted).contains(where: { $0.isEmpty && $1 }) }
```
mdiep: Would you mind putting these over multiple lines? I think that makes them more readable.

Sources/SignalProducer.swift
andersio: What do you mean by structured after `lift`?

PR #346, opened 2017-04-24
Rename `SignalProducer.Type.attempt` to `init(_:)`.

andersio:
Follow-up to #304.

The original signatures are ambiguous, but they somehow work as protocol extensions in Swift 3. Moving them to the concrete type reveals the ambiguity.
```swift
extension SignalProducer {
	public static func attempt(_ operation: @escaping () -> Result<Value, Error>) -> SignalProducer<Value, Error>
}

extension SignalProducer where Error == AnyError {
	public static func attempt(_ operation: @escaping () throws -> Value) -> SignalProducer<Value, Error>
}
```

Consider `.attempt { () -> Result<Int, NoError> in .success(1) }`, the closure can satisfy both methods and produce:
1. `SignalProducer<Int, NoError>`; or
1. `SignalProducer<Result<Int, NoError>, AnyError>` since non-throwing closures are covariant of their throwing version.

### Solution
Renaming `SignalProducer.Type.attempt` to `init(_:)` mitigated the ambiguity.  How this happens is uncertain, given that Swift has no documentation on the overload selection rules at the moment.

Edit: It is suspected that these as static functions require the compiler to consider two additional generic parameters in the return type. Being `init(_:)`, however, binds the return type to be `Self`, and reduce the overload selection factors to the arguments.

### DIFF
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -82,6 +82,26 @@ public struct SignalProducer<Value, Error: Swift.Error> {
 		}
 	}
 
+	/// Create a `SignalProducer` that will attempt the given operation once for
+	/// each invocation of `start()`.
+	///
+	/// Upon success, the started signal will send the resulting value then
+	/// complete. Upon failure, the started signal will fail with the error that
+	/// occurred.
+	///
+	/// - parameters:
+	///   - action: A closure that returns instance of `Result`.
+	public init(_ action: @escaping () -> Result<Value, Error>) {
+		self.init { observer, disposable in
+			action().analysis(ifSuccess: { value in
+				observer.send(value: value)
+				observer.sendCompleted()
+			}, ifFailure: { error in
+				observer.send(error: error)
+			})
+		}
+	}
+
 	/// Creates a producer for a `Signal` that will immediately fail with the
 	/// given error.
 	///
@@ -180,6 +200,25 @@ public struct SignalProducer<Value, Error: Swift.Error> {
 		}
 
 		startHandler(observer, producerDisposable)
+	}
+}
+
+extension SignalProducer where Error == AnyError {
+	/// Create a `SignalProducer` that will attempt the given failable operation once for
+	/// each invocation of `start()`.
+	///
+	/// Upon success, the started producer will send the resulting value then
+	/// complete. Upon failure, the started signal will fail with the error that
+	/// occurred.
+	///
+	/// - parameters:
+	///   - operation: A failable closure.
+	public init(_ action: @escaping () throws -> Value) {
+		self.init {
+			return ReactiveSwift.materialize {
+				return try action()
+			}
+		}
 	}
 }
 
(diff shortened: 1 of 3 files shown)


mdiep:
Does this ambiguity cause problems in practice?

It's ambiguous for our test suite because we do this:

```swift
let operation: () -> Result<String, NSError> = {
	operationRunTimes += 1

	return .success("OperationValue")
}

SignalProducer.attempt(operation).start()
SignalProducer.attempt(operation).start()
```

But I think that's understandable.

Ordinarily, you wouldn't use a `SignalProducer` like that—you'd return it from some function which will add additional type information.

This is okay, for instance:

```swift
let operation: () -> Result<String, NSError> = {
	operationRunTimes += 1

	return .success("OperationValue")
}

func makeProducer() -> SignalProducer<String, NSError> {
	return .attempt(operation)
}

makeProducer().start()
makeProducer().start()
```

andersio:
The type inferrer works in this context.
```swift
func attempt(_ operation: @escaping () -> Result<String, NSError>) -> SignalProducer<(), NSError> {
	return SignalProducer.attempt(operation)
		.on(started: { started += 1 })
		.map { _ in }
}
```

But not if the `Result` has a type of `AnyError`:
```swift
func attempt2(_ operation: @escaping () -> Result<String, AnyError>) -> SignalProducer<(), AnyError> {
	return SignalProducer.attempt(operation)
		.on(started: { started += 1 })
		.map { _ in () }
}
```

mdiep:
> But not if the Result has a type of AnyError

I think we could add a 3rd overload to disambiguate that.

```swift
extension SignalProducerProtocol where Error == AnyError {
	public static func attempt(_ operation: @escaping () -> Result<Value, Error>) -> SignalProducer<Value, Error>
}
```

That seems like the same thing we do for `flatMap`.

andersio:
The problem is that the type checker can infer `Value` to be `Result<Value, Error>`, or `Value` inside the `Result`. Both are valid and cannot be disambiguated by giving more specific constraints.

andersio:
As mentioned on Slack, there is a workaround for `attempt` to work:
```swift
public protocol _AnyError {}
extension AnyError: _AnyError {}
extension SignalProducer where Value: _AnyError { ... }
```

Sources/SignalProducer.swift
```diff
@@ -185,6 +205,25 @@ public struct SignalProducer<Value, Error: Swift.Error> {
...
+	///
+	/// Upon success, the started producer will send the resulting value then
+	/// complete. Upon failure, the started signal will fail with the error that
+	/// occurred.
+	///
+	/// - parameters:
+	///   - operation: A failable closure.
+	public init(_ action: @escaping () throws -> Value) {
+		self.init {
+			return ReactiveSwift.materialize {
```
mdiep: I don't think we actually need this `materialize` alias. I think we can do `Result(attempt: try action())`.

ikesyo:
I've just experienced the weirdness due to this and `attemptMap` overloads: https://github.com/mdiep/Tentacle/pull/86#issuecomment-317451932.

ikesyo:
FYI I suffered this again in Carthage: https://github.com/Carthage/Carthage/pull/2077.

PR #350, opened 2017-04-25
[Gardening] Use a manual allocation for `Signal.state`.

andersio:


RuiAAPeres:
Why?

andersio:
Swift does not define a memory model yet, and does not guarantee C memory semantics (except for unsafe memory primitives that guarantee interoperability with C). So these are kind of a precaution.

```
// The storage of the state.
//
// Since Swift stored property accesses do not guarantee atomicity due to possible
// compiler-generated conversions and reabstractions, a manual allocation via
// `UnsafeMutablePointer` is made to ensure C-like memory semantics.
//
// Related: https://lists.swift.org/pipermail/swift-users/Week-of-Mon-20161205/004147.html
```

andersio:
(Paranoid mode: OFF)

Just checked the generated source of debug builds in Swift 3.1. It does generate a pointer size store with the packed pointer, and is probably safe from changes given that enums with only pointer payloads are given a special treatment. So apparently this is unnecessary.

Issue #349, opened 2017-04-25
Improve producer operator interoperability with a `ExpressibleBySignalProducer` protocol.


andersio:
Introduce a `ExpressibleBySignalProducer` protocol, and constrain the argument of suitable binary and n-ary operators to it. This is a source compatible binary incompatible change.

`Signal`, `SignalProducer` and `PropertyProtocol` would all conform to `SignalProducerRepresentable`, and returns its representation as a `SignalProducer` as requested. This allows all producer operators to accept any reactive primitives without defining explicit overloads.

e.g.
```swift
extension SignalProducer {
    public func combineLatest<Other>(with: Other) -> SignalProducer<(Value, Other.Value), Error>
    where Other: ExpressibleBySignalProducer, Other.Error == Error
} 

// All valid.
producer.combineLatest(with: signal)
producer.combineLatest(with: producer)
producer.combineLatest(with: property)

// Mix and match.
SignalProducer.combineLatest(signal, producer, property)
SignalProducer.combineLatest(property, producer, signal)
```

Additionally, new overloads may be added to improve the `NoError` usability.
```swift
extension SignalProducer {
    public func combineLatest<Other>(with: Other) -> SignalProducer<(Value, Other.Value), Error>
    where Other: SignalProducerConvertible, Other.Error == NoError
}

extension SignalProducer where Error == NoError {
    public func combineLatest<Other>(with: Other) -> SignalProducer<(Value, Other.Value), Other.Error>
    where Other: SignalProducerConvertible
}
```

This may also reduce the number of flattening operator overloads, and may allow the removal of `SignalProtocol` and `SignalProducerProtocol`.

### Note
`ExpressibleBySignal` and `ExpressibleByProperty` were considered, but:

1. Property is the most restricting contract in ReactiveSwift, so practically the only `ExpressibleByProperty` entities are properties, rendering the protocol useless.

1. `ExpressibleBySignal` can ideally cover both producers and properties. But doing so is equivalent to lossy type narrowing, since the events emitted during the producer starting side effect would be lost.

mdiep:
A small PR that demonstrates this with a single operator would probably helpful in evaluating this proposal.

sharplet:
Left a comment on https://github.com/ReactiveCocoa/ReactiveSwift/pull/351#issuecomment-304308268 about naming.

andersio:
Implemented in #351.

PR #351, opened 2017-04-25
Generic n-ary producer operators for n > 1.

andersio:
Implements #349.

Introduce `SignalProducerConvertible` so that n-ary producer operators may take any entity which is declared to be expressible by a `SignalProducer`.

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -519,22 +519,9 @@ extension Signal {
 	}
 }
 
-/// A protocol used to constraint `Signal` operators.
-public protocol SignalProtocol {
-	/// The type of values being sent on the signal.
-	associatedtype Value
-
-	/// The type of error that can occur on the signal. If errors aren't
-	/// possible then `NoError` can be used.
-	associatedtype Error: Swift.Error
-
-	/// Extracts a signal from the receiver.
-	var signal: Signal<Value, Error> { get }
-}
-
-extension Signal: SignalProtocol {
-	public var signal: Signal {
-		return self
+extension Signal: SignalProducerConvertible {
+	public var producer: SignalProducer<Value, Error> {
+		return SignalProducer(self)
 	}
 }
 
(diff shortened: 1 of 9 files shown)


Sources/Signal.swift
```diff
@@ -426,6 +426,12 @@ extension Signal: SignalProtocol {
 	}
 }
 
+extension Signal: SignalProducerRepresentable {
+	public var producer: SignalProducer<Value, Error> {
+		return SignalProducer(self)
```
mdiep: Doesn't this mean taking some performance hit? The `SignalProducer` adds extra work on top of the `Signal`.

andersio: `Signal<Signal>` overloads are all implemented by lifting the inner `Signal`s to `SignalProducer`.

andersio: Lifting a `Signal` to `SignalProducer` does add an extra relaying signal. But there are ways to mitigate the overhead.

andersio:
Hmm perhaps we can call `ExpressibleBySignalProducer` `Observable` instead. :see_no_evil:

andersio:
I think the problem in general is that `extension Signal: SignalProducerConvertible` or `extension Signal: SignalProducerRepresentable` is kinda ambiguous because of the lack of perposition. The protocol here is also a one-way ticket.

sharplet:
I don't think a preposition is necessary though? I think it's pretty clear in your example _what_ is convertible to a `SignalProducer`, in the same way that it's clear what can be converted to a string when you conform to `CustomStringConvertible`.

andersio:
Hmmmmmmm. True.

andersio:
@mdiep The removal of `SignalProtocol` and `SignalProducerProtocol` was an oversight of mine and has been reverted, since they are still useful until we have parameterised extensions. e.g. ReactiveCollections use it to constrain generic delta producers for collection view bindings.

mdiep:
More merge conflicts. 😞

andersio:
It is just the changelog. 🚗

PR #354, opened 2017-04-28
`Bag` improvements.

andersio:
1. Rename `RemovalToken` to `Bag.Token`. This could be source breaking, since `Bag` has to be parameterised.

1. Replace the token object with a monotonic `UInt64` timestamp. In other words, `Bag` itself no longer incurs an ARC overhead except for its `ContiguousArray` storage. This improves the performance in iterations and copying.

1. Store the elements and the tokens in SoA form, improving cache locality in iterations and token searching during removals. That said best case removals seem to be affected by this change.

1. Exploit the timestamps to accelerate average case removals via binary search.

### DIFF
--- a/Sources/Bag.swift
+++ b/Sources/Bag.swift
@@ -10,12 +10,14 @@
 public struct Bag<Element> {
 	/// A uniquely identifying token for removing a value that was inserted into a
 	/// Bag.
-	public typealias Token = UInt64
+	public struct Token {
+		fileprivate let value: UInt64
+	}
 
 	fileprivate var elements: ContiguousArray<Element> = []
-	fileprivate var tokens: ContiguousArray<Token> = []
+	fileprivate var tokens: ContiguousArray<UInt64> = []
 
-	private var identifier: UInt64 = 0
+	private var nextToken: Token = Token(value: 0)
 
 	public init() {}
 
@@ -26,14 +28,14 @@ public struct Bag<Element> {
 	///   - value: A value that will be inserted.
 	@discardableResult
 	public mutating func insert(_ value: Element) -> Token {
-		let token = Token(identifier)
+		let token = nextToken
 
 		// Practically speaking, this would overflow only if we have 101% uptime and we
 		// manage to call `insert(_:)` every 1 ns for 500+ years non-stop.
-		identifier += 1
+		nextToken = Token(value: token.value + 1)
 
 		elements.append(value)
-		tokens.append(token)
+		tokens.append(token.value)
 
 		return token
 	}
@@ -46,7 +48,7 @@ public struct Bag<Element> {
 	///   - token: A token returned from a call to `insert()`.
 	public mutating func remove(using token: Token) {
 		for i in (elements.startIndex ..< elements.endIndex).reversed() {
-			if tokens[i] == token {
+			if tokens[i] == token.value {
 				tokens.remove(at: i)
 				elements.remove(at: i)
 				return
--- a/Sources/Bag.swift
+++ b/Sources/Bag.swift
@@ -6,13 +6,16 @@
 //  Copyright (c) 2014 GitHub. All rights reserved.
 //
 
-/// A uniquely identifying token for removing a value that was inserted into a
-/// Bag.
-public final class RemovalToken {}
-
 /// An unordered, non-unique collection of values of type `Element`.
 public struct Bag<Element> {
-	fileprivate var elements: ContiguousArray<BagElement<Element>> = []
+	/// A uniquely identifying token for removing a value that was inserted into a
+	/// Bag.
+	public typealias Token = UInt64
+
+	fileprivate var elements: ContiguousArray<Element> = []
+	fileprivate var tokens: ContiguousArray<Token> = []
+
+	private var identifier: UInt64 = 0
 
 	public init() {}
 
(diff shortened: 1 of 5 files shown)


NachoSoto:
Do you have a benchmark to show what this accomplishes?

andersio:
|Item|`bag-refactor`|`master`|
|---|---|---|
| Iteration, 1 element | 0.010 sec | 0.013 sec |
| Iteration, 128 elements | 0.407 sec | 0.794 sec |
| Insertion | 0.019 sec | 0.030 sec |
| Fast Path Removal | 0.030 sec | 0.020 sec |
| Worst Cast Removal | 0.304 | 24.258 sec |
| Signal Delivery | 0.119 sec | 0.195 sec |
| Disposable Detachment | 0.0006-0.0009 sec | 0.001-0.002 sec |

[Benchmark Source](https://gist.github.com/andersio/14e636c03b6915254fd2b9a948a83a20)
i5 6360U, Compiled with `-Owholemodule`.

mdiep:
Those benchmarks don't seem like real-world usage. I would expect most bags to have <10 elements and nearly all to have <100 elements. Do you agree? If so, could you re-run your benchmark with realistic sizes?

Sources/Bag.swift
```diff
@@ -22,11 +41,22 @@ public struct Bag<Element> {
...
+
+		repeat {
+			// `DispatchTime` is expected to use monotonic clocks. That said a guard is
+			// put here just in case.
+			timestamp = DispatchTime.now().uptimeNanoseconds
+		} while timestamp <= lastTimestamp
+
+		lastTimestamp = timestamp
+
+		let token = Token(timestamp)
```
mdiep: At this point, it doesn't seem like there's any point in using a timestamp. This could just use a counter? `Bag` requires mutual exclusion for cross-thread use anyway.

andersio: `Bag` was using a monotonic counter before. But overflowing seems to be a concern, so the identifier was stored inside a referenced token to allow reindexing as the overflow happens. Then after a while, we have stripped it since relying on the object identity is enough if an token object is used.

The point of a timestamp here is that the monotonic clock is (sadly?) guaranteed not to overflow.

andersio:
[Rewritten benchmark](https://gist.github.com/andersio/295492f3b91da8575574c28b92c551ac). It is either 1 or 8 elements/observers.

`bag-refactor`
```
@testAverageRemoval(): avg 9863 ns; min 8663 ns
@testCopying(): avg 1683 ns; min 1452 ns
@testDisposableDetachment(): avg 33346 ns; min 29684 ns
@testFastPathRemoval(): avg 8831 ns; min 7653 ns
@testInsertion(): avg 6118 ns; min 5299 ns
@testIteration(): avg 452 ns; min 393 ns
@testIterationOfOne(): avg 155 ns; min 122 ns
@testSignalDelivery(): avg 749 ns; min 651 ns
```

`master`
```
@testAverageRemoval(): avg 9476 ns; min 8332 ns
@testCopying(): avg 1315 ns; min 1145 ns
@testDisposableDetachment(): avg 27795 ns; min 24664 ns
@testFastPathRemoval(): avg 7781 ns; min 6696 ns
@testInsertion(): avg 6199 ns; min 5234 ns
@testIteration(): avg 906 ns; min 797 ns
@testIterationOfOne(): avg 189 ns; min 159 ns
@testSignalDelivery(): avg 1151 ns; min 988 ns
```

Iterations and insertions are no doubt faster even at small element count (1/8). But apparently removals and copying are slower.

Sources/Bag.swift
mdiep: Sure… but if we still used a `UInt64`, then this would have the same domain. Using a counter would start at an earlier value. It would also only increment by 1 for each insertion—instead of skipping values in the case of a timestamp. So using a counter would be less likely to overflow.

mdiep:
Without a real world benchmark, I don't know how we can evaluate the performance of those two implementations and make an informed choice.

Sources/Bag.swift
andersio: Yeah, true.

andersio:
Iteration alone is already a real world benchmark, if you consider the multiplicative effect in composed signal graphs.

mdiep:
But these seem like very low overhead. Will having from `906ms` to `452ms` really make a difference when scrolling a collection/scroll view?

If not, we should consider the conciseness/maintainability of the code.

andersio:
I considered it qualifying both conciseness and speediness. :)

Sources/Bag.swift
```diff
@@ -6,13 +6,16 @@
...
-/// A uniquely identifying token for removing a value that was inserted into a
-/// Bag.
-public final class RemovalToken {}
-
 /// An unordered, non-unique collection of values of type `Element`.
 public struct Bag<Element> {
-	fileprivate var elements: ContiguousArray<BagElement<Element>> = []
+	/// A uniquely identifying token for removing a value that was inserted into a
+	/// Bag.
+	public typealias Token = UInt64
```
mdiep: We should make this a `struct`—not just a `typealias`. A type alias won't prevent someone from passing a random `UInt64` to `remove(using:)`. `Token`s shouldn't be publicly instantiable.

Sources/Bag.swift
```diff
@@ -6,13 +6,16 @@
...
+
+	fileprivate var elements: ContiguousArray<Element> = []
+	fileprivate var tokens: ContiguousArray<Token> = []
+
+	private var identifier: UInt64 = 0
```
mdiep: Maybe call this `nextToken`?

PR #355, opened 2017-04-30
Revisit changes to the `Signal` lifetime semantics to improve tooling interoperability.

andersio:
### Rationale
Cease the self reference in `AliveState` which causes the `Signal` and all its retained objects being reported as leaks, so that ReactiveSwift would work better with the memory debuggers.

### Details
This PR adapts the original design in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2959.

To recap, the rule of automatic disposal is having no observer and having no external retain. This PR does not change this principle, but makes a few changes in the implementation:

1. A signal is now broken into a public shell `Signal` and an internal core `Signal.Core`.
   The shell is what the public users retain and interact with. The shell is neither strongly or weakly referenced by the core, the generator disposable or any disposable created.

1. The generator observer now holds a strong reference to `Signal.Core`, and pipes events to `Signal.Core.send(_:)`. Previously, it holds a _weak_ reference to `Signal`.

Here is the ownership graph from the perspective of an operator. Note that there is no circular strong reference in the graph as we had before.
 ```
 ------------               --------------                --------
 |          |               | endObserve |                |      |
 |          | <~~ weak ~~~  | disposable | <== strong === |      |
 |          |               --------------                |      | ... downstream(s)
 |          |                ------------                 |      |
 | upstream | === strong ==> | Observer |  === strong ==> | Core |
 ------------                ------------                 --------
                             ------------------              ^^
                             | Signal (shell) | === strong ==//
                             ------------------
 ```

These changes consequently lead to a few boundary cases that is **not source compatible** with ReactiveSwift 1.0:

1. A `nil` weak reference to a signal no longer means that the signal has been disposed of. `on(disposed:)` would become the only mean to assert this.

1. If the generator observer deinitialises, it now triggers the automatic disposal since no event can be sent to the `Signal` from this point of time. Previously, it has no side-effect.

    In other words, the rule of automatic disposal is now expanded to be (1) having no observer and having no external retain; OR (2) having the generator observer deinitialized".

These are not expected to be of huge impact.

### Implementation
The threading design and the logic of event delivery is generally untouched.

1. The implementation in `Signal` is moved to `Signal.Core`. `Signal` now only forwards calls to `Signal.Core`.

1. `Signal.Core.init` now disposed of the generator disposable immediately, if the signal is terminated within the generator. This was the correct and the original behavior, but had not been covered by the test cases and had been removed by previous changes. (Bugfix)

1. `tryTerminate` is now promoted to be `Signal.Core.tryCommitTermination`.

1. `Signal.Core.tryTerminateSilently` is introduced for automatic disposal. Calling it guarantees the `Signal` to be eventually terminated safely, but the termination may not happen immediately if there is an on-going event delivery.

1. `Signal.Core.shellDidDeinitialize` is introduced for the `Signal` shell to inform the `Signal` core of its deinitialization.

### Alternatives considered
1. Replace the explicit self reference with manual retains and releases via `Unmanaged`. This did not work as expected — the memory debuggers still report leaks.

1. Status quo, which was the alternative of this more complicated design in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2959.

### Testing

This has been tested with [a very barebone single UITableView project](https://github.com/andersio/RACTest) that had a tremendous amount of false positives from the memory debugger when compiled with `master`. With the PR, the memory debugger reports no leak.

Edit: This has also been tested with Carthage 0.22, and all test specs have passed. 

### Potential benefits
The design might enable an optimisation in property composition to dynamically collapse intermediate signals. Specifically in #340, it might allow the observers of the composed property's relay signal to be collapsed into the upstream signal when it is not externally retained. This is not possible at this moment due to the self reference and the semantic of `isKnownUniquelyReferenced`.


Tests/ReactiveSwiftTests/SignalSpec.swift
```diff
@@ -157,6 +157,42 @@ class SignalSpec: QuickSpec {
...
+				let signal: Signal<AnyObject, NoError> = Signal { observer in
+					observer.sendCompleted()
+					expect(disposable.isDisposed) == false
+					return disposable
+				}
+
+				expect(disposable.isDisposed) == true
+			}
+
+			it("should dispose of the returned disposable upon interrupted") {
```
NachoSoto: Failure

mdiep:
> A nil weak reference to a signal no longer means that the signal has been disposed of. on(disposed:) would become the only mean to assert this.

I'll have to think about the impact of this.

> If the generator observer deinitialises, it now triggers the automatic disposal since no event can be sent to the Signal from this point of time. Previously, it has no side-effect.
> 
> In other words, the rule of automatic disposal is now expanded to be (1) having no observer and having no external retain; OR (2) having the generator observer deinitialized".

Can you explain this a bit more? I haven't looked over the code, but I'm not sure when this would happen.

andersio:
> Can you explain this a bit more? I haven't looked over the code, but I'm not sure when this would happen.

The rule of active observers keeping the `Signal` remains unchanged, but the means to achieve it has changed:

1. Currently, it is done via the self reference. The self reference is conditionally cleared.

1. In this PR, it is done via the strong +1 of the generator observer to the `Signal.Core`.
   The `Signal.Core` conditionally dispose of its generator disposable, which causes the upstream to remove the generator observer from its bag, and eventually cease the strong +1 as the generator observer deinitializes.

So generally speaking, this change does not affect operators, but only the explicit, non-composition use of `Signal.init` and `Signal.pipe`.

Say we have a pipe and we have observed it:
```swift
func scope() {
    let (signal, observer) = Signal.pipe()
    signal.observe { _ in }
}
```

Currently, as `scope()` returns, the self reference asserted by the sole active observer would keep the `Signal` alive, despite the fact that statically no further event can be sent.

But with this PR, the `Signal.Core` would lost all its retains, so it would be disposed of itself.

mdiep:
Can you add the external observers to your diagram above? I think that's the tricky piece to think about.

Would it work to replace the self reference with an actual strong reference from the observer to the signal? That seems like it would maintain the existing semantics (which are conceptually simpler IMO) but also fix the tooling issues.

> A nil weak reference to a signal no longer means that the signal has been disposed of. on(disposed:) would become the only mean to assert this.

I think it's preferable to _not_ have this be the case, which is why I ask. But if this the only way to improve tooling interop, then I think it's worth it.

andersio:
> Would it work to replace the self reference with an actual strong reference from the observer to the signal? That seems like it would maintain the existing semantics (which are conceptually simpler IMO) but also fix the tooling issues.

It does not solve the issue we have with the current implementation, since it is an retain cycle despite being deliberately made.

Specifically, if an observer now holds a strong reference to the `Signal`, it forms a legit retain cycle. We cannot break the cycle with weak references as we would never require observers to be retained externally.

```
 ------------               --------------                --------
 |          |               | endObserve |                |      |
 |          | <~~ weak ~~~  | disposable | <== strong === |      |
 |          |               --------------                |      | ... downstream(s)
 |          |                ------------                 |      |
 | upstream | === strong ==> | Observer |  === strong ==> | Core |
 ------------ ===\\          ------------                 -------- ===\\
                  \\         ------------------              ^^        \\
                   \\        | Signal (shell) | === strong ==//         \\
                    \\       ------------------                          \\
                    || strong                                            || strong
                    vv                                                   vv
            -------------------                                 -------------------
            | Other observers |                                 | Other observers |
            -------------------                                 -------------------
```

andersio:
A side note: I have tried many tricks, hoping to trick the memory debugger not to recognise the retain cycle. But it doesn't seem working at all.

1. A `Unmanaged` pointer.

2. A `UnsafeMutablePointer` pointer.

3. An `OpaquePointer`.

4. Storing a `UnsafeMutablePointer` as a raw `Int`.

5. Storing a `UnsafeMutablePointer` as a raw `Int` that is right shifted by 1 bit.
    This does "break" the retain cycle in the debugger, but the debugger still consider it a leak.

Sources/Signal.swift
```diff
@@ -18,119 +18,106 @@ import Result
...
-				}
+			updateLock = Lock.make()
+			sendLock = Lock.make()
 
-				signal.updateLock.unlock()
-				return false
-			}
+			shellDeinitialized = false
+			disposable = SerialDisposable()
+			disposable.inner = generator(Observer(self.send))
```
mdiep: Maybe add a comment here to highlight that it's the other +1?

mdiep:
That test failure looks like it might be legit.

andersio:
`MutableProperty should not deadlock` occasionally fails on single-core CI. 🙈

PR #363, opened 2017-05-05
Remove `CompositeDisposable.DisposableHandle`.

andersio:
Cherry-picked from #334.

Use `ActionDisposable` instead, akin to `Signal.observe`.

### DIFF
--- a/Sources/Disposable.swift
+++ b/Sources/Disposable.swift
@@ -170,18 +170,18 @@ public final class CompositeDisposable: Disposable {
 	///            `disposable` is `nil`.
 	@discardableResult
 	public func add(_ disposable: Disposable?) -> Disposable? {
-		if isDisposed {
+		guard let d = disposable, !d.isDisposed, !isDisposed else {
 			disposable?.dispose()
 			return nil
 		}
 
-		return disposable.flatMap { disposable in
-			return disposables.modify { disposables in
-				let token = disposables!.insert(disposable)
-				return ActionDisposable { [weak self] in
-					self?.disposables.modify {
-						$0?.remove(using: token)
-					}
+		return disposables.modify { disposables in
+			guard disposables != nil else { return nil }
+
+			let token = disposables!.insert(d)
+			return ActionDisposable { [weak self] in
+				self?.disposables.modify {
+					$0?.remove(using: token)
 				}
 			}
 		}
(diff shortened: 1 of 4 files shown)


Sources/Disposable.swift
```diff
@@ -202,41 +160,43 @@ public final class CompositeDisposable: Disposable {
...
-		}
-
-		if let handle = handle {
-			return handle
-		} else {
-			d.dispose()
-			return DisposableHandle.empty
+		return disposable.flatMap { disposable in
+			return disposables.modify { disposables in
+				let token = disposables!.insert(disposable)
```
mdiep: I don't think this `!` is safe. This could have been disposed between the above check and here.

Sources/Disposable.swift
```diff
@@ -202,41 +160,43 @@ public final class CompositeDisposable: Disposable {
...
-		let handle: DisposableHandle? = disposables.modify {
-			return ($0?.insert(d)).map { DisposableHandle(bagToken: $0, disposable: self) }
-		}
-
-		if let handle = handle {
-			return handle
-		} else {
-			d.dispose()
-			return DisposableHandle.empty
+		return disposable.flatMap { disposable in
```
mdiep: I think the previous `guard` structure is preferable to `disposable.flatMap` to reduce the nesting by a level.

Sources/Disposable.swift
```diff
@@ -202,41 +160,43 @@ public final class CompositeDisposable: Disposable {
...
-	/// - returns: An instance of `DisposableHandle` that can be used to
-	///            opaquely remove the disposable later (if desired).
+	/// - returns: A disposable to remove `disposable` from the composite. `nil` if the
+	///            composite has been disposed of, `disposable` has been disposed of, or
+	///            `disposable` is `nil`.
 	@discardableResult
-	public func add(_ d: Disposable?) -> DisposableHandle {
-		guard let d = d else {
-			return DisposableHandle.empty
+	public func add(_ disposable: Disposable?) -> Disposable? {
```
mdiep: We could leave this as non-optional by returning a disposable that does nothing. I'm not sure which would be better. Is there a particular reason you decided to return an optional?

andersio: It follows the precedence of `observe`, and the fact that `CompositeDisposable.add` accepts `Disposable?`.

PR #365, opened 2017-05-08
Add inout variants of `reduce` and `scan`

ikesyo:
Rationale: https://github.com/apple/swift-evolution/blob/master/proposals/0171-reduce-with-inout.md


NachoSoto:
👌

Issue #204, opened 2016-12-31
Rename `Signal.observe(_:)`


dmcrodrigues:
`observe(_:)` is a helpful operator but I think the name can be misleading. Considering the following common pattern.

```swift
signal.observe(observer)
```

We may be induced to think the `signal` is observing `observer` and in fact it's the other way around (indirectly).

Introducing a named parameter could help to better express intent. 

```swift
signal.observe(using: observer)
signal.observe(on: observer)
```

However we already have `func observe(on scheduler: SchedulerProtocol)` which is something completely different. Another option could be renaming the operator to a completely different name and more close of the expected intent.

A few options I'm thinking.

```swift
signal.bind(observer)
signal.register(observer)
signal.subscribe(observer)
```

Does anyone agrees with this?

andersio:
`observe` is an established term of art since the ObjC days, so as `start` since 3.0. I don't think we'd ever gonna change it.

Adding a preposition as label should be fairly enough to clarify the intent, and form grammatical phrases at the a use site (_Strive for Fluent Usage, Swift API Guidelines)_.

`with:` is also a contender — shorter than `using` but generally inferred by many to the same meaning.

RuiAAPeres:
> observe is an established term of art since the ObjC days, so as start since 3.0. I don't think we'd ever gonna change them.

That's true, but on the other hand I liked these both:

```swift
signal.register(observer)
signal.subscribe(observer)
``` 

The `with:` label is probably a good mid ground.

mdiep:
I think the key point is that you _observe_ a signal. So `observe` makes sense as the method name. That's clearest and it's a term of art.

I don't think a named parameter is necessary either. If we _did_ add a label, `with` or `using` seems most correct. But I think that draws attention away from the _observe_ part of the name.

yoching:
I have same feelings about the misleading of the naming.
I always convert in my head
from
`
signal.observe(observer)
`
to
`
signal.observed(by: observer)
`

liscio:
> I don't think a named parameter is necessary either. If we did add a label, with or using seems most correct. But I think that draws attention away from the observe part of the name.

@mdiep The trouble, though, is the most likely variable naming to follow as illustrated above. `observe(observer)` is far less readable than `observe(using: observer)`. Those that supply a trailing closure won't need to worry about the label…

[...]

PR #369, opened 2017-05-11
Documentation: Add note about observe() side effects in SignalProducer

tikitu:
This tripped me up (as a reactive newbie), leading to a rather confused debugging session: I hadn't realised that `observe()` on a SignalProducer triggers the underlying operation (with all side effects).

I see renaming `observe()` has been discussed in #204 and #178 so I presume that's not an option. If it were, my vote would be to rename `SignalProducer.observe()` to something like `SignalProducer.start(with:during:)` to emphasise the side-effecty similarity with plain `start()`. But without a rename, I think it's at least useful to make this a little more prominent in the docs.

### DIFF
--- a/Documentation/FrameworkOverview.md
+++ b/Documentation/FrameworkOverview.md
@@ -79,7 +79,7 @@ A **signal producer**, represented by the [`SignalProducer`][SignalProducer] typ
 [signals](#signals) and performs side effects.
 
 They can be used to represent operations or tasks, like network
-requests, where each invocation of `start()` will create a new underlying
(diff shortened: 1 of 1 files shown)


andersio:
`SignalProducer.observe(_:during:)` has been removed in #359.

tikitu:
Whoah, quick response! Thanks, I'm glad the problem is going away (and I definitely should take a look at the plans for 2.0, sounds interesting).

andersio:
2.0 is generally about cleaning up the APIs here and there. If you find anything else that doesn't make sense to you, feel free to bring it up. 😸

PR #371, opened 2017-05-11
Make `UnsafeAtomicState` safe.

andersio:
Replace the manual allocation of `UnsafeAtomicState` with `withUnsafeMutablePointer`.

### DIFF
--- a/Sources/Atomic.swift
+++ b/Sources/Atomic.swift
@@ -12,26 +12,17 @@ import MachO
 #endif
 
 /// A simple, generic lock-free finite state machine.
-///
-/// - warning: `deinitialize` must be called to dispose of the consumed memory.
-internal struct UnsafeAtomicState<State: RawRepresentable> where State.RawValue == Int32 {
+internal struct AtomicState<State: RawRepresentable> where State.RawValue == Int32 {
 	internal typealias Transition = (expected: State, next: State)
 #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
-	private let value: UnsafeMutablePointer<Int32>
+	private var value: Int32
 
 	/// Create a finite state machine with the specified initial state.
 	///
 	/// - parameters:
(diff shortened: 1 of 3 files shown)


Sources/Atomic.swift
```diff
@@ -55,10 +48,12 @@ internal struct UnsafeAtomicState<State: RawRepresentable> where State.RawValue
...
-	internal func tryTransition(from expected: State, to next: State) -> Bool {
-		return OSAtomicCompareAndSwap32Barrier(expected.rawValue,
-		                                       next.rawValue,
-		                                       value)
+	internal mutating func tryTransition(from expected: State, to next: State) -> Bool {
+		return withUnsafeMutablePointer(to: &value) { valuePointer in
+			return OSAtomicCompareAndSwap32Barrier(expected.rawValue,
+			                                       next.rawValue,
+			                                       valuePointer)
+		}
```
mdiep: I'm not sure this is valid.

This:

> The pointer argument to body is valid only for the lifetime of the closure. Do not escape it from the closure for later use.

Makes me think that the whole construct may not be atomic.

andersio: I did check the compiler source and generated assembly. `withUnsafeMutablePointer` uses `Builtin.addressof`, which would emit the in-place address if the variable has direct storage. IOW it is not atomic only if `self` is not a final stored property.

andersio: I don't see this bit being dropped in the foreseeable future. It is essential, well-known part of the `inout` model since Swift 1.0.

Moreover, it is fair to say that anything concurrent written in Swift is in fact depending on "undocumented" implementation details of the Swift compiler, since Swift defines no concurrent and memory model.

😅

andersio:
[I guess it'd be better to follow Apple's suit](https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSLock.swift).

PR #376, opened 2017-05-13
Nest `Event` inside `Signal`.

andersio:


Sources/Deprecations+Removals.swift
```diff
@@ -235,21 +238,6 @@ extension Action {
 }
 
 // Renamed Enum cases
-
-extension Event {
```
andersio: These are no longer necessary, since SourceKit has auto-suggestion that covers capitalisation.

Sources/Signal.swift
```diff
@@ -391,14 +391,14 @@ public final class Signal<Value, Error: Swift.Error> {
 		fileprivate let observers: Bag<Signal<Value, Error>.Observer>
 
 		///  The termination event.
-		fileprivate let event: Event<Value, Error>
+	  fileprivate let event: Event
```
sharplet: oops

PR #377, opened 2017-05-13
Nest `Observer` inside `Signal`.

andersio:


Sources/Deprecations+Removals.swift
```diff
@@ -324,22 +327,6 @@ extension CompositeDisposable {
 	public func addDisposable(_ d: Disposable) -> Disposable? { fatalError() }
 }
 
-extension Observer {
```
andersio: This extension is removed since somehow the compiler cannot resolve `extension Signal.Observer` in this particular file.

mdiep: I think all of the 1.0 `unavailable`s can be removed at this point.

PR #391, opened 2017-05-15
Add more general timeout() operator to send any event

sharplet:
This came out of a discussion in Slack, and seemed like a useful addition.

This more general implementation of `timeout()` allows any kind of event to be sent, not just errors. For backwards compatibility, the existing `timeout(after:raising:on:)` operator now calls through to the new implementation.

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -2363,10 +2363,10 @@ extension Signal {
 }
 
 extension Signal {
-	/// Forward events from `self` until `interval`. Then if signal isn't 
+	/// Forward events from `self` until `interval`. Then if signal isn't
 	/// completed yet, fails with `error` on `scheduler`.
 	///
-	/// - note: If the interval is 0, the timeout will be scheduled immediately. 
+	/// - note: If the interval is 0, the timeout will be scheduled immediately.
 	///         The signal must complete synchronously (or on a faster
 	///         scheduler) to avoid the timeout.
 	///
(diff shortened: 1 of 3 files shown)


Sources/SignalProducer.swift
```diff
@@ -1385,6 +1410,35 @@ extension SignalProducer where Error == NoError {
...
+	///            when `interval` passes.
+	///   - interval: Number of seconds to wait for `self` to complete.
+	///   - scheduler: A scheduler to deliver error on.
+	///
+	/// - returns: A signal that sends events for at most `interval` seconds,
+	///            then, if not `completed` - sends `error` with failed event
+	///            on `scheduler`.
+	public func timeout<NewError: Swift.Error>(
+		after interval: TimeInterval,
+		sending event: SignalProducer<Value, NewError>.ProducedSignal.Event,
```
andersio: This can be shortened to `Signal<Value, NewError>.Event`.

Sources/Signal.swift
```diff
@@ -2260,23 +2260,55 @@ extension Signal {
...
 		precondition(interval >= 0)
 
 		return Signal { observer in
 			let disposable = CompositeDisposable()
 			let date = scheduler.currentDate.addingTimeInterval(interval)
+			var signalDisposable: Disposable?
 
 			disposable += scheduler.schedule(after: date) {
-				observer.send(error: error)
+				signalDisposable?.dispose()
```
andersio: `signalDisposable` doesn't seem necessary.

`disposable` would be disposed of when the created `Signal` terminates, which in turn disposes of the observer handle anyway.

Tests/ReactiveSwiftTests/SignalSpec.swift
```diff
@@ -2529,6 +2529,78 @@ class SignalSpec: QuickSpec {
...
+			var observer: Signal<Int, TestError>.Observer!
+
+			beforeEach {
+				testScheduler = TestScheduler()
+				let (baseSignal, incomingObserver) = Signal<Int, TestError>.pipe()
+				signal = baseSignal.timeout(after: 2, sending: .value(-1), on: testScheduler)
+				observer = incomingObserver
+			}
+
+			it("should complete if within the interval") {
```
andersio: Could you please test also sending `interrupted`, and whether the value is really delivered when `sending: .value(...)` is used?

Sources/SignalProducer.swift
sharplet: I briefly considered introducing a typealias to allow `SignalProducer<Value, NewError>.Event` — what do you think about that approach?

Sources/Signal.swift
sharplet: I think there could be a race condition between sending a value and sending the completed event, where another event on the underlying signal could be sent:

```
// scheduled timeout
|--------------------A->|
// underlying signal
|---------------------B------------->
// returned signal
|--------------------AB>|
```

But I couldn't think of a test for it. Any ideas? Does this look like a real issue to you?

Tests/ReactiveSwiftTests/SignalSpec.swift
sharplet: There's another test that covers the `.value` case.

Sources/SignalProducer.swift
andersio: For this specific case, we need to parameterise `Signal.Event` anyway, so the type alias doesn't seem helping at all.

andersio: In general, `SignalProducer` is forever intertwined with `Signal` anyway, so IMO such typealias of `Event` or say `Observer` do not seem valuable.

Sources/Signal.swift
andersio: Hmm, it is a legit race if we guarantee that the `value` sent by `timeout` must be followed by a terminal event. You might want to use `SerialDisposable` for `signalDisposable` then, since statically it is possible that the scheduled block runs before the store of the observer handle.

Sources/SignalProducer.swift
sharplet: I was thinking more for consistency: in `Signal` we can just reference `Event`, but in `SignalProducer` we need to use `ProducedSignal.Event`. I was thinking that

```
typealias Event = ProducedSignal.Event
```

would allow for consistency across the two types.

mdiep: 👍 to the `typealias`

Tests/ReactiveSwiftTests/SignalSpec.swift
mdiep: This is still waiting on an `interrupted` test, right?

sharplet: Yes

sharplet: And the race condition: @andersio I could use a hand there, I'm not sure I follow your suggestions.

Sources/Signal.swift
andersio: @sharplet What I mean is that you can keep the current form, and just replace `Disposable?` with a `SerialDisposable` for `signalDisposable`.

andersio: This should work.

```swift
return Signal { observer in
	let disposable = SerialDisposable()

	disposable.inner = self.observe { event in
		// Guarantee that no event is delivered before we emit
		// the timeout event sequence.
		guard !disposable.isDisposed else { return }
		observer.action(event)
	}

	let d2 = scheduler.schedule(after: date) {
		disposable.dispose()
		observer.action(event)

		if !event.isTerminating {
			observer.sendCompleted()
		}
	}

	return ActionDisposable { disposable.dispose(); d2?.dispose() }
}
```

andersio:
Closing due to inactivity. It also appears that our friends in Combine and Rx do not have such a counterpart, so I will reevaluate later to see if a follow-up is necessary. 😣

Issue #309, opened 2017-04-01
Race condition on `enabledIf` property of Action


almassapargali:
With code like this: 
```swift
let notResponding = MutableProperty(true)
let acceptAction = respondAction(respondable, true, enabledIf: notResponding)
let declineAction = respondAction(respondable, false, enabledIf: notResponding)
notResponding <~ acceptAction.isExecuting.producer, declineAction.isExecuting.producer)
    .map { !($0 || $1) }
```

I get `Signal.sendLock` as soon as any of actions stars. I think it's locking when user defined `isEnabled` and action's own `isEnabled` changes at the same time, since following code works:

```swift
let notRespondingAccept = MutableProperty(true)
let notRespondingDecline = MutableProperty(true)
let acceptAction = respondAction(respondable, true, enabledIf: notRespondingAccept)
let declineAction = respondAction(respondable, false, enabledIf: notRespondingDecline)
notRespondingAccept <~ declineAction.isExecuting.producer.map { !$0 }
notRespondingDecline <~ acceptAction.isExecuting.producer.map { !$0 }
```

I'm on `github "ReactiveCocoa/ReactiveSwift" "1.1.0"`

andersio:
There is one single internal property overseeing the state. So if you indirectly pipe `isExecuting` into the `Action` as the user-controled availability, it deadlocks.

Could you state what you want to achieve here, so that we may perhaps figure out an alternative?

almassapargali:
So those two actions eventually get bound to two buttons `accept | decline`. And if user presses accept, I want decline be disabled while accepting in progress.

almassapargali:
Personally I think ReactiveSwift should provide a way to allow only one of multiple actions at a time, and they way I coded it looks totally ok from developers perspective (imho). Maybe we can allow that code, but somehow avoid deadlocking?

andersio:
One alternative I can think of is having both of them covered by the same `Action`, and differentiate the course of action by the input (e.g. an enum) instead.

almassapargali:
Wow, great idea, how I missed that. Thank you very much, it really eliminates need for other action and property. I think I should close this one since it really led me to more complex implementation, and now after thinking the proper way to solve

> provide a way to allow only one of multiple actions at a time

would be one Action with enum as state or input.

PR #400, opened 2017-05-22
Enable `Action` executing state feedback to the state property.

andersio:
Related: #309.

There are legitimate uses of `isExecuting` feedback to the `Action` state property. For example, we may link the availability of multiple actions together, so that all disables when one executes.

This PR refactors the internal of `Action` so that such use cases are permitted. Legitimate infinite feedback loops would still result in deadlocks.

Example:
```swift
let isIdle = MutableProperty(true)

candyCrusher = Action(enabledIf: isIdle) { ... }
fruitCrusher = Action(enabledIf: isIdle) { ... }
carrotCrusher = Action(enabledIf: isIdle) { ... }

isIdle <~ candyCrusher.isExecuting.negate()
isIdle <~ fruitCrusher.isExecuting.negate()
isIdle <~ carrotCrusher.isExecuting.negate()
```

### Alternative/Follow-up
Could we have some kind of "action group"s that determine the availability at a coarser granularity?

### DIFF
--- a/Sources/Action.swift
+++ b/Sources/Action.swift
@@ -98,15 +98,16 @@ public final class Action<Input, Output, Error: Swift.Error> {
 		deinitToken = Lifetime.Token()
 		lifetime = Lifetime(deinitToken)
 
-		// `isExecuting` has its own `Signal`, so that legitimate feedbacks would
-		// not deadlock.
 		let actionState = MutableProperty(ActionState<State.Value>(isUserEnabled: true, isExecuting: false, value: state.value))
 		self.isEnabled = actionState.map { $0.isEnabled }
 
+		// `isExecuting` has its own backing so that when the observer of `isExecuting`
+		// synchronously affects the action state, the signal of the action state does not
+		// deadlock due to the recursion.
 		let isExecuting = MutableProperty(false)
 		self.isExecuting = Property(capturing: isExecuting)
 
-		// Associate the state property with the created `Action`.
+		// `Action` retains its state property.
 		lifetime.observeEnded { _ = state }
 
 		(events, eventsObserver) = Signal<Signal<Output, Error>.Event, NoError>.pipe()
@@ -127,19 +128,25 @@ public final class Action<Input, Output, Error: Swift.Error> {
 
 		self.execute = { action, input in
 			return SignalProducer { observer, lifetime in
-				func tryStart() -> State.Value? {
-					return actionState.modify { state in
-						guard state.isEnabled else {
-							return nil
-						}
+				var notifiesExecutionState = false
 
-						state.isExecuting = true
+				func didSet() {
+					if notifiesExecutionState {
 						isExecuting.value = true
-						return state.value
 					}
 				}
 
-				guard let state = tryStart() else {
+				let latestState: State.Value? = actionState.modify(didSet: didSet) { state in
+					guard state.isEnabled else {
+						return nil
+					}
+
+					state.isExecuting = true
+					notifiesExecutionState = true
+					return state.value
+				}
+
+				guard let state = latestState else {
 					observer.send(error: .disabled)
 					action.disabledErrorsObserver.send(value: ())
 					return
@@ -153,9 +160,8 @@ public final class Action<Input, Output, Error: Swift.Error> {
 				lifetime.observeEnded {
 					interruptHandle.dispose()
 
-					actionState.modify { state in
+					actionState.modify(didSet: { isExecuting.value = false }) { state in
 						state.isExecuting = false
-						isExecuting.value = false
 					}
 				}
 			}
--- a/Sources/Action.swift
+++ b/Sources/Action.swift
@@ -18,11 +18,21 @@ import Result
 ///
 /// `Action` enforces serial execution, and disables the `Action` during the execution.
 public final class Action<Input, Output, Error: Swift.Error> {
-	private let deinitToken: Lifetime.Token
+	private struct ActionState<Value> {
+		var isEnabled: Bool {
+			return isUserEnabled && !isExecuting
+		}
 
-	private let executeClosure: (_ state: Any, _ input: Input) -> SignalProducer<Output, Error>
+		var isUserEnabled: Bool
+		var isExecuting: Bool
+		var value: Value
+	}
+
+	private let execute: (Action<Input, Output, Error>, Input) -> SignalProducer<Output, ActionError<Error>>
 	private let eventsObserver: Signal<Signal<Output, Error>.Event, NoError>.Observer
 	private let disabledErrorsObserver: Signal<(), NoError>.Observer
+
+	private let deinitToken: Lifetime.Token
 
 	/// The lifetime of the `Action`.
 	public let lifetime: Lifetime
@@ -60,8 +70,6 @@ public final class Action<Input, Output, Error: Swift.Error> {
 	/// Whether the action is currently enabled.
 	public let isEnabled: Property<Bool>
 
-	private let state: MutableProperty<ActionState>
-
 	/// Initializes an `Action` that would be conditionally enabled depending on its
 	/// state.
 	///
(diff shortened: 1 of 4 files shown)


Sources/Action.swift
```diff
@@ -80,39 +91,103 @@ public final class Action<Input, Output, Error: Swift.Error> {
...
+			return SignalProducer { observer, disposable in
+				func tryStart() -> State.Value? {
+					return actionState.modify { state in
+						guard state.availability == .enabledIdle else {
+							return nil
+						}
+
+						state.availability = .enabledExecuting
+						isExecuting.value = true
+						isEnabled.value = false
```
mdiep: Why is `isEnabled` `false` if `availability` is `enabledExecuting`?

Sources/Action.swift
```diff
@@ -80,39 +91,103 @@ public final class Action<Input, Output, Error: Swift.Error> {
...
 
-		executeClosure = { state, input in execute(state as! State.Value, input) }
+		// `isExecuting` and `isEnabled` should use their own locks, so that legitimate
+		// feedbacks would not deadlock.
+		let initial = (availability: Availability.enabledIdle, value: state.value)
+		let actionState = MutableProperty<ActionState>(initial)
+		let isExecuting = MutableProperty(false)
+		let isEnabled = MutableProperty(true)
+		self.isExecuting = Property(capturing: isExecuting)
+		self.isEnabled = Property(capturing: isEnabled)
```
mdiep: Can these be derived from `actionState` instead of duplicating that information in separate properties?

Sources/Action.swift
```diff
@@ -80,39 +91,103 @@ public final class Action<Input, Output, Error: Swift.Error> {
...
-		property.signal
-			.take(during: state.lifetime)
-			.observeValues { [weak state] newValue in
-				state?.modify {
-					$0.value = newValue
+				case (false, .enabledExecuting):
+					state.availability = .disabledExecuting
+
+				default:
+					break
```
mdiep: This `default` shouldn't be needed?

Sources/Action.swift
```diff
@@ -80,39 +91,103 @@ public final class Action<Input, Output, Error: Swift.Error> {
...
 		deinitToken = Lifetime.Token()
 		lifetime = Lifetime(deinitToken)
-		
-		// Retain the `property` for the created `Action`.
-		lifetime.observeEnded { _ = property }
 
-		executeClosure = { state, input in execute(state as! State.Value, input) }
+		// `isExecuting` and `isEnabled` should use their own locks, so that legitimate
+		// feedbacks would not deadlock.
+		let initial = (availability: Availability.enabledIdle, value: state.value)
```
mdiep: Can you make this a `struct` instead of a `typealias`?

Sources/Action.swift
```diff
@@ -18,12 +18,19 @@ import Result
 ///
 /// `Action` enforces serial execution, and disables the `Action` during the execution.
 public final class Action<Input, Output, Error: Swift.Error> {
-	private let deinitToken: Lifetime.Token
+	private enum Availability {
+		case enabledIdle
+		case enabledExecuting
+		case disabledIdle
+		case disabledExecuting
```
mdiep: This is really modeling two independent boolean states. Can we separate these and add them to `ActionState` individually? I think that will simplify the code a bit as well.

Sources/Action.swift
```diff
@@ -37,13 +44,17 @@ public final class Action<Input, Output, Error: Swift.Error> {
 	///
 	/// In other words, this sends every value from every unit of work that the `Action`
 	/// executes.
-	public let values: Signal<Output, NoError>
+	public private(set) lazy var values: Signal<Output, NoError> = { [unowned self] in
```
mdiep: Can we make this `lazy` change in a separate PR? I'm not sold on it, and it's not essential to this PR.

Sources/Action.swift
andersio: `Action` s are disabled when executing.

Sources/Action.swift
andersio: There are eight combinations in total, and only four need action.

Sources/Action.swift
andersio: The signal of `ActionState` is deliberately ignored, since `Signal` doesn't support recursion with value events. 

`Signal` deadlocks when one do `isEnabled <~ isExecuting` and both are derived from the same `MutableProperty`, i.e. the current implementation.

Sources/Action.swift
andersio: The problem is that `isEnabled` and `isExecuting` are not independent. They together form a finite state machine, and IMO named cases are easier to understand than a switch with tuples of booleans.

andersio: Changed in https://github.com/ReactiveCocoa/ReactiveSwift/pull/400/commits/4acaf81a54c9da4940b8484cf745f73ae3fd14d7 anyway.

Sources/Action.swift
andersio: Reverted in https://github.com/ReactiveCocoa/ReactiveSwift/pull/400/commits/4acaf81a54c9da4940b8484cf745f73ae3fd14d7.

Sources/Action.swift
andersio: Done in https://github.com/ReactiveCocoa/ReactiveSwift/pull/400/commits/4acaf81a54c9da4940b8484cf745f73ae3fd14d7.

Sources/Action.swift
andersio: In https://github.com/ReactiveCocoa/ReactiveSwift/pull/400/commits/4acaf81a54c9da4940b8484cf745f73ae3fd14d7, `isEnabled` is derived since we are permitting only `state <~ isExecuting` here, where `state` is the state of the `Action` that flips its availability (`isEnabled`).

Sources/Action.swift
```diff
@@ -80,39 +88,78 @@ public final class Action<Input, Output, Error: Swift.Error> {
...
-		lifetime.observeEnded { _ = property }
 
-		executeClosure = { state, input in execute(state as! State.Value, input) }
+		// `isExecuting` has its own `Signal`, so that legitimate feedbacks would
+		// not deadlock.
+		let actionState = MutableProperty(ActionState<State.Value>(isUserEnabled: true, isExecuting: false, value: state.value))
+		self.isEnabled = actionState.map { $0.isEnabled }
+
+		let isExecuting = MutableProperty(false)
+		self.isExecuting = Property(capturing: isExecuting)
```
mdiep: Can this be `self.isExecuting = actionState.map { $0.isExecuting }`?

Sources/Action.swift
```diff
@@ -80,39 +88,78 @@ public final class Action<Input, Output, Error: Swift.Error> {
...
+
+		// Associate the state property with the created `Action`.
+		lifetime.observeEnded { _ = state }
```
mdiep: This comment is describing _what_ this line does. Can you update it to describe _why_ it needs to be associated?

Sources/Action.swift
```diff
@@ -80,39 +88,78 @@ public final class Action<Input, Output, Error: Swift.Error> {
...
+			}
+		}
+
+		lifetime.observeEnded(disposable.dispose)
 
-		let initial = ActionState(value: property.value, isEnabled: { isEnabled($0 as! State.Value) })
-		state = MutableProperty(initial)
+		self.execute = { action, input in
+			return SignalProducer { observer, lifetime in
+				func tryStart() -> State.Value? {
```
mdiep: This is only called once. Maybe take the code out of the function?

```swift
let state = actionState.modify { state in
    ...
}

guard let state = state else {
...
```

Sources/Action.swift
andersio: The point of giving it its own backing is that `isExecuting` calls out to the observers outside the signal of `actionState`.

However, it does seem violating exclusivity of access so it might need a change anyway. We might need to add guards in `MutableProperty` to prevent nested modifications too.

Sources/Action.swift
```diff
@@ -80,39 +88,84 @@ public final class Action<Input, Output, Error: Swift.Error> {
...
+				func didSet() {
+					if notifiesExecutionState {
+						isExecuting.value = true
+					}
 				}
-			}
 
-		self.isEnabled = state.map { $0.isEnabled }.skipRepeats()
-		self.isExecuting = state.map { $0.isExecuting }.skipRepeats()
+				let latestState: State.Value? = actionState.modify(didSet: didSet) { state in
```
mdiep: This `didSet` seem pointless. The function is only called here, right after `notifiesExecutionState` is set to `false`—so it will always do nothing. It seem like both `didSet()` and `notifiesExecutionState` can be removed?

andersio: `notifiesExecutionState` is set to true when the current execution attempt succeeds and starts. `didSet` is invoked after the trailing closure modifying `actionState`.

andersio: `notifiesExecutionState` is set to true when the current execution attempt succeeds and starts. `didSet` is invoked after the trailing closure modifying `actionState`.

PR #404, opened 2017-05-23
Signal resource management with `Lifetime`.

andersio:
Follow-up to #334.

This is not as trivially migratable as #334, since `Signal` had a different design choice in the signal generator closure than the `SignalProducer`. Let alone the fact that [SE-0110](https://github.com/apple/swift-evolution/blob/master/proposals/0110-distingish-single-tuple-arg.md) is not implemented in Swift 3, adding obstacles to automatic migration.

1. Change the `Signal` generator signature to `(Observer, Lifetime) -> Void`, so that it matches the style of `SignalProducer.init`.
   ```swift
   let signal = Signal { observer, lifetime in
      let disposable = self.observe { ... }
      _ = (disposable?.dispose).map(lifetime.observeEnded)
   }

   let producer = SignalProducer { observer, lifetime in
      let disposable = producer.start { ... }
      lifetime.observeEnded(disposable.dispose)
   }
   ```

### DIFF
--- a/Sources/Flatten.swift
+++ b/Sources/Flatten.swift
@@ -289,8 +289,7 @@ extension Signal where Value: SignalProducerConvertible, Error == Value.Error {
 		precondition(limit > 0, "The concurrent limit must be greater than zero.")
 
 		return Signal<Value.Value, Error> { relayObserver, lifetime in
-			let disposable = self.observeConcurrent(relayObserver, limit, lifetime)
-			_ = (disposable?.dispose).map(lifetime.observeEnded)
+			lifetime += self.observeConcurrent(relayObserver, limit, lifetime)
 		}
 	}
 
@@ -528,10 +527,8 @@ extension Signal where Value: SignalProducerConvertible, Error == Value.Error {
 	fileprivate func switchToLatest() -> Signal<Value.Value, Error> {
 		return Signal<Value.Value, Error> { observer, lifetime in
 			let serial = SerialDisposable()
-			lifetime.observeEnded(serial.dispose)
-
-			let disposable = self.observeSwitchToLatest(observer, serial)
-			_ = (disposable?.dispose).map(lifetime.observeEnded)
+			lifetime += serial
+			lifetime += self.observeSwitchToLatest(observer, serial)
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


mdiep:
> Change the Signal generator signature to (Observer, Lifetime) -> Void, so that it matches the style of SignalProducer.init.

This makes sense to me. (Although I did think about it for quite a while because I wasn't sure.)

> Replace pipe(disposable:) with a new variant of pipe returning a 3-tuple (input, output and lifetime).

This change seems less necessary and also orthogonal. I'm not convinced we should do this.

---

Unfortunately, this is all too late for 2.0, so it will also need to be changed to be a non-breaking change (adding an `init` instead of changing it) or wait for 3.0.

andersio:
@mdiep Sadly as a new `init` it would be ambiguous, unless we expose it as static function.

mdiep:
3.0 it is!

Hopefully we/I can be a little more prompt with 3.0. We could ship as soon as Xcode 9 / Swift 4 ship this fall.

Sources/Flatten.swift
```diff
@@ -288,18 +288,13 @@ extension Signal where Value: SignalProducerConvertible, Error == Value.Error {
...
-			let disposable = CompositeDisposable()
-			let relayDisposable = CompositeDisposable()
-
-			disposable += relayDisposable
-			disposable += self.observeConcurrent(relayObserver, limit, relayDisposable)
-
-			return disposable
+		return Signal<Value.Value, Error> { relayObserver, lifetime in
+			let disposable = self.observeConcurrent(relayObserver, limit, lifetime)
+			_ = (disposable?.dispose).map(lifetime.observeEnded)
```
mdiep: I strongly dislike using `map` like this. `map` shouldn't be used for side effects. This is also not very clear IMO.

I think we're better off with `if let dispose = disposable?.dispose { lifetime.observeEnded(dispose) }`.

andersio: `lifetime += disposable` should work, and I have already replaced most of them.

mdiep:
This should be mentioned in the changelog, but otherwise it looks great. 👍

PR #405, opened 2017-05-23
Utilize the return value of `Set.insert`

ikesyo:


mdiep:
Weird. I did not know that was a thing!

Issue #395, opened 2017-05-18
Action compositions: wrapping one Action in another, fallback-if-disabled, etc


tikitu:
Hi folks,

I'm finding several cases where I'd like to "open up" the internals of `Action` so I can combine them in various ways. One case was in #366 ("given two actions, apply whichever of them is enabled"), and now I have another: wrap an existing Action, disabling it in more cases than the original.

I can solve that with a similar workaround to #366, but it's a bit hacky. To do it "properly" I think I would need to add to `Actions.swift`, to get access to the private internals there. Would y'all be interested in a pull request, once I've worked something out? Or is that a direction you'd rather not take the API in?

(For a first step I'm thinking just to add an initializer `Action.init(wrapping base: Action, enabledIf enabled: Property<Bool>`), with semantics that it's enabled if `enabled` *and* `base.isEnabled` are truthy. If I'm lucky, doing that might give me insight in more fundamental building-blocks that could be exposed for something like #366, but I'm not holding my breath.)

andersio:
I am not quite sure about initialiser vs operator at this point, but I do see the value of `Action` composition.

tikitu:
Good enough for me! I'll try to put together a PR this week: even if it's not accepted, I hope it will be useful to work through a concrete use case.

andersio:
I am personally interested in transforming and trivialising input, output and error, and I think these could help impose encapsulation. Let's see how we can push this forward.

RuiAAPeres:
Hello. 👋  Thanks for opening this issue. Due to inactivity, we will soft close the issue. If you feel that it should remain open, please let us know. 😄

PR #407, opened 2017-05-25
Action composition (WIP)

tikitu:
Some first attempts towards #395.

In its current state this PR is *not ready for merge*: it's a working-out of ideas to figure out what parts of the `Action` internals would need to be exposed to allow the kinds of composition I'm missing. If we go with it at all, I expect the implementation to change a lot, and probably the API also; hopefully the tests will still be useful.

### DIFF
--- a/Sources/Action.swift
+++ b/Sources/Action.swift
@@ -156,6 +156,18 @@ public final class Action<Input, Output, Error: Swift.Error> {
 			})
 	}
 
+	public convenience init(either first: Action<Input, Output, Error>, or second: Action<Input, Output, Error>) {
+		let bothStates = first.state.combineLatest(with: second.state)
+		let bothExecuting = first.isExecuting.and(second.isExecuting)
+		self.init(state: bothStates, enabledIf: { $0.0.isEnabled || $0.1.isEnabled }, isExecuting: bothExecuting, execute: { (states, input) in
+			if states.0.isEnabled {
+				return first.executeClosure(states.0.value, input)
+			} else {
+				return second.executeClosure(states.1.value, input)
+			}
+		 })
+	}
+
 	/// Initializes an `Action` that would always be enabled.
 	///
 	/// When the `Action` is asked to start the execution with an input value, a unit of
--- a/Sources/Action.swift
+++ b/Sources/Action.swift
@@ -86,11 +86,11 @@ public final class Action<Input, Output, Error: Swift.Error> {
 	///              executed by the `Action`.
 	public convenience init<State: PropertyProtocol>(state property: State, enabledIf isEnabled: @escaping (State.Value) -> Bool,
 										 execute: @escaping (State.Value, Input) -> SignalProducer<Output, Error>) {
-		self.init(state: property, enabledIf: isEnabled, isExecuting: { $0.isExecuting }, execute: execute)
+		self.init(state: property, enabledIf: isEnabled, isExecutingInner: nil, execute: execute)
 	}
 
 	fileprivate init<State: PropertyProtocol>(state property: State, enabledIf isEnabled: @escaping (State.Value) -> Bool,
-											 isExecuting: @escaping (ActionState) -> Bool,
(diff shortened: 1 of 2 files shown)


Sources/Action.swift
```diff
@@ -84,7 +84,14 @@ public final class Action<Input, Output, Error: Swift.Error> {
...
 	///   - execute: A closure that produces a unit of work, as `SignalProducer`, to be
 	///              executed by the `Action`.
-	public init<State: PropertyProtocol>(state property: State, enabledIf isEnabled: @escaping (State.Value) -> Bool, execute: @escaping (State.Value, Input) -> SignalProducer<Output, Error>) {
+	public convenience init<State: PropertyProtocol>(state property: State, enabledIf isEnabled: @escaping (State.Value) -> Bool,
+										 execute: @escaping (State.Value, Input) -> SignalProducer<Output, Error>) {
+		self.init(state: property, enabledIf: isEnabled, isExecuting: { $0.isExecuting }, execute: execute)
+	}
+
+	fileprivate init<State: PropertyProtocol>(state property: State, enabledIf isEnabled: @escaping (State.Value) -> Bool,
+											 isExecuting: @escaping (ActionState) -> Bool,
```
tikitu: The only piece of internal structure that I *had* to expose is hooking into `Action.isExecuting`. I think a tidier abstraction would expose more though.

tikitu:
@andersio I think this is ready for you to take a look: I don't expect it to merge as is, but I hope it prompts fruitful discussion.

I made use of the following internals that I couldn't have reached from an out-of-file extension:
* `Action.executeClosure` (to get `SignalProducer<Output, Error>` instead of `SignalProducer<Output, ActionError<Error>>`).
* `Action.state` to update the outer `Action` with any update to the inner ones.
* `ActionState.isEnabled`: this isn't strictly necessary, since one could use `Action.isEnabled` instead, but it's tidier to ask the inner states directly, since we're using them to synchronise updates.
* `ActionState.isExecuting` to ask whether the inner states are executing: I extended this to allow `outer.isExecuting` to depend directly on the inner states (so that running an inner `Action` can mark the outer one as executing) as well as on the outer `Action` being run directly. Again, reading this isn't strictly necessary (because we could use `Action.isExecuting`) but writing into it is.

tikitu:
Oops, and I only now see #400: definitely relevant, although I don't understand the consequences 100% yet.

andersio:
Yep. Note that #301 is also relevant. It does look interesting to me, esp. `Action(first:then:)`. But it would also be great if you may share how these fit into your needs, since [sometimes there could be alternatives that also make sense](https://github.com/ReactiveCocoa/ReactiveSwift/issues/309#issuecomment-290922797).

tikitu:
Funny, `Action(first:then:)` is the one I added only because it seems obvious to me (actions are function-like, so surely somebody will want to compose them?) -- I don't have a concrete use case myself.

My use case for `Action(either:or:)` is similar to #309; I'm guessing I could apply a similar refactoring to your suggestion there. I have a button that triggers either "add a favstar" or "remove the favstar": those were distinct `Action`s but I think combining them as you suggested in #309 will be an improvement: thanks!

The use case for `Action(enabledIf:wrapping:)` is: I already have an Action that does the right thing, but I want to disable it in more cases. This is a recurring pattern for me. I write a service like "make an authenticated network request and parse the result" as an `Action<URL, SomeOutputType, NetworkError>` which is enabled when it has authentication credentials. Then often I want to wrap some additional model-level or even view-level conditions around it: enabled only if some model object, or even some view-model, allows it. (Concrete example of the second: the underlying action is "send a message", and it's reused and wrapped into "resend a failed message" which is only enabled when the view-model records that a send attempt has already been made and has failed.) It's always possible to push this logic down into the underlying `Action`, of course, but I'd prefer to use small, compositional building blocks.

andersio:
Closing due to inactivity. 😣

PR #410, opened 2017-05-26
Prefer generic parameters to `Scheduler` existentials.

andersio:


sharplet:
What's the benefit of this change?

andersio:
No apparent benefit for now. Ideally this would allow the compiler to "see through" the interface and perform specialisation. But it doesn't because we do not have cross-module optimisation (yet).

Auto-promotion of existentials to generic parameters is one of the ABI stability items, btw.

sharplet:
Given that it makes the method signatures a little harder to read, I feel like I'm 👎 on this for now

mdiep:
> Given that it makes the method signatures a little harder to read, I feel like I'm 👎 on this for now

I agree. Let's do this when/if there's some benefit.

PR #412, opened 2017-05-26
s/(ActionDisposable|SimpleDisposable)/AnyDisposable/g

andersio:
1. Fold `ActionDisposable` and `SimpleDisposable` into `AnyDisposable`.

1. Move `+=` operator to the concrete types.

### DIFF
--- a/Sources/Disposable.swift
+++ b/Sources/Disposable.swift
@@ -53,6 +53,10 @@ public final class AnyDisposable: Disposable {
 			self.action = action
 		}
 
+		deinit {
+			state.deinitialize()
+		}
+
 		func dispose() {
 			if state.tryDispose() {
 				action?()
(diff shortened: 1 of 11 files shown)


Sources/Disposable.swift
```diff
@@ -40,68 +40,81 @@ extension UnsafeAtomicState where State == DisposableState {
...
-	public func dispose() {
-		disposable.dispose()
-	}
-}
-
-/// A disposable that only flips `isDisposed` upon disposal, and performs no other
-/// work.
-public final class SimpleDisposable: Disposable {
-	private var state = UnsafeAtomicState(DisposableState.active)
+	private let base: Backing
```
mdiep: I don't care for this implementation. I find this "multiple implementation via a backing enum" approach to be difficult to reason about.

If we kept `SimpleDisposable` and `ActionDisposable`, but made them `private` and added `init`s that exposed the functionality publicly, then I think I'd be onboard.

andersio: True.

PR #408, opened 2017-05-25
Rename `promoteErrors` to `promoteError`.

andersio:
Every `Signal` has only one error. Should it be `promote(_:)` though?

PR #413, opened 2017-05-26
`promoteError` type inference.

andersio:
Preceded by #408.

This is useful when combining multiple signals with `NoError` in the mix, reducing the verbosity.

```swift
let (signal, _) = Signal<(), NoError>.pipe()

print(type(of: signal.promoteError(CocoaError.self)))
print(type(of: signal.promoteError()))
print(type(of: signal.promoteError() as Signal<(), CocoaError>))

// The promotion of the second operand is inferred.
print(type(of: Signal.zip(signal.promoteError(CocoaError.self), signal.promoteError())))
```

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -2405,7 +2405,7 @@ extension Signal where Error == NoError {
 	///   - _ An `ErrorType`.
 	///
 	/// - returns: A signal that has an instantiatable `ErrorType`.
-	public func promoteError<F: Swift.Error>(_: F.Type) -> Signal<Value, F> {
+	public func promoteError<F: Swift.Error>(_: F.Type = F.self) -> Signal<Value, F> {
 		return Signal<Value, F> { observer in
 			return self.observe { event in
 				switch event {
@@ -2420,6 +2420,21 @@ extension Signal where Error == NoError {
 				}
 			}
 		}
+	}
+
+	/// Promote a signal that does not generate failures into one that can.
+	///
+	/// - note: This does not actually cause failures to be generated for the
+	///         given signal, but makes it easier to combine with other signals
+	///         that may fail; for example, with operators like
+	///         `combineLatestWith`, `zipWith`, `flatten`, etc.
+	///
+	/// - parameters:
+	///   - _ An `ErrorType`.
+	///
+	/// - returns: A signal that has an instantiatable `ErrorType`.
+	public func promoteError(_: Error.Type = Error.self) -> Signal<Value, Error> {
+		return self
 	}
 
 	/// Forward events from `self` until `interval`. Then if signal isn't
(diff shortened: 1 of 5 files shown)


Sources/Signal.swift
```diff
@@ -2405,7 +2405,7 @@ extension Signal where Error == NoError {
 	///   - _ An `ErrorType`.
 	///
 	/// - returns: A signal that has an instantiatable `ErrorType`.
-	public func promoteErrors<F: Swift.Error>(_: F.Type) -> Signal<Value, F> {
+	public func promoteError<F: Swift.Error>(_: F.Type = F.self) -> Signal<Value, F> {
```
NachoSoto: Awesome

Sources/Signal.swift
```diff
@@ -2422,6 +2422,21 @@ extension Signal where Error == NoError {
...
+	/// - note: This does not actually cause failures to be generated for the
+	///         given signal, but makes it easier to combine with other signals
+	///         that may fail; for example, with operators like
+	///         `combineLatestWith`, `zipWith`, `flatten`, etc.
+	///
+	/// - parameters:
+	///   - _ An `ErrorType`.
+	///
+	/// - returns: A signal that has an instantiatable `ErrorType`.
+	public func promoteError(_: Error.Type = Error.self) -> Signal<Value, Error> {
```
NachoSoto: Wait why is this overload necessary?

Sources/SignalProducer.swift
```diff
@@ -1397,8 +1397,23 @@ extension SignalProducer where Error == NoError {
...
+	///         given producer, but makes it easier to combine with other
+	///         producers that may fail; for example, with operators like
+	///         `combineLatestWith`, `zipWith`, `flatten`, etc.
+	///
+	/// - parameters:
+	///   - _ An `ErrorType`.
+	///
+	/// - returns: A producer that has an instantiatable `ErrorType`.
+	public func promoteError(_: Error.Type = Error.self) -> SignalProducer<Value, Error> {
+		return self
```
NachoSoto: Ditto

Sources/Signal.swift
andersio: It isn't necessary, just that it allows one (brainlessly) adds `promoteError` to everything, while it doesn't create extra `Signal`s in cases with the identical error types.

NachoSoto: I'm very much 👎 for that. Since when has RAC promoted brainless thinking? :P A user calling this operator when they shouldn't is likely an error and shouldn't be silenced IMO.

andersio: Perhaps I should be more careful about my wordings. Nothing is silenced here, and the overload does not enable anything new. It already works today, and the overload is just a fast path when it knows `Self.Error` is the `Error` of the return producer.

NachoSoto: Oooh that's right, because it's not possible to do negative conditions on extensions. This is a great idea then 👍

NachoSoto:
Can you add a test to cover this automatic inference?

PR #420, opened 2017-05-31
Fix bad productName values

liscio:
This (should be) a very minimally invasive PR. 

**The background:**

In general, Xcode appears to behave very badly when using workspaces and its "Find Implicit Dependencies" feature in the scheme editor. Many folks throw up their hands and resort to manual dependency specifications.

Well I think that its terrible behavior can be attributed to a bug in Xcode that causes it to not update the `productName` value in the `.pbxproj` file when a target is renamed. This happens very commonly when creating a framework project, then duplicating the framework targets with suffixes for other platforms.

For some folks, this patch may improve their build experience in Xcode quite a bit—especially for follks like me with very large workspaces that involve a lot of subprojects.

Sorry it took so long to write this one up after making the branch—I hadn't verified that the fix actually worked until I hit the issue again today, and fixing the `productName` value resolved my dependency issue. I'll try and cook up a Radar for this now that I know the steps.

### DIFF
--- a/ReactiveSwift.xcodeproj/project.pbxproj
+++ b/ReactiveSwift.xcodeproj/project.pbxproj
@@ -654,7 +654,7 @@
 			dependencies = (
 			);
 			name = "ReactiveSwift-tvOS";
-			productName = ReactiveCocoa;
+			productName = "ReactiveSwift-tvOS";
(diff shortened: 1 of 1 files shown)


mdiep:
🚢

PR #422, opened 2017-06-01
Swift 4

mdiep:
This gets `master` building with the Swift 4 snapshots.

Changes fall into 2 categories:

- Type inference failures
- Removal of implicit tuple destructuring from SE-0110

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: objective-c
 osx_image: xcode8.3
-before_install: 
+before_install:
   - brew uninstall carthage
   - HOMEBREW_NO_AUTO_UPDATE=1 brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/6ae4f69a652fb0ecb102b0c9216378679a4f1b92/Formula/carthage.rb # 0.22.0
 install: true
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -55,6 +55,15 @@ jobs:
         submodules: false
       env:
         - JOB=SWIFTPM_DARWIN
+    - before_install:
+        - eval "$(curl -sL https://gist.githubusercontent.com/kylef/5c0475ff02b7c7671d2a/raw/9f442512a46d7a2af7b850d65a7e9bd31edfb09b/swiftenv-install.sh)"
+      script:
+        - swift build
```
mdiep: Nimble doesn't build with Swift 4, so this only builds for now. 😞

andersio:
Can you add a job for the Swift 3 compatibility mode?

andersio:
(Not related to the PR) The implicit tuple destructing makes me wonder if we need to introduce sth like `combineLatestMap` or `zipMap`. 🤦‍♂️

mdiep:
> Could you add a job for the Swift 3 compatibility mode?

I'm not sure how to do that, so I'd love to leave it for a separate PR.

PR #427, opened 2017-06-04
Lazily initialize Action signals.

andersio:
Lazily initialise all derived signals in `Action`, so that for those who do not use them, it costs just an allocation of `Atomic<AnyObject? * 4>` w/ `nil` checking in the worker.

The implementation takes advantage of the nature & practicality of hot signals, in the sense that a `Signal` observation is substantial only if it is done before side effects that yield the anticipated results.

#### Checklist
- ~~Updated CHANGELOG.md.~~ (Implementation detail)


mdiep:
This would need to have conflicts resolved.

mdiep:
This isn't compiling. 🙈

PR #429, opened 2017-06-05
Warning for observation APIs when used with inhabitable types & `promoteValue`.

andersio:
Raise warnings when using observation API or operators that is essentially a logic error due to `Value` or `Error` being an inhabitable type. Please refer to `InhabitableTypeGuards.swift` for the lists of affected operators.

Also note that these overloads cannot be marked as `unavailable` or `obsolete`, or otherwise the regular version would take precedence over this.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/Sources/InhabitableTypeGuards.swift
+++ b/Sources/InhabitableTypeGuards.swift
@@ -1,0 +1,127 @@
+import Result
+
+// Observation
+extension SignalProducer where Value == Never {
+	@discardableResult
+	@available(*, deprecated, message:"`success` event is never delivered.")
+	public func startWithResult(_ action: @escaping (Result<Value, Error>) -> Void) -> Disposable { fatalError() }
+
+	@discardableResult
+	@available(*, deprecated, message:"The observer is never invoked.")
+	public func startWithValues(_ action: @escaping (Value) -> Void) -> Disposable { fatalError() }
+}
+
+extension SignalProducer where Value == Never, Error == NoError {
+	@discardableResult
+	@available(*, deprecated, message:"The observer is never invoked.")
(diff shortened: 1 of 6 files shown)


Sources/InhabitableTypeGuards.swift
```diff
@@ -0,0 +1,127 @@
...
+	SignalProducer<Never, NoError>.never.flatMapError { _ in SignalProducer<Never, NoError>.empty }
+
+	Signal<Never, AnyError>.never.flatMap(.latest) { _ in SignalProducer<Int, AnyError>.empty }
+	Signal<Never, AnyError>.never.flatMap(.latest) { _ in SignalProducer<Int, NoError>.empty }
+	Signal<Never, NoError>.never.flatMap(.latest) { _ in SignalProducer<Int, AnyError>.empty }
+	Signal<Never, NoError>.never.flatMap(.latest) { _ in SignalProducer<Int, NoError>.empty }
+	Signal<Never, NoError>.never.flatMapError { _ in SignalProducer<Never, AnyError>.empty }
+	Signal<Never, NoError>.never.flatMapError { _ in SignalProducer<Never, NoError>.empty }
+}
+*/
```
mdiep: What's this?

andersio: Just a bunch of test cases. Leaving them in the spec would flood the warnings. 🤦‍♂️

mdiep:
Just needs a changelog entry or two.

andersio:
I think it needs better error message to tell that ignoring the error message would trap at runtime & also fill in the blanks for `fatalError`.

andersio:
@mdiep Please check. The error message are written in reduced written register btw, following the swift stdlib style.

PR #430, opened 2017-06-06
Mitigate random failures in the MutableProperty async deadlock test and Action infinite feedback test.

andersio:
#### Checklist
- ~~Updated CHANGELOG.md.~~

### DIFF
--- a/Tests/ReactiveSwiftTests/ActionSpec.swift
+++ b/Tests/ReactiveSwiftTests/ActionSpec.swift
@@ -162,25 +162,15 @@ class ActionSpec: QuickSpec {
 						return .empty
 					}
 
-					let disposable = CompositeDisposable()
+					var count = 0
 
-					waitUntil(timeout: 0.01) { done in
-						let target = DispatchQueue(label: "test target queue")
-						let highPriority = QueueScheduler(qos: .userInitiated, targeting: target)
-						let lowPriority = QueueScheduler(qos: .default, targeting: target)
+					action.isExecuting.producer
+						.startWithValues { _ in
+							condition.value = 10
 
-						disposable += action.isExecuting.producer
-							.observe(on: highPriority)
-							.startWithValues { _ in
-								condition.value = 10
-							}
-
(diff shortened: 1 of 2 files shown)


mdiep:
It doesn't seem to have helped. 🙈

andersio:
Eww.

Tests/ReactiveSwiftTests/PropertySpec.swift
```diff
@@ -285,7 +285,7 @@ class PropertySpec: QuickSpec {
 
 					target! <~ source
 
-					queue.async(group: group) {
+					queue.async(group: group, flags: .barrier) {
```
andersio: It limits the concurrency, but IIRC does not affect the assertion here which was about the mitigated data race between the two blocks, not across iterations.

mdiep: I think it does:

> Instead, the queue waits until its currently executing blocks finish executing. At that point, the barrier block executes by itself. Any blocks submitted after the barrier block are not executed until the barrier block completes.

(https://developer.apple.com/documentation/dispatch/1452797-dispatch_barrier_async?language=objc)

I think you'd need a separate, empty barrier block.

Tests/ReactiveSwiftTests/ActionSpec.swift
```diff
@@ -162,25 +162,15 @@ class ActionSpec: QuickSpec {
...
-
-						disposable += lowPriority.schedule {
-							done()
+							count += 1
+							expect(count) == 1
 						}
-					}
-
-					disposable.dispose()
 				}
```
mdiep: Was this test also occasionally failing?

andersio: Yes.

Tests/ReactiveSwiftTests/PropertySpec.swift
andersio: Ugh, you are right.

PR #433, opened 2017-06-07
Address redundant conformance constraint warnings for Swift 4 compiler (Xcode 9)

ikesyo:
The changes should be compatible with Swift 3.1.

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -2414,7 +2414,7 @@ extension Signal where Error == NoError {
 	///   - _ An `ErrorType`.
 	///
 	/// - returns: A signal that has an instantiatable `ErrorType`.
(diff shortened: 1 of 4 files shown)


andersio:
IIRC I tried to remove them, but Swift 3.1 complains about it.

Sources/Signal.swift
```diff
@@ -2179,7 +2179,7 @@ extension Signal {
 		}
 	}
 
-	private convenience init<Strategy: SignalAggregateStrategy>(_ builder: AggregateBuilder<Strategy>, _ transform: @escaping (ContiguousArray<Any>) -> Value) {
+	private convenience init<Strategy>(_ builder: AggregateBuilder<Strategy>, _ transform: @escaping (ContiguousArray<Any>) -> Value) {
```
ikesyo: Hm only this doesn't work. 🤔

ikesyo:
Thank you! :shipit:

PR #437, opened 2017-06-07
Expanded CI builds

mdiep:
- Log the Swift version
- Test Swift 4 builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ jobs:
       env:
         - JOB=PODSPEC
     - script:
-      - swift --version
+        - swift --version
         - swift build
         - SWIFTPM_TEST_ReactiveSwift=YES swift test
       git:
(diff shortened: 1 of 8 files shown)


andersio:
https://github.com/apple/swift-evolution/blob/master/proposals/0151-package-manager-swift-language-compatibility-version.md

Looks like this is the way for a 3.2 build.

mdiep:
Yup. I just learned that in a WWDC lab. ☺️

.travis.yml
```diff
@@ -49,6 +49,7 @@ jobs:
       env:
         - JOB=PODSPEC
     - script:
+      - swift --version
```
sharplet: indentation looks off here

mdiep: Thanks! 😅

mdiep:
The tests need some more updates for Swift 4.

mdiep:
Ready!

andersio:
Instead of changing `apply()` to `apply(())`, we should fix it by adding an overload for `Input == Void`.

PR #439, opened 2017-06-09
No cancel token if the action is immediately consumed by UIScheduler.

andersio:
Change of an Implementation detail.

#### Checklist
- ~Updated CHANGELOG.md.~

### DIFF
--- a/Sources/Scheduler.swift
+++ b/Sources/Scheduler.swift
@@ -136,37 +136,44 @@ public final class UIScheduler: Scheduler {
 	///            begins.
 	@discardableResult
 	public func schedule(_ action: @escaping () -> Void) -> Disposable? {
-		let disposable = AnyDisposable()
-		let actionAndDecrement = {
-			if !disposable.isDisposed {
-				action()
-			}
-
-			#if os(Linux)
-				self.queueLength.modify { $0 -= 1 }
-			#else
-				OSAtomicDecrement32(self.queueLength)
-			#endif
-		}
-
-		#if os(Linux)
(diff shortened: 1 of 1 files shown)


Sources/Scheduler.swift
```diff
@@ -162,11 +159,17 @@ public final class UIScheduler: Scheduler {
...
 			actionAndDecrement()
+			return nil
 		} else {
-			DispatchQueue.main.async(execute: actionAndDecrement)
-		}
+			let disposable = AnyDisposable()
 
-		return disposable
+			DispatchQueue.main.async {
+				guard !disposable.isDisposed else { return }
```
sharplet: Is this a change in behaviour? Should there be a new test for this?

andersio: The only change is that `schedule` that is done immediately no longer returns a disposable, aligning to the semantic of `ImmediateScheduler`. Per the `Scheduler` contract, it is wrong to assume `schedule` would always return a cancellation token anyway.

andersio: Argh, you catch a flaw I've made.

sharplet: Specifically though, why was an explicit check of the disposable added?

andersio: It was there before.
https://github.com/ReactiveCocoa/ReactiveSwift/pull/439/commits/04b4ddf4500a2f93dcbef286ace2d899c0918621#diff-f045af08e7f58fd005590d376ccd1215L141

PR #443, opened 2017-06-10
Nesting the result and the output type into ValidatingProperty.

andersio:
#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/Sources/ValidatingProperty.swift
+++ b/Sources/ValidatingProperty.swift
@@ -23,7 +23,7 @@ public final class ValidatingProperty<Value, ValidationError: Swift.Error>: Muta
 	private let setter: (Value) -> Void
 
 	/// The result of the last attempted edit of the root property.
-	public let result: Property<ValidationResult<Value, ValidationError>>
+	public let result: Property<Result>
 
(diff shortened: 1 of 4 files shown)


andersio:
Note that `ValidatorOutput` is renamed to be `Decision`.

Sources/ValidatingProperty.swift
```diff
@@ -254,74 +254,74 @@ public final class ValidatingProperty<Value, ValidationError: Swift.Error>: Muta
...
-}
 
-/// Represents a decision of a validator of a validating property made on a
-/// proposed value.
-public enum ValidatorOutput<Value, Error: Swift.Error> {
-	/// The proposed value is valid.
-	case valid
+	/// Represents a decision of a validator of a validating property made on a
+	/// proposed value.
+	public enum Decision {
```
sharplet: How would you feel about `Status`?

andersio: It is a result of an aggressive action, while status is passive state of a matter.

mdiep:
I don't love `Decision`, but no better options come to mind.

Issue #322, opened 2017-04-13
Reactive collections


andersio:
Collections are essential to any application logic, so as collection views in any user experience. So at some point, we should consider merging ReactiveCollections into ReactiveSwift and ReactiveCocoa.

To achieve it within a reasonable time frame (Swift 4.0 and ReactiveSwift 3.0), it is proposed to scope the collection effort to the following goals:

- [ ] Define a collection snapshot & changeset format.
   #444 

- [ ] Collection diffing operator that computes deltas across versions of collections.
    #444

- [ ] Collection diffing operator that supports nested, two-level collections.
    #494 

- [ ] Minimum viable binding solution: changeset application.
    (TBD in ReactiveCocoa)

- [ ] Full fledged a collection binding API.
    (TBD in ReactiveCocoa)

With the defined abstractions and the API model, the project can be deemed ready to merge into ReactiveSwift and ReactiveCocoa. Feature parity can be improved on top of these works without significant design effort, but generally just implementation effort.

#### ReactiveCollections
[The ReactiveCollections project](https://github.com/RACCommunity/ReactiveCollections/) remains the evolving ground for dedicated, mutable containers that provide improved performance characteristics than the O(n) Heckel's diffing algorithm.

#### List of possible directions

- Live-updating filtered/sorted views.
- Collection aggregates (i.e. multiple collections presented as one).
- [RAC] Bindings for other AppKit and UIKit collection views.
- [RAC] CoreData and Realm integration.

/cc @ReactiveCocoa/reactiveswift @ReactiveCocoa/reactivecocoa

andersio:
Closing this issue as agreed in https://github.com/RACCommunity/ReactiveCollections/pull/38 that ReactiveSwift should have no part in the collection binding effort.

PR #444, opened 2017-06-10
Collection diffing operators.

andersio:
`diff()` is based on Paul Heckel's diffing algorithm, which is also used by IGListKit.

`diff()` can precisely detect inserts, removals, mutations and moves in O(n) time as long as all elements are unique. If there are repeated elements, `diff()` currently uniformly emits removals and inserts for every single diff until it becomes unique.

### Supported collections
`diff()` expects only collections that maintain a stable element order, e.g. `Array`. It supports:

| `Collection.Element` | Basis of comparison |
| ---- | --- |
| Any type | Custom comparator and identifier transform |
| `AnyObject` | Identified and compared by object identity |
| `AnyObject & Equatable` | Identified by object identity. Compared by value equality. Configurable. |
| `AnyObject & Hashable` | Identified by object identity. Compared by value equality. Configurable. |
| `Hashable`| Value equality |

For a glimpse in how deltas can be consumed: [the WIP `DataSource` in ReactiveCollections](https://github.com/RACCommunity/ReactiveCollections/pull/19/files#diff-03efcc672d299bd6721edd8814d8a1d4R218).

### Preconditions
1. the collection must exhibit array semantics, i.e. has a stable element order that is akin to arrays. By order the relative order is being referred to — indices may still be invalidated.

### How would the delta react to non-unique elements?
Non-unique elements resolving to the same identity would all be emitted as delete-insert pairs until the identity again becomes unique. The order is not guaranteed to be stable for these elements.

### Why doesn't the delta provide a snapshot for previous values?

For its premier use case of feeding the presentation layer, the historical values are generally irrelevant, and only the metadata of these values is useful.

Speaking of observers, since in practice most observers would need to special case the first delta for the full picture of the collection, caching the history on one's own is hardly a concern. There is also a tool at the stream level, i.e. `combinePrevious`, to help with this kind of uses.

Requiring such snapshot also poses additional costs from a composition perspective, where more information might need to be kept in memory in order to generate transformed delta that abides to the contract. #494 is a prime example — the operator would need to keep an extra version of the collection in memory, should it oblige to such requirement.

### Concerning ReactiveCollections
ReactiveCollections would remain as the ground to evolve standalone, reference-type collections that strike for the best performance with large data sets.

The collection view binding effort would be transferred to ReactiveCocoa once we settle on the collection delta format and `diff()` here.

### Future work
Composable sorting and filtering that works upon the continuous diffs.

### Checklist
- ~~Offer `diffIndex` that turns offsets into indices?~~
- [x] Initial test coverage.
- [x] Updated CHANGELOG.md.

/cc @ReactiveCocoa/reactiveswift


andersio:
WIP:
1. Revert the event enum type.
2. Make it work like `combinePrevious()` with the initial delta sent starting from the 2nd collection received.
3. Disable `diff()` for properties on Linux since the compiler is wonky.

andersio:
Note that `PropertyProtocol.diff` is disabled for Swift 3.2/4.0 due to a compiler bug with Xcode 8.3.3. It can be reenabled as soon as Xcode 9 is available.

andersio:
Working on adapting the diffing algorithm for two-level nested collections. Single level diff would be computed using `CollectionOfOne`.

Sources/CollectionDiffing.swift
```diff
@@ -0,0 +1,560 @@
...
+public struct Changeset {
+	/// Represents the context of a move operation applied to a collection.
+	public struct Move {
+		public let source: Int
+		public let isMutated: Bool
+
+		public init(source: Int, isMutated: Bool) {
+			(self.source, self.isMutated) = (source, isMutated)
+		}
+	}
```
mdiep: Any particular reason for choosing this representation?

The other possibility is to do something like this:

```swift
public enum Change {
    case insert(Int)
    case remove(Int)
    case move(from: Int, to: Int, mutated: Bool)
    case mutate(Int)
}
```

Not saying one is better than the other. But curious as to your rationale.

Tests/ReactiveSwiftTests/CollectionDiffingSpec.swift
```diff
@@ -0,0 +1,203 @@
...
+#if os(Linux)
+import Glibc
+#else
+import Darwin.C
+#endif
+
+private class ObjectValue {}
+
+class CollectionDiffingSpec: QuickSpec {
+	override func spec() {
```
mdiep: I would definitely want to see more tests for the actual diffing algorithm.

Sources/CollectionDiffing.swift
```diff
@@ -0,0 +1,560 @@
...
+			             areEqual: areEqual)
+		}
+	}
+
+	private static func _diff<View: Collection, Identifier: Hashable>(
+		previous: View,
+		current: View,
+		identifier: (Iterator.Element) -> Identifier,
+		areEqual: (Iterator.Element, Iterator.Element) -> Bool
+	) -> Changeset where View.Iterator.Element == Iterator.Element, View.Index == View.Indices.Iterator.Element {
```
mdiep: This could probably be a `Changeset.init`.

Moving all of that to a separate file—to let this file focus on the reactive portion—would probably be beneficial as well.

Sources/CollectionDiffing.swift
```diff
@@ -0,0 +1,560 @@
...
+		case is Array<Iterator.Element>.Type:
+			fallthrough
+		case is ContiguousArray<Iterator.Element>.Type:
+			// The standard library does not copy the content when `ContiguousArray.init`
+			// is passed an `Array` backed by a native buffer or a `ContiguousArray`. So
+			// we exploit this optimization to obtain an unsafe buffer pointer.
+			//
+			// `ArraySlice` is not supported at the moment, since it may have a non-zero
+			// start index that needs to be taken care of.
+			return ContiguousArray(previous).withUnsafeBufferPointer { previousBuffer in
```
mdiep: I don't love this. 🙈 

Have you measure this optimization? I wouldn't think that this would take up a significant portion of the runtime.

Sources/CollectionDiffing.swift
```diff
@@ -0,0 +1,560 @@
...
+
+/// Represents a snapshot of a collection.
+///
+/// The changeset associated with the first snapshot received by any given observation
+/// is generally irrelevant. Observations should special case the first snapshot as a
+/// complete replacement, or whatever semantic that fits their purpose.
+///
+/// `Snapshot` does not include a reference to the previous snapshot. If the
+/// history is relevant for delta consumption, it is up to the observers to cache the
+/// history on their own.
```
mdiep: Why not include that reference? With COW I wouldn't think it'd be a real performance concern.

Sources/CollectionDiffing.swift
```diff
@@ -0,0 +1,560 @@
...
+///
+/// While the protocol does not bind the `Changeset` associated type with any constraint,
+/// it is expected to be either a `Changeset`, or a composite of `Changeset`s for a nested
+/// collection.
```
mdiep: I recognize that you want this to be generic so that you can add support for nested collections, but I think it'd be better to remove this type parameter for the time being. I think it's usually best to wait to add something until you actually need it.

mdiep:
Having built something like this before (on the Obj-C APIs), I can say that it's definitely useful. So I'd definitely like to see something like this.

But I think it might be better to start this as an external project—much like Rex. This is a purely additive API. Including it here will force it to maintain the same source compatibility as the rest of the project. Starting it elsewhere would let these APIs mature at their own pace. Once it matured, it could then be moved into ReactiveSwift.

(I'd still be willing to review PRs on that codebase if desired.)

Sources/CollectionDiffing.swift
andersio: Yep that was removed during the iteration. But at this point I think it is okay to put it back in.

Sources/CollectionDiffing.swift
andersio: #494 uses it for its `SectionedChangeset`.

Sources/CollectionDiffing.swift
andersio: This heterogeneous representation is not a not nice format to consume. There is an implicit order between all these operations, and say if you are interested in only a specific kind of operations at a time, you are forced to go over all of them again and again.

Moreover, these information can be (are) all prepared as standalone sets upfront.

Last but not least, while it has an memory overhead for small # of changes (4 references + 2 Ints per change), it is more compact when there is a lot of contiguous changes which `IndexSet` is optimised for.

https://github.com/ReactiveCocoa/ReactiveSwift/pull/494/files#diff-4a097c76f0bff02c4c19b2d40ecbc652R125

Sources/CollectionDiffing.swift
andersio: Not yet. :p

andersio:
@mdiep Given the scale of our community and my personal experience of ReactiveCollections, I don't think it is worth evolving separately. If we are not going to commit to it yet, we can either declare these as experimental or keeping a feature branch given their additive nature.

Moreover, the collection diffing itself is not that large as a feature to iterate on.

#322 proposed a series of small steps to achieve this very goal of collection support.

Tests/ReactiveSwiftTests/CollectionDiffingSpec.swift
andersio: Agreed.

andersio:
ICYMI: https://github.com/ReactiveCocoa/ReactiveSwift/pull/444#discussion_r129304837

mdiep:
I'll continue to review this, but it may take a little time for me to review and understand the difference pieces.

andersio:
> The problem with a feature branch is that it's hard for people to use it. So you can't get the same level of feedback about the APIs. Frameworks benefit from real world usage because it informs the design of the API.

Neither would a split help IMO. No spotlight, low volume, and most users are not going to use something that isn't mature. Let alone making contribution. That's what ReactiveCollections ends up to be.

We can always make staging releases alongside the mainline releases and pre-releases, which effectively gives you the same result.

> It's not that I'm worried about distraction. If we want to make a breaking change to collection diffing, do we (1) wait for a breaking release of ReactiveSwift or (2) make the next release a major version so that the collection diffing can have a breaking change?
>
> Neither of those is a great option. (1) means that these APIs evolve more slowly than they could otherwise. (2) makes upgrading harder for everyone who uses the framework.

You are basing this on top of the assumption that every merge must mean to be part of a release.

andersio:
To be transferred back to ReactiveCollections.

andersio:
https://github.com/RACCommunity/ReactiveCollections/pull/34

PR #445, opened 2017-06-11
A variant of `combinePrevious` that doesn't need an initial value.

andersio:
Note that `combinePrevious()` is not implemented using default arguments and optionals, so that existing use cases as concerned by @NachoSoto would not break.

In other words, `combinePrevious(nil)` and `combinePrevious()` are not equivalent, given the constraint `<U> Value == Optional<U>`.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # master
 *Please add new entries at the top.*
 
+1. `combinePrevious` for `Signal` and `SignalProducer` no longer requires an initial value. The first tuple would be emitted as soon as the second value is received by the operator if no initial value is given. (#445, kudos to @andersio)
+
 1. In Swift 3.2 or later, you may create `BindingTarget` for a key path of a specific object. (#440, kudos to @andersio)
 
 # 2.0.0-alpha.2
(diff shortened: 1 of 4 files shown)


andersio:
🤔
Haven't considered optionals. That said we can use this just for `combinePrevious()`, and keeping the original `combinePrevious(initial:)` so that `combinePrevious()` != `combinePrevious(nil)`. Nothing would be break in this case.

> since this can be replicated with skip(first: 1)
Not quite if you consider how many extra hops needed for `(U?, U?)` to `(U, U)` (or probably just a `filterMap`, but still).

sharplet:
For reference, some previous discussion on this general idea: https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2632

andersio:
@sharplet Thanks for pointing out the previous discussion. Some responses to @NachoSoto 

> One of the combinePrevious implementation is not implemented in terms of the other.

Both share the same implementation in this PR.

> This operator is redundant: the existing combinePrevious is more general. You can accomplish the same thing by doing something like producer.combinePrevious(nil).skip(1).

It is more general, but from `Signal<U, E>` to precisely `Signal<(U, U), E>` with the current `combinePrevious`, one has to jump through multiple hops:

```swift
values // Signal<U, E>
    .optionalize() // Signal<U?, E>
    .combinePrevious(nil) // Signal<(U?, U?), E>
    .filterMap { $0.0 != nil && $0.1 != nil ? ($0.0!, $0.1!) : nil } // Signal<(U, U), E>
```

which is hardly sane. `combinePrevious()` with optional initial value is actually the more general version.

PR #446, opened 2017-06-11
Convenience `Action.apply` that avoids `ActionError`.

andersio:
#### Checklist
- [x] Updated CHANGELOG.md.


mdiep:
What's the motivation behind this? This is a reasonable thing to do, but without a strong use case, I'd rather not expand the API.

PR #451, opened 2017-06-14
Inlined lock storage in `Lock`.

andersio:
Initialise `PthreadLock` and `UnfairLock` using `ManagedBufferPointer`, having the lock as the tail allocated header.

Unlike using `&` on a stored property, the location is guaranteed to be the same. :)

#### Checklist
- ~~Updated CHANGELOG.md.~~

### DIFF
--- a/Sources/Atomic.swift
+++ b/Sources/Atomic.swift
@@ -107,80 +107,114 @@ internal struct UnsafeAtomicState<State: RawRepresentable> where State.RawValue 
 /// `Lock` exposes `os_unfair_lock` on supported platforms, with pthread mutex as the
 // fallback.
 internal class Lock {
+	// Both `UnfairLock` and `PthreadLock` use `ManagedBufferPointer` to allocate inline
+	// storage for the lock. Inout reference to a stored property is deliberately avoided
+	// because Swift does not make any guarantee on the memory location of a variable. It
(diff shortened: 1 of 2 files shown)


Sources/Atomic.swift
```diff
@@ -107,80 +107,114 @@ internal struct UnsafeAtomicState<State: RawRepresentable> where State.RawValue
 /// `Lock` exposes `os_unfair_lock` on supported platforms, with pthread mutex as the
 // fallback.
 internal class Lock {
+	// Both `UnfairLock` and `PthreadLock` use `ManagedBufferPointer` to allocate inline
```
mdiep: Why use `ManagedBufferPointer` instead of an `UnsafeMutablePointer`? This seems more complex.

andersio: One less level of pointer chasing at runtime. Since the header is part of the class allocation `ManagedBufferPointer`, it can be resolved at static time as an offset to `self`.

mdiep: It's the "abuse" of `ManagedBufferPointer` here that concerns me. I'd like to see us pursue the more standard approach first and only switch to a more complex, unintended approach when/if we have benchmarks that show that show a measurable difference.

andersio:
No performance difference is measured.

PR #452, opened 2017-06-14
Wrap `Signal.Core.state` with `Atomic`.

andersio:
#### Checklist
- ~~Updated CHANGELOG.md.~~


andersio:
Not ready yet... There is a mysterious race that was once fixed with the current lock-free design that I cannot wrap my head around.

PR #454, opened 2017-06-14
swiftpm test dependency updates

sharplet:
There are two independent changes here:

 1. Use `NSString.boolValue` in `Package.swift` to parse the
    `SWIFTPM_TEST_ReactiveSwift` environment variable for a truthy value. This
    is cross-platform, and handles more truthy values like `true` and `1`, which is
    nice.

 2. Use a more general `SWIFT_PACKAGE_TEST` environment variable. I introduced
    this in https://github.com/sharplet/Regex/pull/53, and like the idea of
    having a more standard variable name that can be used across different
    projects.

I think (1) is a nice improvement, and would love some input on (2) as a potential standard variable name. What do you think?

- N/A: ~Updated CHANGELOG.md.~

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -2,7 +2,9 @@ import Foundation
 import PackageDescription
 
 var isSwiftPackagerManagerTest: Bool {
-    return ProcessInfo.processInfo.environment["SWIFTPM_TEST_ReactiveSwift"] == "YES"
+    let environment = ProcessInfo.processInfo.environment
+    guard let value = environment["SWIFTPM_TEST_ReactiveSwift"] else { return false }
+    return NSString(string: value).boolValue
 }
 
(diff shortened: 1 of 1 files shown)


ikesyo:
> SWIFT_PACKAGE_TEST

:-1: for that. I intentionally used a project-specific variable to avoid pulling transitive depencies' test dependencies.

sharplet:
Right, transitive dependencies, of course. What a bummer.

mdiep:
👍 for (1) though!

mdiep:
Somehow I missed that this was updated to only do (1). 🚢

sharplet:
Oh, thanks for reviewing this again! It was on my list, but I apparently also forget that I already updated this. 😅

PR #456, opened 2017-06-15
Improve `Signal` state management to avoid an ARC race.

andersio:
Preceded by #453.

### What is the race in particular?
Let's look at how the compiled assembly does the work:
```swift
Thread A                           | Thread B
------------------------------------------------------------------
Lock `updateLock`.                 |
                                   |
Pack the new enum.                 |
(with retain)                      |
                                   |
Load the current `self.state`.     | Load `self.state`.
                                   |
Store to `self.state`.             | Retain the enum.
                                   |
Release the old enum.              | Call the observers.
                                   |
Unlock `updateLock`.               |
```
It seems safe, especially since writes are serialised.

But what if we shift the timeline of thread B a little bit further?
```swift
Thread A                           | Thread B
------------------------------------------------------------------
Lock `updateLock`.                 |
                                   |
Pack the new enum.                 |
(with retain)                      |
                                   |
Store to `self.state`.         <<!!!!!>> Load `self.state`.
                                   |
Release the old enum.          <<!!!!!>> Retain the enum.
                                   |
Unlock `updateLock`.               | Call the observers.
```
Data races now exist due to ARC not being part of the atomic read/write. Our RCU model considers only the atomicity of `self.state` alone without the hidden side effect of ARC.

While the contention window is extremely small (a few instruction cycles) thus being extremely hard to expose, it is possible on paper that the `AliveState` is released before a concurrent relaxed reader retains it.

### What is the solution proposed by this PR?
~~The Signal core now tracks the number of pending state updates.~~

~~If a sender detects that there is any pending state update, it would acquire also `updateLock` just to exploit the serialisation to eliminate any release-retain race with the state updates.~~

`Signal.Core.send` locks `updateLock` whenever it accesses `state`.

### What is the cost of the solution?
~~It requires an `Int32` read for every `value` event delivered, which is offset by another change in the patch. It also requires atomic increments & decrements before and after state updates.~~

### What else is changed in this PR?
1. ~~Since Swift retains the enum as it loads it, even if only the tag is probed, the relaxed terminating check is also prone to the race.~~

    ~~The terminating check now relies instead on a special negative value in the new atomic counter, which would be asserted as the state is bumped to `terminating`.~~

1. ~~All the state management is encapsulated in `CoreBlackBox`, which `Signal.Core` now inherits from.~~

1. Recursive terminations are no longer prioritized, and are now prone to competition with concurrent senders (if any). This is only a change in the implementation detail, as the `Signal` contract does not guarantee any deterministic order between concurrent senders.

    ~~Note that the removal offsets the event delivery overhead brought by the atomic counter.~~

#### Checklist
- ~Updated CHANGELOG.md.~

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -108,16 +108,11 @@ public final class Signal<Value, Error: Swift.Error> {
 		/// Used to ensure that events are serialized during delivery to observers.
 		private let sendLock: Lock
 
-		/// Used to indicate if the `Signal` has deinitialized.
-		private var hasDeinitialized: Bool
-
 		fileprivate init(_ generator: (Observer) -> Disposable?) {
-			state = .alive(AliveState(observers: Bag()))
+			state = .alive(Bag(), hasDeinitialized: false)
 
 			updateLock = Lock.make()
 			sendLock = Lock.make()
-
-			hasDeinitialized = false
 			disposable = SerialDisposable()
 
 			// The generator observer retains the `Signal` core.
@@ -147,9 +142,8 @@ public final class Signal<Value, Error: Swift.Error> {
 
 				self.updateLock.lock()
 
-				if case let .alive(state) = self.state {
-					let newSnapshot = TerminatingState(observers: state.observers, event: event)
-					self.state = .terminating(newSnapshot)
+				if case let .alive(observers, _) = state {
+					self.state = .terminating(observers, .init(event))
 					self.updateLock.unlock()
 
 					// Check whether the terminating state has been handled by a
@@ -184,10 +178,10 @@ public final class Signal<Value, Error: Swift.Error> {
 				self.sendLock.lock()
 				self.updateLock.lock()
 
-				if case let .alive(state) = self.state {
+				if case let .alive(observers, _) = self.state {
 					self.updateLock.unlock()
 
-					for observer in state.observers {
+					for observer in observers {
 						observer.action(event)
 					}
 				} else {
@@ -227,10 +221,10 @@ public final class Signal<Value, Error: Swift.Error> {
 
 			updateLock.lock()
 
-			if case let .alive(state) = state {
-				var observers = state.observers
-				token = observers.insert(observer)
-				self.state = .alive(AliveState(observers: observers))
+			if case let .alive(observers, hasDeinitialized) = state {
+				var newObservers = observers
+				token = newObservers.insert(observer)
+				self.state = .alive(newObservers, hasDeinitialized: hasDeinitialized)
 			}
 
 			updateLock.unlock()
@@ -252,10 +246,10 @@ public final class Signal<Value, Error: Swift.Error> {
 		private func removeObserver(with token: Bag<Observer>.Token) {
 			updateLock.lock()
 
-			if case let .alive(state) = state {
-				var observers = state.observers
-				let observer = observers.remove(using: token)
-				self.state = .alive(AliveState(observers: observers))
+			if case let .alive(observers, hasDeinitialized) = state {
+				var newObservers = observers
+				let observer = newObservers.remove(using: token)
+				self.state = .alive(newObservers, hasDeinitialized: hasDeinitialized)
 
 				var result = OperationResult.none
 
@@ -305,12 +299,12 @@ public final class Signal<Value, Error: Swift.Error> {
 				// handled, take it over and bump the status to `terminated`.
 				self.updateLock.lock()
 
-				if case let .terminating(state) = self.state {
+				if case let .terminating(observers, terminationKind) = self.state {
 					self.state = .terminated
 					self.updateLock.unlock()
 
-					if let event = state.event {
-						for observer in state.observers {
+					if let event = terminationKind.materialize() {
+						for observer in observers {
 							observer.action(event)
 						}
 					}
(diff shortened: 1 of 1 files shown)


Sources/Signal.swift
```diff
@@ -391,6 +302,137 @@ public final class Signal<Value, Error: Swift.Error> {
...
+		final func updateState<R>(_ body: (inout State) -> R) -> R {
+			var count: Int32
+
+			repeat {
+				count = updaterCount.value
+				guard count >= 0 else {
+					assert(count == -1)
+					break
+				}
+			} while !updaterCount.swap(from: updaterCount.value, to: updaterCount.value + 1)
```
mdiep: This seems like a poor imitation of a spin/unfair lock. I'd prefer to see us use `Atomic` like in #453—at least to reestablish correctness. Afterwards we could consider an optimization like this.

andersio: This is not a lock, just a custom CAS loop for increment/decrement that takes care of the special `-1` case.

andersio: > I'd prefer to see us use Atomic like in #453—at least to reestablish correctness. Afterwards we could consider an optimization like this.

To be fair, the Signal core itself is already an `Atomic`, just with the storage and the lock inlined. If you prefer not to have this optimisation, the `Signal.CoreBlackBox` can drop it without affecting any use site.

andersio:
Speed stat:
```
// this branch
@test(): avg 1725 ns; min 1449 ns

// master
@test(): avg 1260 ns; min 1039 ns

// #453
@test(): avg 2376 ns; min 2045 ns
```

andersio:
It turns out that locking on the hot path with https://github.com/ReactiveCocoa/ReactiveSwift/pull/456/commits/1ad25242a10f355805b369fb707374f5023070cb is less harmful than un-inlining the lock and the storage.

🤦‍♂️

```swift
@test(): avg 1350 ns; min 1104 ns
```

andersio:
The removal of these two reference types push the overhead back to the level of `master`.

```
@test(): avg 1276 ns; min 1045 ns
@test(): avg 1249 ns; min 1043 ns
@test(): avg 1250 ns; min 1043 ns
@test(): avg 1229 ns; min 1043 ns
@test(): avg 1269 ns; min 1039 ns
@test(): avg 1244 ns; min 1051 ns
```

Sources/Signal.swift
```diff
@@ -108,16 +108,11 @@ public final class Signal<Value, Error: Swift.Error> {
 		/// Used to ensure that events are serialized during delivery to observers.
 		private let sendLock: Lock
 
-		/// Used to indicate if the `Signal` has deinitialized.
-		private var hasDeinitialized: Bool
-
 		fileprivate init(_ generator: (Observer) -> Disposable?) {
```
mdiep: The above documentation needs to be updated to reflect that the lock must be held to read the state.

andersio:
_(Fun fact: Swift 4 runs everything slower.)_

Sources/Signal.swift
```diff
@@ -127,28 +119,20 @@ public final class Signal<Value, Error: Swift.Error> {
...
-
+			// Check if the status has been bumped to `terminating` due to a
+			// terminal event being sent concurrently or recursively.
+			//
+			// The check is deliberately made outside of the `sendLock` so that it
+			// covers also any potential concurrent terminal event in one shot.
+			//
+			// Related PR:
+			// https://github.com/ReactiveCocoa/ReactiveSwift/pull/112
+			if .shouldDispose == self.tryToCommitTermination() {
```
andersio: Both the terminal path and the value path run this at the end.

Sources/Signal.swift
```diff
@@ -127,28 +119,20 @@ public final class Signal<Value, Error: Swift.Error> {
...
-				//
-				// The check is deliberately made outside of the `sendLock` so that it
-				// covers also any potential concurrent terminal event in one shot.
-				//
-				// Related PR:
-				// https://github.com/ReactiveCocoa/ReactiveSwift/pull/112
-				self.stateLock.lock()
-				if result == .none, case .terminating = self.state {
-					self.stateLock.unlock()
-					result = self.tryToCommitTermination()
```
andersio: `tryToCommitTermination` already locks the state to check if it is terminating or not. So the locking here is pointless.

Sources/Signal.swift
```diff
@@ -119,20 +119,26 @@ public final class Signal<Value, Error: Swift.Error> {
...
+				//
+				// The check is deliberately made outside of the `sendLock` so that it
+				// covers also any potential concurrent terminal event in one shot.
+				//
+				// Related PR:
+				// https://github.com/ReactiveCocoa/ReactiveSwift/pull/112
+				//
+				// While calling `tryToCommitTermination` is sufficient, this is a fast
+				// path for the recurring value delivery.
+				if stateLock.try() {
```
andersio: This is resurrected because (trying to) acquiring the spinlock is faster than calling `tryToCommitTermination()`. 🤦‍♂️

Sources/Signal.swift
```diff
@@ -119,20 +119,28 @@ public final class Signal<Value, Error: Swift.Error> {
...
+				// Related PR:
+				// https://github.com/ReactiveCocoa/ReactiveSwift/pull/112
+				//
+				// While calling `tryToCommitTermination` is sufficient, this is a fast
+				// path for the recurring value delivery.
+				//
+				// Note that this cannot be `try` since any concurrent observer bag
+				// manipulation might then cause the terminating state being missed.
+				stateLock.lock()
+				if case .terminating = state {
```
andersio: This is resurrected because acquiring the spinlock is consistently & considerably faster than calling `tryToCommitTermination()` on my machine (>50ns per call). 🤦‍♂️

PR #459, opened 2017-06-17
Extensible FlattenStrategy.

andersio:
In case we might want to add new flatten strategies in the future without a major version bump, `FlattenStrategy` now hides its enum nature. This on paper breaks switches over `FlattenStrategy`, but I don't see any real use case that would be affected.

#### Checklist
- ~~Updated CHANGELOG.md.~~

### DIFF
--- a/Sources/Flatten.swift
+++ b/Sources/Flatten.swift
@@ -9,20 +9,32 @@
 import enum Result.NoError
 
 /// Describes how multiple producers should be joined together.
-public enum FlattenStrategy: Equatable {
+public struct FlattenStrategy {
+	fileprivate enum Kind {
+		case concurrent(limit: UInt)
+		case latest
+		case race
+	}
+
+	fileprivate let kind: Kind
+
+	private init(kind: Kind) {
+		self.kind = kind
+	}
+
 	/// The producers should be merged, so that any value received on any of the
 	/// input producers will be forwarded immediately to the output producer.
 	///
 	/// The resulting producer will complete only when all inputs have
 	/// completed.
-	public static let merge = FlattenStrategy.concurrent(limit: .max)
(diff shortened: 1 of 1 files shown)


Sources/Flatten.swift
```diff
@@ -53,16 +73,21 @@ public enum FlattenStrategy: Equatable {
...
 			return true
 
 		case (.concurrent(let leftLimit), .concurrent(let rightLimit)):
 			return leftLimit == rightLimit
 
+		case (.race, .race):
+			return true
+
 		default:
 			return false
```
NachoSoto: Not saying I agree with this change, but if we merge it, it doesn’t make sense to keep the `Equatable` conformance.

Sources/Flatten.swift
```diff
@@ -9,20 +9,36 @@
...
+		self.base = base
+	}
+
 	/// The producers should be merged, so that any value received on any of the
 	/// input producers will be forwarded immediately to the output producer.
 	///
 	/// The resulting producer will complete only when all inputs have
 	/// completed.
-	public static let merge = FlattenStrategy.concurrent(limit: .max)
+	public static var merge: FlattenStrategy {
```
NachoSoto: This doesn’t need to be a derived property, why not keep a constant?

NachoSoto:
Can you elaborate on the “without a major version bump” thing? Is that in case users are `switch`ing over it?

andersio:
Yes, that's the reason @mdiep blocked #233 from merging into 1.1.x.

NachoSoto:
Cool then I’m on board after those 2 comments

mdiep:
I'll review soon, but this is a great idea. 💯

Issue #461, opened 2017-06-20
Auto-interrupt `Signal`.


andersio:
In 2.0 alpha 3, we have automatic silent disposal of `Signal` when its input (generator) observer deinitializes _and_ the `Signal` is not retained, regardless of still having observers or not.

Despite having this new semantic over 1.x, there is still a hole left for leaks. Specifically, a signal would be kept alive if it is retained, even if the input observer has deinitialized. (recall: the new semantic works if both — not either — conditions are satisfied.)

Whether or not active observers should be notified of the disposal is also debatable.

So I'd propose to have `Signal`s automatically & unconditionally interrupted by the deinitialization of the input observer (unless a terminal event has been sent, of course). With this, `Signal` would now be cleaned up by either ends completely cease the interests to send/observe.

@ReactiveCocoa/reactiveswift

mdiep:
That makes sense to me. 👍

andersio:
@mdiep Sadly, it would be still kind of a leak (the signal and the core), but at least the observers and disposables are released.

mdiep:
I wouldn't call it a leak if something is holding on to the signal.

andersio:
Implemented.

PR #463, opened 2017-06-20
Send `interrupted` when an input observer deinitializes.

andersio:
Implements #461.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/Sources/Observer.swift
+++ b/Sources/Observer.swift
@@ -15,6 +15,21 @@ extension Signal {
 		/// An action that will be performed upon arrival of the event.
 		public let action: Action
 
+		/// Whether the observer should send an `interrupted` event as it deinitializes.
+		private let interruptsOnDeinit: Bool
+
+		/// An initializer that accepts a closure accepting an event for the
+		/// observer.
+		///
+		/// - parameters:
+		///   - action: A closure to lift over received event.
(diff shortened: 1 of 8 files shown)


Sources/Observer.swift
```diff
@@ -68,6 +84,15 @@ extension Signal {
 			}
 		}
 
+		deinit {
+			if interruptsOnDeinit {
+				// Since `Signal` would ensure that only one terminal event would ever be
+				// sent for any given `Signal`, we do not need to assert any condition
+				// here.
+				action(.interrupted)
```
mdiep: Would it mess up anything if we always sent interrupted on `deinit`? I'm wondering if that'd be a better option.

andersio: It would mess up the promise of no event being sent after any terminal event, unless we track termimation per observer.

Detaching an observer could use an `interrupted` event though.

mdiep: The promise is only at the Signal level though. That wouldn't be broken.

andersio: IMO `Observer` is part of the `Signal` contract. It also breaks the useful assumption of terminal events being sent at most once, that would potentially break things e.g. cleaning up resource twice, triggering cancellation code after successful completion.

PR #471, opened 2017-06-30
Fixed a deadlock upon disposal in AggregateBuilder.

### LINKED ISSUE #470: Crash in Lock.UnfairLock.lock()
I found a locking-related crash while updating our project to ReactiveSwift 2.0.0-rc.1.

Managed to reduce it down to this test case, which reproduces the problem in this project as well:

```swift
it("should not cause EXC_BAD_INSTRUCTION") {
	let property1 = MutableProperty<Int?>(nil)
	let property2 = MutableProperty<Int?>(nil)

	SignalProducer
		.zip(property1.producer, property2.producer)
		.take(first: 1)
		.startWithValues { _ in }
}
```

Haven't had a chance to dig too deep on this yet beyond reproducing it.

andersio:
Fixes #470.
Removed also a stale test case.

The deadlock happens when a `value` event of a combined signal triggers termination of one of its source signals (except for the one that indirectly yields the `value` event).

Example: https://github.com/ReactiveCocoa/ReactiveSwift/pull/471/files#diff-8c46f2726f3ee5a6dbf7ad24b29dbd57R2795

The value from `B` yields a new `(A, B)` tuple, leading to termination of `A` and eventually informing the strategy with the completion of `A`. Since the delivery of the tuple is in progress, the strategy's internal lock is still acquired, resulting in the deadlock.

Threading of the signal aggregate strategies now mirrors that of `Signal.Core` to allow recursive termial event sent by a `value` event.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # master
 *Please add new entries at the top.*
 
+1. Fixed a deadlock upon disposal when combining operators, i.e. `zip` and `combineLatest`, are used. (#471, kudos to @stevebrambilla for catching the bug)
+
 # 2.0.0-rc.1
 1. If the input observer of a `Signal` deinitializes while the `Signal` has not yet terminated, an `interrupted` event would now be automatically sent. (#463, kudos to @andersio)
 
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -1959,42 +1959,48 @@ private enum ThrottleWhileState<Value> {
 	}
 }
 
-private protocol SignalAggregateStrategy {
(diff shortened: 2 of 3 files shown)


andersio:
(not ready)

andersio:
```
// Threading of `CombineLatestStrategy` and `ZipStrategy`.
//
// The threading model of these strategies mirror that of `Signal.Core` to allow
// recursive termial event sent by a `value` event. Specifically, completion status
// is tracked separately with its own lock to permit recursive completion, even if the
// event delivery lock is still acquired.
```

PR #473, opened 2017-07-05
Action input transformation & additional availability conditions.

andersio:
Related: #407 

Transforming `Action` input has become a recurring pain point for me to carry an `Action` across layers of abstractions.

`Action` yields `SignalProducer<Output, ActionError<E>>`, and without access to the internals, simple things like creating an `Action` mapping all inputs are blatantly fragile with data race etc.

```swift
// `outerAction` can race with `innerAction.isEnabled`.
outerAction = Action(enabledIf: innerAction.isEnabled) { input in
    // How to deal with `ActionError<InnerError>`?
    // Without the new initialiser, either we pollute the interface of `outerAction`,
    // ignore the `disabled` error, or transform it as `interrupted`.
    return innerAction.apply(transform(input)).????
}
```

This PR proposes a new form of initialisers:
1. `Action(inner:execute:)`;
1. `Action(inner:enabledIf:)`; and
1. `Action(inner:enabledIf:execute:)`.

From a high-level perspective, the outer `Action` pipes its output to the inner `Action` as its input.

These wrapping `Action`s do not maintain their own serialisation and state, and instead rely on the inner `Action` for execution.

To achieve this, the `Action` internal has been changed to accept deferred input in the form of `SignalProducer`. For example, `apply(_:)` has been changed to:
```swift
// `apply(_:)`: synchronous input.
return execute(self, SignalProducer(value: input))

// `init(inner:enabledIf:execute:)`: The input can be asynchronous.
self.execute = { action, input in
	// ...
	return inner.execute(inner, input.take(first: 1).flatMap(.race, execute))
}
```

In other words, the actor of execution attempts on the outer `Action`s is always the inner `Action`, regardless how many levels are nested. Moreover, the outer `Action`s are basically passing through the availability & execution state of the inner `Action`, except for those being applied with additional conditions.

The call site would look like this:
```swift
outerAction = Action(innerAction) { outerInput in
    return SignalProducer(value: getInnerInput(outerInput))
}
```

#### Use cases
Encapsulation/Layering

```swift
update = Action(userService.update) { (newAccountDetails: AccountDetailsViewModel) in
    return SignalProducer(value: buildRequestModel(for: newAccountDetails))
}
```

Abstraction/Conversion

```swift
let section = ...
hasSelected = Action(viewModel.hasSelected) { (relativeIndex: Int) in
    return SignalProducer(value: absoluteIndex(from: relativeIndex, for: section))
}
```

#### Checklist
- [ ] `take(atLeast:)` operator so that if a `Signal` would interrupt upon completion if the source doesn't yield at least one value.
- [ ] Unit Tests
- [ ] Updated CHANGELOG.md.


Sources/Action.swift
```diff
@@ -168,6 +171,74 @@ public final class Action<Input, Output, Error: Swift.Error> {
...
+	/// Create an `Action` that wraps an inner `Action` with an additional availability
+	/// condition applied.
+	///
+	/// - parameters:
+	///   - inner: The inner `Action`.
+	///   - isEnabled: An optional additional `Action` availability condition.
+	public convenience init<Enabled: PropertyProtocol>(
+		inner: Action<Input, Output, Error>,
+		enabledIf isEnabled: Enabled
+	) where Enabled.Value == Bool {
```
mdiep: The general idea makes sense, but this API may need some refinement. If nothing else, `inner` should probably have no name externally. But I'm not totally convinced that these should be `init`s.

andersio:
Not being actively worked on.

Issue #421, opened 2017-05-31
Operators for binding `Disposable` to a `Lifetime`.


andersio:
In #334, we replaced `CompositeDisposable` with `Lifetime` for collecting cleanup callbacks in producers, and `+=` is also deprecated. So for now adding a `Disposable` to `Lifetime` has to be written as:

```swift
// Two-liner
let disposable = self.start(observer)
lifetime.observeEnded(disposable.dispose)

// One-liner
lifetime.observeEnded(self.start(observer).dispose)

// Previously
disposable += self.start(observer)
```

If #404 is to be merged, the lack of an operator causes it even nastier for `Signal`s.

```swift
// Two-liner
let disposable = self.observe(observer) // Optional
_ = (disposable?.dispose).map(lifetime.observeEnded)

if let disposable = self.observe(observer) {
    lifetime.observeEnded(disposable.dispose)
}

// Previously
disposable += self.observe(observer)

// Or
return self.observe(observer)
```

It makes me wonder if we can reintroduce an operator shorthand, but instead of `+=` we would reuse `<~`.
```swift
disposable <~ lifetime

self.observe { event in
    // (event manipulation)
} <~ lifetime
```

This makes sense if we treat `Disposable` as a binding target of `()`. But on the other hand the LHS/RHS of the operator has to be swapped as compared to `+=`.

mdiep:
Moving `lifetime` to the rhs does't seem to help with readability either.

I'll keep this in mind when I review #404 and then we can discuss more. We should definitely do something to improve things if we're going to be merge that PR.

ikesyo:
Coming from https://github.com/Carthage/ReactiveTask/pull/97 and 👍 for some sort of operator or something like `Disposable.bind(to: Lifetime)`.

ikesyo:
> `lifetime.observeEnded(self.start(observer).dispose)`

I think this will not do the right thing––unexpectedly entire expression will be executed when a liftime ended.

mdiep:
I'm on board with an operator. This suggestion from @andersio seems like it might make sense:

```swift
disposable <~ lifetime
```

ikesyo:
This will be difficult to read / understand and will not be natural:

```swift
stderrAggregated
    .then(stdoutAggregated)
    .map(TaskEvent.success)
    .start(observer) <~ lifetime
```

The following (placing a lifetime on LHS) is preferable for me (as a natural replacement of `+=` operator):

```swift
[...]

PR #475, opened 2017-07-06
[POC] Add convenient `Disposable.dispose(whenEnded lifetime: Lifetime)`

ikesyo:
https://github.com/ReactiveCocoa/ReactiveSwift/issues/421#issuecomment-312890841

I'd like to have feedbacks.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # master
 *Please add new entries at the top.*
 
+1. Added a convenient `dispose(whenEnded:)` to `Disposable`. (#475, kudos to @ikesyo)
+
 # 2.0.0-rc.2
 1. Fixed a deadlock upon disposal when combining operators, i.e. `zip` and `combineLatest`, are used. (#471, kudos to @stevebrambilla for catching the bug)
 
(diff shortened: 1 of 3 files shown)


andersio:
It looks natural, but `.bind(to:)` is kinda misleading in the context. I would suggest `dispose(when:)`.

ikesyo:
Hmm using the same word `dispose` may be a bit confusing:

```swift
producer.start().dispose()
// versus
producer.start().dispose(when: lifetime)
```

andersio:
I think it is a reasonable mental model. If you don't specify anything, it disposes immediately. If you specify `whenEnded`, it disposes of at a certain point in the future.

ikesyo:
`whenEnded` sounds good (over just `when`).

sharplet:
What about `dispose(atEndOf: lifetime)`?

mdiep:
Do you think this is still needed now that `+=` is back?

ikesyo:
I'm happy with `+=` so withdraw this.

PR #478, opened 2017-07-09
Action input view.

andersio:
There are circumstances that, for the benefit of encapsulation and API compactness, only the subset of input & statuses APIs is needed for control binding purposes.

For example, a form VM may own an `Action` that would be shared by its cell VMs. The form VM would handle the result and the error uniformly on behalf of the cells, so the only thing the cell VMs need would be the essentials for control bindings.

So an `ActionInput<Value>` wrapper is proposed here. Instead of a producer of outputs, it yields a producer of a single flag indicating the successfulness of the execution attempt.

#### Checklist
- [ ] Updated CHANGELOG.md.

### DIFF
--- a/Sources/Action.swift
+++ b/Sources/Action.swift
@@ -70,6 +70,10 @@ public final class Action<Input, Output, Error: Swift.Error> {
 	/// Whether the action is currently enabled.
 	public let isEnabled: Property<Bool>
 
+	public var input: ActionInput<Input> {
+		return ActionInput(self)
+	}
+
 	/// Initializes an `Action` that would be conditionally enabled depending on its
 	/// state.
 	///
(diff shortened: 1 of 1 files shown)


mdiep:
I'm 👎 on this. I don't think provides much value. The cells in your example can ignore the uninteresting properties; I don't think having access to them has any negative effect.

andersio:
It is not as simple as ignoring. If I need to expose the full `Action`, the cells would need to be parameterised with `Output` and `Error`. This is the thing I'd like to avoid.

mdiep:
Could you provide a code example of how you're using it?

andersio:
Say given a cell view model:
```swift
final class ActionCellViewModel: FormComponent {
        let isPressed: ActionInput<()>
}
```

with a form view model:
```swift
final class ProfileViewModel {
    enum Event {
        case showConfirmation(TransactionDetails)
        case showAlert(AlertDescription)
    }

    let events: Signal<Event, NoError>
    let eventsObserver: Signal<Event, NoError>.Observer
    let formComponents: [FormComponent]
    let submit: Action<(), Event, NoError>

    init() {
        submit = Action(...) { ... }

        // Pipe the action results to the VM event bus.
        submit.values.observe(eventsObserver)

        // Define the form.
        formComponents = [ActionCellViewModel(isPressed: submit.input)]
    }
}
```

Without `ActionInput`, currently we are forced to define `Action<(), (), NoError>` for anything bound with UI controls, or wrap one as such.

But in the later case, `ActionError.disabled` of the inner `Action` is just dropped when it isn't supposed to be, because of the limitation of the `Action` API (or otherwise it has to be `Action<(), (), ActionError<NoError>>`).

mdiep:
I'm not convinced that this is the best solution for connecting an actual to a cell view model. 😕 But I'd need to spend some time playing with a complete example to have a more informed opinion.

andersio:
To put it in another way, this is a `BindingTarget` that can tell if it is enabled or not.

andersio:
Close it for now. I assume the ability to ignore output (`Never`) and transforming error would be a more generalised solution. In the end, what I'd like to achieve is that any arbitrary `Action` can be represented as `Action<(), Never, NoError>` for the purpose of binding. 🤔

PR #479, opened 2017-07-09
Include the unavailable reason in `ActionError.disabled`.

andersio:
This frees retrying logic with `Action`s from the need of composing `isExecuting` and `isEnabled`. More importantly, it allows the observer to differentiate between temporary and potentially permanent unavailability. `isEnabled` does not convey such information.

### Impact
[The breaking bit affects only those who instantiates `ActionError.disabled`, presumably in unit tests](https://github.com/ReactiveCocoa/ReactiveSwift/pull/479/files#diff-cb74afb7f7cae7399eddbb00da5ae227L102). Usage in patterns is not affected.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # master
 *Please add new entries at the top.*
 
+1. `ActionError.disabled` now includes the reason why the `Action` refused the execution attempt. (#479, kudos to @andersio)
+
 # 2.0.0-rc.2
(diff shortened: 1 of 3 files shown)


mdiep:
I'm not sure I see the value in this. I don't see why it'd be useful.

andersio:
I just realised while this was on my queue, I had managed to find some other sensible ways to avoid retrying `Action`s. Close it for now. 🤦‍♂️

PR #486, opened 2017-07-16
Optimized constant `SignalProducer`s.

andersio:
These speedups are applicable to only closure-based observation APIs (i.e. `startWithSignal` is ineligible).

| Item | testSequence (32 items) | testValue |
| --- | --- | --- |
| `master` | avg 30296 ns; min 26426 ns | avg 8910 ns; min 8171 ns |
| `unsafeAction` | avg 16092 ns; min 14627 ns | avg 2928 ns; min 2568 ns |
| +`unsafeUninterruptibleAction` | avg 16140 ns; min 14502 ns | avg 664 ns; min 587 ns |
| Speedup | ~1.85x | ~13.5x |

[Benchmark source](https://gist.github.com/andersio/8d519b366aace3e85063ef1a2cbb38b1).

This PR also relaxed `UnsafeAtomicState.is`.

#### Checklist
- [ ] Updated CHANGELOG.md.

PR #487, opened 2017-07-17
Cutting the SignalProducer overhead further.

andersio:
### Goals
Taking advantage of the deferred nature of `SignalProducer` to further reduce the overhead.

### TL;DR
1. Composing `SignalProducer` is going to be _way, way_ cheaper than composing `Signal`.

1. Constant `SignalProducer`s, e.g. `init(value:)` and sequence `init(_:)`, are going to do _significantly_ less work.

1. There is still a ~16x overhead over standard library `lazy` for purely collection manipulation. But given the type erasing and reactive (passive) nature of FRP/ReactiveSwift, there is probably very little stuff we can do about it. IOW optimizing for this is a non-goal. :)

yeah

The actual speedup is subject to the new internal API adoption — unmigrated `SignalProducer` operators would still be relying on the conventional behavior and/or `lift`.

For example, if you do:

```swift
SignalProducer(Array(repeating: 1, count: 1000)).skipRepeats()
```

The performance remains the same as `master`, because `skipRepeats` hasn't been migrated yet.

### Introducing `SignalProducerCore`
`SignalProducerCore` is the new internal abstraction of `SignalProducer`. It has subsumed three major responsibilities from `SignalProducer`:

1. ~~the actual implementation of `startWithSignal`;~~
1. the actual implementation of `start` which accepts an `Signal.Observer`; and
1. the builder closure, now as `Core.make()`, which produces a `Signal` and the relevant context for composition and `startWithSignal`.

`SignalProducerCore` provides default implementations of all these capabilities for its subclasses, except for `Core.make()`.

_The cores are not nested in `SignalProducer` because the Swift 3.1 compiler would get stuck. The Swift 4 compiler has resolved the issue though._

#### `SignalCore`
This is the conventional `SignalProducer` with the `Signal` operator lifting optimisation in ReactiveSwift 2.0 (#140).

#### `EventTransformingCore`
`EventTransformingCore` composes _event transforms_, and is exposed as a new operator `Core.flatMapEvent` (internal for now).

It takes advantage of the deferred, single-observer nature of `SignalProducer`. For example, when we do:

```swift
upstream.map(transform).filterMap(filteringTransform).start()
```

It is contractually guaranteed that these operators would always end up producing a chain of streams, each with a **single and persistent** observer to its upstream. The multicasting & detaching capabilities of `Signal` is useless in these scenarios.

So `EventTransformingCore` builds on top of this very fact, and composes directly at the level of event transforms. This change implicitly nullifies #129, since producers no longer use `Signal` unless it absolutely has to.

It is intended for all synchronous `SignalProducer` operators. To encourage code reusing, the pilot operators have their implementation moved under `Signal.Event`, and `Signal` has gained `flatMapEvent` to use these `Event`-level operator implementations.

This represents a new type of `SignalProducer` that **does not use `Signal`** unless it is requested via `make()`.

##### Example
```swift
extension Signal {
	public func filterMap<U>(_ transform: @escaping (Value) -> U?) -> Signal<U, Error> {
		return flatMapEvent(Signal.Event. filterMap(transform))
	}
}

extension SignalProducer {
	public func filterMap<U>(_ transform: @escaping (Value) -> U?) -> SignalProducer<U, Error> {
		return core.flatMapEvent(Signal.Event. filterMap(transform))
	}
}
``` 

#### `EventGeneratingCore`
`EventGeneratingCore` wraps a generator closure that would be invoked when started. All the generator closure have are the observer and the cancel disposable.

It is intended for the constant `SignalProducer`s, as in #486, which synchronously emits all events.

It represents a new type of `SignalProducer` that **does not use `Signal`** unless it is requested via `make()`.

### Pilot public APIs

1. `map`, `mapError`, `filter` and `filterMap`.

1. Various `SignalProducer.init` overloads that accept a constant.

### Results

| Test | Sequence<br>(32 items) | Value | SequenceMapFilter |
| --- | --- | --- | --- |
| `master` | avg 30296 ns<br>min 26426 ns | avg 8910 ns<br>min 8171 ns | avg 61555 ns<br>min 50378 ns |
| This PR | avg 14073 ns<br>min 13079 ns | avg 839 ns<br>min 739 ns | avg 8304 ns<br>min 7198 ns | 
| Speedup | ~2.15x | ~10.5x | ~7.5x |
| `master`<br>WMO | avg 29333 ns<br>min 26136 ns | avg 8613 ns<br>min 7169 ns  | 31340 ns<br>min 27497 ns  |
| This PR<br>WMO | avg 13591 ns<br>min 11382 ns | avg 388 ns<br>min 325 ns | avg 5923 ns<br>min 5002 ns |
| Speedup<br >WMO | ~2.15x | ~22.00x | ~5.29x |

[Benchmark Source](https://gist.github.com/andersio/1aae47e6e9173b5dc6e8609356668ec1)

#### Checklist
- [ ] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # master
 *Please add new entries at the top.*
 
+1. The `SignalProducer` internals have undergone a significant refactoring, which bootstraps the effort to reduce the overhead of constant producers and producer compositions. (#487, kudos to @andersio)
+
 1. New method ``retry(upTo:interval:on:)``. This delays retrying on failure by `interval` until hitting the `upTo` limitation.
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,6 @@
 
 1. New method ``retry(upTo:interval:on:)``. This delays retrying on failure by `interval` until hitting the `upTo` limitation.
 
-1. Addressed the exceptionally high build time. (#495)
-
 # 2.0.0
 # 2.0.0-rc.3
 1. `Lifetime.+=` which ties a `Disposable` to a `Lifetime`, is now part of the public API and is no longer deprecated.
--- a/Sources/Disposable.swift
+++ b/Sources/Disposable.swift
@@ -38,6 +38,13 @@ extension UnsafeAtomicState where State == DisposableState {
 	}
 }
 
+internal final class NopDisposable: Disposable {
+	static let shared = NopDisposable()
+	var isDisposed = false
+	func dispose() {}
+	private init() {}
+}
+
 /// A type-erased disposable that forwards operations to an underlying disposable.
 public final class AnyDisposable: Disposable {
 	private final class ActionDisposable: Disposable {
--- a/Sources/Disposable.swift
+++ b/Sources/Disposable.swift
@@ -38,9 +38,27 @@ extension UnsafeAtomicState where State == DisposableState {
 	}
 }
 
+/// A disposable that does not have side effect upon disposal.
+internal final class _SimpleDisposable: Disposable {
+	private let state = UnsafeAtomicState<DisposableState>(.active)
+
+	var isDisposed: Bool {
+		return state.is(.disposed)
+	}
+
+	func dispose() {
+		_ = state.tryDispose()
+	}
+
+	deinit {
+		state.deinitialize()
+	}
+}
+
+/// A disposable that has already been disposed.
 internal final class NopDisposable: Disposable {
 	static let shared = NopDisposable()
-	var isDisposed = false
+	var isDisposed = true
 	func dispose() {}
 	private init() {}
 }
--- a/Sources/Observer.swift
+++ b/Sources/Observer.swift
@@ -18,6 +18,24 @@ extension Signal {
 		/// Whether the observer should send an `interrupted` event as it deinitializes.
 		private let interruptsOnDeinit: Bool
 
+		/// The target observer of `self`.
+		private let wrapped: AnyObject?
+
+		/// An initializer that transforms the action of the given observer with the
+		/// given transform.
+		///
+		/// If the given observer would perform side effect on deinitialization, the
+		/// created observer would retain it.
+		///
+		/// - parameters:
+		///   - observer: The observer to transform.
+		///   - transform: The transform.
+		internal init<U, E: Swift.Error>(_ observer: Signal<U, E>.Observer, _ transform: @escaping (@escaping Signal<U, E>.Observer.Action) -> Action) {
+			self.action = transform(observer.action)
+			self.wrapped = observer.interruptsOnDeinit ? observer : nil
+			self.interruptsOnDeinit = false
+		}
+
 		/// An initializer that accepts a closure accepting an event for the
 		/// observer.
 		///
@@ -27,6 +45,7 @@ extension Signal {
 		///                         event as it deinitializes. `false` otherwise.
 		internal init(action: @escaping Action, interruptsOnDeinit: Bool) {
 			self.action = action
+			self.wrapped = nil
 			self.interruptsOnDeinit = interruptsOnDeinit
 		}
 
@@ -37,6 +56,7 @@ extension Signal {
 		///   - action: A closure to lift over received event.
 		public init(_ action: @escaping Action) {
 			self.action = action
+			self.wrapped = nil
 			self.interruptsOnDeinit = false
 		}
 
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -19,27 +19,7 @@ import Result
 public struct SignalProducer<Value, Error: Swift.Error> {
 	public typealias ProducedSignal = Signal<Value, Error>
 
-	/// Wraps a closure which, when invoked, produces a new instance of `Signal`, a
-	/// customized `observerDidSetup` post-creation side effect for the `Signal` and a
-	/// disposable to interrupt the produced `Signal`.
-	///
-	/// Unlike the safe `startWithSignal(_:)` API, `builder` shifts the responsibility of
-	/// invoking the post-creation side effect to the caller, while it takes from the
-	/// caller the responsibility of the `Signal` creation.
-	///
-	/// The design allows producer lifting to be as efficient as native `Signal`
-	/// operators, by eliminating the unnecessary relay `Signal` imposed by the old
-	/// `startWithSignal(_:)`, regardless of the fact that lifted operators can rely on
-	/// the upstreams for producer interruption.
-	///
-	/// `observerDidSetup` must be invoked before any other post-creation side effect.
-	fileprivate struct Instance {
-		let producedSignal: Signal<Value, Error>
-		let observerDidSetup: () -> Void
-		let interruptHandle: Disposable
-	}
-
-	fileprivate let builder: () -> Instance
+	fileprivate let core: Core
 
 	/// Convert an entity into its equivalent representation as `SignalProducer`.
 	///
@@ -77,23 +57,22 @@ public struct SignalProducer<Value, Error: Swift.Error> {
 	/// - parameters:
 	///   - startHandler: A closure that accepts observer and a disposable.
 	public init(_ startHandler: @escaping (Signal<Value, Error>.Observer, Lifetime) -> Void) {
-		self.init { () -> Instance in
+		core = SignalCore {
 			let disposable = CompositeDisposable()
 			let (signal, observer) = Signal<Value, Error>.pipe(disposable: disposable)
 			let observerDidSetup = { startHandler(observer, Lifetime(disposable)) }
 			let interruptHandle = AnyDisposable(observer.sendInterrupted)
 
-			return Instance(producedSignal: signal, observerDidSetup: observerDidSetup, interruptHandle: interruptHandle)
+			return Core.Product(signal: signal, observerDidSetup: observerDidSetup, interruptHandle: interruptHandle)
 		}
 	}
 
-	/// Create a SignalProducer that will invoke the given factory once for each
-	/// invocation of `start()`.
+	/// Create a SignalProducer.
 	///
 	/// - parameters:
-	///   - builder: A builder that is used by `startWithSignal` to create new `Signal`s.
-	fileprivate init(_ builder: @escaping () -> Instance) {
-		self.builder = builder
+	///   - core: The `SignalProducer` core.
+	internal init(_ core: Core) {
+		self.core = core
 	}
 
 	/// Creates a producer for a `Signal` that will immediately send one value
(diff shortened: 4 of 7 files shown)


andersio:
Heh, the compiler gets stuck.

Sources/SignalProducer.swift
```diff
@@ -229,13 +203,124 @@ public struct SignalProducer<Value, Error: Swift.Error> {
 	///            `Signal` commences. Both the produced `Signal` and an interrupt handle
 	///            of the signal would be passed to the closure.
 	public func startWithSignal(_ setup: (_ signal: Signal<Value, Error>, _ interruptHandle: Disposable) -> Void) {
-		let instance = builder()
-		setup(instance.producedSignal, instance.interruptHandle)
+		return core.startWithSignal(setup)
+	}
+}
+
+internal class SignalProducerCore<Value, Error: Swift.Error> {
```
mdiep: Can we make this a `protocol` instead of a `class`? I have an aversion to inheritance.

andersio: Since there is no generalised existential, we would need a type-erased wrapper if we use protocol.

andersio: Instead of inheritance, it can also be implemented with an enum backing (what Foundation `Data` uses).

https://github.com/apple/swift/blob/master/stdlib/public/SDK/Foundation/Data.swift

Sources/SignalProducer.swift
```diff
@@ -19,27 +19,7 @@ import Result
...
-	///
-	/// `observerDidSetup` must be invoked before any other post-creation side effect.
-	fileprivate struct Instance {
-		let producedSignal: Signal<Value, Error>
-		let observerDidSetup: () -> Void
-		let interruptHandle: Disposable
-	}
-
-	fileprivate let builder: () -> Instance
+	fileprivate let core: SignalProducerCore<Value, Error>
```
mdiep: Can you add some documentation for this?

Sources/SignalProducer.swift
mdiep: I think an enum backing would be preferable. 👍

Sources/SignalProducer.swift
```diff
@@ -229,13 +203,134 @@ public struct SignalProducer<Value, Error: Swift.Error> {
...
+	/// invoking the post-creation side effect to the caller, while it takes from the
+	/// caller the responsibility of the `Signal` creation.
+	///
+	/// The design allows producer lifting to be as efficient as native `Signal`
+	/// operators, by eliminating the unnecessary relay `Signal` imposed by the old
+	/// `startWithSignal(_:)`, regardless of the fact that lifted operators can rely on
+	/// the upstreams for producer interruption.
+	///
+	/// `observerDidSetup` must be invoked before any other post-creation side effect.
+	struct Product {
```
mdiep: I really dislike the name _Product_ here.  Could we come up with something more specific?

Sources/SignalProducer.swift
```diff
@@ -229,13 +203,134 @@ public struct SignalProducer<Value, Error: Swift.Error> {
...
+		return core.startWithSignal(setup)
+	}
+}
+
+internal class SignalProducerCore<Value, Error: Swift.Error> {
+	/// Wraps a closure which, when invoked, produces a new instance of `Signal`, a
+	/// customized `observerDidSetup` post-creation side effect for the `Signal` and a
+	/// disposable to interrupt the produced `Signal`.
+	///
+	/// Unlike the safe `startWithSignal(_:)` API, `builder` shifts the responsibility of
```
mdiep: `builder` here needs to be updated

Sources/SignalProducer.swift
andersio: Hmm, I just realised that `EventTransformingCore` relies on the inheritance to erase two generic parameters. So unless we ditch it, an enum isn't quite sufficient. :thinking:

Sources/Disposable.swift
```diff
@@ -38,6 +38,13 @@ extension UnsafeAtomicState where State == DisposableState {
 	}
 }
 
+internal final class NopDisposable: Disposable {
+	static let shared = NopDisposable()
+	var isDisposed = false
+	func dispose() {}
+	private init() {}
+}
```
mdiep: This breaks the contract of `Disposable` by not changing `isDisposed` when you call `dispose`. It may be fine in practice, but this makes me very wary.

I think we'd be better off bringing back the implementation of `SimpleDisposable` and using that instead.

Sources/SignalProducer.swift
```diff
@@ -229,10 +211,149 @@ public struct SignalProducer<Value, Error: Swift.Error> {
...
+/// compositional and single-observer use cases.
+internal class SignalProducerCore<Value, Error: Swift.Error> {
+	/// `ProducedSignalReceipt` represents an instance of `Signal` created from a
+	/// `SignalProducer`. In addition to the `Signal` itself, it includes also the
+	/// starting side effect and an interrupt handle for this particular instance.
+	///
+	/// It is the responsibility of the `ProducedSignalReceipt` consumer to ensure the
+	/// starting side effect is invoked exactly once, and is invoked after observations
+	/// has properly setup.
+	struct ProducedSignalReceipt {
```
mdiep: How about calling this `Instance`?

Sources/SignalProducer.swift
```diff
@@ -471,13 +585,33 @@ extension SignalProducer {
...
+	}
+
+	private func lift<U, F, V, G>(leftFirst: Bool, _ transform: @escaping (Signal<Value, Error>) -> (Signal<U, F>) -> Signal<V, G>) -> (SignalProducer<U, F>) -> SignalProducer<V, G> {
+		return { otherProducer in
+			return SignalProducer<V, G>(SignalCore {
+				let left = self.core.make()
+				let right = otherProducer.core.make()
+
+				return .init(signal: transform(left.signal)(right.signal),
+				             observerDidSetup: {
```
mdiep: Could we put all the arguments on their own line? 🙈 

```swift
return .init(
    signal: transform(left.signal),
    observerDidSetup: {
        ...
    },
    ...
)
```

It's minor, but I think it has readability benefits. (Including outdenting the block body.)

Sources/Observer.swift
```diff
@@ -18,6 +18,24 @@ extension Signal {
 		/// Whether the observer should send an `interrupted` event as it deinitializes.
 		private let interruptsOnDeinit: Bool
 
+		/// The target observer of `self`.
+		private let wrapped: AnyObject?
```
mdiep: I don't love adding more to `Observer`. 😕

Sources/SignalProducer.swift
```diff
@@ -77,23 +63,24 @@ public struct SignalProducer<Value, Error: Swift.Error> {
 	/// - parameters:
 	///   - startHandler: A closure that accepts observer and a disposable.
 	public init(_ startHandler: @escaping (Signal<Value, Error>.Observer, Lifetime) -> Void) {
-		self.init { () -> Instance in
+		core = SignalCore {
```
mdiep: It's small, but the other `init`s case `self.init(<core instance>)`.

Can we update this one or the others to make them consistent?

Sources/SignalProducer.swift
```diff
@@ -229,10 +211,149 @@ public struct SignalProducer<Value, Error: Swift.Error> {
...
+	/// It is the responsibility of the `ProducedSignalReceipt` consumer to ensure the
+	/// starting side effect is invoked exactly once, and is invoked after observations
+	/// has properly setup.
+	struct ProducedSignalReceipt {
+		let signal: Signal<Value, Error>
+		let observerDidSetup: () -> Void
+		let interruptHandle: Disposable
+	}
+
+	func make() -> ProducedSignalReceipt {
```
mdiep: I think this should be `makeInstance()` (or whatever we call the struct).

Sources/SignalProducer.swift
```diff
@@ -229,10 +211,149 @@ public struct SignalProducer<Value, Error: Swift.Error> {
...
+	}
+}
+
+/// `EventGeneratingCore` wraps a generator closure that would be invoked upon a produced
+/// `Signal` when started. The generator closure is passed only the input observer and the
+/// cancel disposable.
+///
+/// It is intended for constant `SignalProducers`s that synchronously emits all events
+/// without escaping the `Observer`.
+private final class EventGeneratingCore<Value, Error: Swift.Error>: SignalProducerCore<Value, Error> {
```
mdiep: We could probably just call this `GeneratorCore`

Sources/SignalProducer.swift
```diff
@@ -229,10 +211,149 @@ public struct SignalProducer<Value, Error: Swift.Error> {
...
+///
+/// It is contractually guaranteed that these operators would always end up producing a
+/// chain of streams, each with a _single and persistent_ observer to its upstream. The
+/// multicasting & detaching capabilities of Signal is useless in these scenarios.
+///
+/// So EventTransformingCore builds on top of this very fact, and composes directly at the
+/// level of event transforms, without any `Signal` in between.
+///
+/// - note: This core does not use Signal unless it is requested via `make()`.
+private final class EventTransformingCore<Value, Error: Swift.Error, SourceValue, SourceError: Swift.Error>: SignalProducerCore<Value, Error> {
```
mdiep: We could probably call this `TransformerCore` or `FlatMapCore`.

mdiep:
This is looking good overall. 👍 👍

Sources/Disposable.swift
andersio: Can we treat it as disposed then? It still makes sense.

mdiep: Why not bring back `SimpleDisposable`? This still strikes me as a little weird.

andersio: This is a shared disposable for constant producers that cannot be disposed of. When I iterate on the design, benchmark shows allocation contributes considerable overhead.

andersio:
@mdiep 

For `SignalProducer(value:)`, the allocation of a non-atomic `SimpleDisposable` contributes 29% overhead.

```
@testValue(): avg 546 ns; min 475 ns
@testValue(): avg 388 ns; min 325 ns
```

Sources/Observer.swift
andersio: Sadly, we have to retain stuff marked with `interruptsOnDeinit`, and retaining it using a thunk adds an overhead to all calls to the observer, as opposed to an 8 byte storage of a reference. :/

```swift
// @testEventTransformingCoreMapFilter(): avg 6091 ns; min 5250 ns
self.action = transform(observer.action)
self.wrapped = observer

// @testEventTransformingCoreMapFilter(): avg 6406 ns; min 5428 ns
self.action = transform { observer.action($0) }
```

andersio:
The cost of sequence is slightly up with the implementation being corrected in https://github.com/ReactiveCocoa/ReactiveSwift/pull/487/commits/56131fc38db512af012515c80c750d5308cc582f, but the difference is trivial given its scale.

All constant producers except the sequence one still uses `NopDisposable`, which is now always disposed.

```
@testEventTransformingCoreMapFilter(): avg 6091 ns; min 5250 ns
@testSequence(): avg 13277 ns; min 11720 ns
@testValue(): avg 369 ns; min 317 ns
```

Sources/SignalProducer.swift
```diff
@@ -229,10 +213,155 @@ public struct SignalProducer<Value, Error: Swift.Error> {
 	///            `Signal` commences. Both the produced `Signal` and an interrupt handle
 	///            of the signal would be passed to the closure.
 	public func startWithSignal(_ setup: (_ signal: Signal<Value, Error>, _ interruptHandle: Disposable) -> Void) {
-		let instance = builder()
-		setup(instance.producedSignal, instance.interruptHandle)
-		guard !instance.interruptHandle.isDisposed else { return }
+		let receipt = core.makeInstance()
```
mdiep: `receipt` should be `instance` here

CHANGELOG.md
```diff
@@ -1,11 +1,15 @@
...
 *Please add new entries at the top.*
 
+1. The `SignalProducer` internals have undergone a significant refactoring, which bootstraps the effort to reduce the overhead of constant producers and producer compositions. (#487, kudos to @andersio)
+
 # 2.0.1
 1. Addressed the exceptionally high build time. (#495)
 
 1. New method ``retry(upTo:interval:on:)``. This delays retrying on failure by `interval` until hitting the `upTo` limitation.
 
+1. Addressed the exceptionally high build time. (#495)
```
mdiep: This line is duplicated in the 2.0.1 notes. 🙈

mdiep:
🎉

PR #488, opened 2017-07-17
Correct some Linux-specific build issues

liscio:
While working on a server-side project I noticed that neither of the latest 1.1.x or 2.x releases would build for me on Linux with the latest 4.x snapshots (2017-07-11 and 2017-07-13 as of this writing). Turns out the `DispatchTimeInterval` enum change isn't available there (yet?)

I'd like to think that this will be unnecessary in a future Swift snapshot on Linux, but perhaps it's worth integrating this small fix so that the release candidates aren't busted until that happens.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,6 +84,8 @@ jobs:
       env:
         - JOB=SWIFTPM_DARWIN
     - osx_image: xcode9
+      before_install:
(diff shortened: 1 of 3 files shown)


andersio:
Could you please update also the Darwin snapshot?

liscio:
@andersio I'm assuming you mean the `SWIFT_4` job? Is what I did in d7eb8bc sufficient?

PR #492, opened 2017-07-19
Fix (some) SwiftLint violations

erwald:
#### Checklist
- [x] Updated CHANGELOG.md.

This PR adds the `.swiftlint.yml` config file, and runs `swiftlint autocorrect` to fix _some_ issues. Mind, there are still 740 violations left (of which 601 are classified as "serious").

I disabled the `opening_braces` rule because it didn't like arrays of closures, like this:

```swift
let expectations: [(String) -> Void] = [
  { event in expect(event) == "[] value 1" },
  { event in expect(event) == "[] completed" },
  { event in expect(event) == "[] terminated" },
  { event in expect(event) == "[] disposed" }
]
```

Maybe some other rules need to be disabled as well, such as the tuple size and identifier name length rules.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,0 +1,2 @@
+excluded:
+  - Carthage
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,5 +1,6 @@
 disabled_rules:
   - opening_brace
+  - trailing_comma
 included:
   - Sources
   - Tests
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,2 +1,3 @@
-excluded:
-  - Carthage
+included:
+  - Sources
+  - Tests
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,3 +1,5 @@
+disabled_rules:
+  - opening_brace
 included:
   - Sources
   - Tests
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -3,3 +3,5 @@ disabled_rules:
 included:
   - Sources
   - Tests
+
+line_length: 200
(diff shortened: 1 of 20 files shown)


andersio:
It doesn’t need to be in changelog, since it is not relevant to the public API.

erwald:
I suspected as much. Fixed!

Sources/EventLogger.swift
```diff
@@ -14,15 +14,15 @@ public enum LoggingEvent {
 		case value, completed, failed, terminated, disposed, interrupted
 
 		public static let allEvents: Set<Signal> = [
-			.value, .completed, .failed, .terminated, .disposed, .interrupted,
+			.value, .completed, .failed, .terminated, .disposed, .interrupted
```
mdiep: Can we change this rule? By default SwiftLint enforces no trailing comma. I think we should enforce the trailing comma. This tends to make diffs more readable and merges easier since adding a line doesn't alter the previously last item in the array.

erwald: Makes sense. I added the exception and reinserted the commas.

.swiftlint.yml
```diff
@@ -0,0 +1,8 @@
+disabled_rules:
+  - opening_brace
+  - trailing_comma
```
mdiep: Instead of disabling it, you can actually _require_ the comma by adding this at the top level:

```
trailing_comma:
  mandatory_comma: true
```

Would you mind doing that instead? Sorry that wasn't clear before. 😅

erwald: Ah-ha! Done and done.

erwald:
Not quite sure why the Linux build fails – it seems `Test Case 'XCTestCase.MutableProperty, should not deadlock' started at 18:02:17.226 ` does, in fact, deadlock.

But I'm not sure how this PR could affect that. Are the tests sometimes unstable, by any chance?

mdiep:
I think that particular test _is_ unstable. I've rerun those tests—hopefully they'll pass this time.

PR #495, opened 2017-07-23
Address the exceptionally high build time.

andersio:
Swift 3 compiler is less performant in situations that need sophisticated inference. It appears not to be an issue for Swift 4 though. 💁‍♂️ 

While the constraints have been tightened, IMO it shouldn't be considered API breaking, since the `Property` requirements mandate `NoError`, and the `Error` associated type is just an implementation detail that is always `NoError` at conformance time.

Ideally we would get rid of these in RAS 3.0 by constraining `PropertyProtocol` with `Error == NoError`.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # master
 *Please add new entries at the top.*
 
+1. Addressed the exceptionally high build time. (#495)
+
 # 2.0.0
 # 2.0.0-rc.3
 1. `Lifetime.+=` which ties a `Disposable` to a `Lifetime`, is now part of the public API and is no longer deprecated.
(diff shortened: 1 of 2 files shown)


mdiep:
Changes to type constraints always make me nervous because of the potential to change the type inference. 🙈

mdiep:
Would adding `associatedtype Error = NoError` directly on `PropertyProtocol` have the same effect?

andersio:
`master` and default type for `Error`:
screen shot 2017-08-09 at 11 14 12 pm

This branch:
screen shot 2017-08-09 at 11 16 03 pm

It seems the problem isn't very severe in `master`. But I managed to blow it up in #444. 🙈

PR #497, opened 2017-07-26
Add `SignalProducer.init` overloads for `NoError`

ikesyo:
See https://github.com/Carthage/Carthage/pull/2077#discussion_r129483994 and https://github.com/Carthage/Carthage/pull/2077#discussion_r129484077.

#### Checklist
- ~~[ ] Updated CHANGELOG.md.~~

### DIFF
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -250,23 +250,6 @@ extension SignalProducer where Error == NoError {
 		}
 	}
 
-	/// Creates a producer for a `Signal` that immediately sends one value, then
-	/// completes.
-	///
-	/// This initializer differs from `init(value:)` in that its sole `value`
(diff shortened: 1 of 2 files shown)


mdiep:
I don't understand what this adds. Could you elaborate? It seems like this would just increase conflicts and require more manual disambiguation.

ikesyo:
To disambiguate something like this:

```swift
let producer1 = SignalProducer(value: "")
	.flatMap(.concat) { str -> SignalProducer<Int, AnyError> in .empty }

let producer2 = SignalProducer(["", ""])
	.flatMap(.concat) { str -> SignalProducer<Int, AnyError> in .empty }
```

mdiep:
This breaks source compatibility, but maintains binary compatibility, correct?

andersio:
I am not convinced that it is source breaking.

PR #499, opened 2017-07-31
Skip Carthage and CocoaPods CI tests except for push & tag builds.

andersio:


mdiep:
I'm not a fan of disabling these. The CocoaPods test, in particular, is useful for catching issues since the spec needs to be kept in sync with the information in the Xcode files.

As a reviewer, these make my job easier.

PR #505, opened 2017-08-26
Improve TransformerCore and move `take(first:)`.

andersio:
The requirement of `SignalProducerCore.start` has been changed slightly, so as to support `TransformerCore` yielding terminal events from value events.

It intercepts the output of the combined event transformation to:

1. interrupt the upstream when appropriate; and
2. ensure only one terminal event would ever be delivered.

The overhead should be per block of contiguous applications of `TransformerCore` operators.

### Example
Using `master` for this new `take(first:)` (#487) causes the `SignalProducer.repeat` test case to fail with stack overflow.

```swift
SignalProducer<Int, NoError>(value: 1)
    .repeat(.max)
    .take(first: 1)
    .start()
```

This is because the `TransformerCore` implementation in `master` is not designed in mind with transformations that could yield terminal events. So even if `take(first: 1)` yields a `completed` as it receives the first value event, the upstream is not interrupted.

#### Checklist
- ~~Updated CHANGELOG.md.~~

### DIFF
--- a/Sources/Event.swift
+++ b/Sources/Event.swift
@@ -180,7 +180,9 @@ extension Signal.Event: EventProtocol {
 }
 
 extension Signal.Event {
-	internal static func filter(_ isIncluded: @escaping (Value) -> Bool) -> (@escaping Signal<Value, Error>.Observer.Action) -> (Signal<Value, Error>.Event) -> Void {
+	internal typealias Transformation<U, E: Swift.Error> = (@escaping Signal<U, E>.Observer.Action) -> (Signal<Value, Error>.Event) -> Void
+
+	internal static func filter(_ isIncluded: @escaping (Value) -> Bool) -> Transformation<Value, Error> {
 		return { action in
 			return { event in
 				switch event {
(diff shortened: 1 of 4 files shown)


andersio:
With the same test from #487, the overhead has increased quite a bit on the explicit map-then-filter test by ~500ns, which adds 8% CPU time.

The constant producer test has also been affected, likely due to the change with `SignalProducerCore.start`.

```
@testEventTransformingCoreMapFilter(): avg 6503 ns; min 5709 ns
@testSequence(): avg 13188 ns; min 11751 ns
@testValue(): avg 400 ns; min 352 ns
```

mdiep:
1. Can you add the sample code above as a test?

2. Does this affect the other `take` operators? (`take(until:)` and `take(during:)`)

3. Does this affect other operators that send terminations? (`sample(on:)`, `timeout`, etc.)

andersio:
> 1. Can you add the sample code above as a test?

It is a reduced form of an existing test case "should evaluate lazily".

> Does this affect the other take operators? (take(until:) and take(during:))

Only `take(first:)` is changed.

> Does this affect other operators that send terminations? (sample(on:), timeout, etc.)

For synchronous operator, until we migrate them to use `TransformerCore`, it is unlikely be affected, unless it depends on other `TransformerCore` operators.

Asynchronous operators like `timeout` cannot use `TransformerCore`.

mdiep:
Thanks for the explanation. That helped a lot. 👍

PR #508, opened 2017-09-04
Fixed a pthread issue in 32-bit ARM debug builds.

andersio:
https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3507

`pthread_mutex_trylock` somehow returns `EAGAIN` on 32-bit ARM devices when the mutex is configured to be error checking. It should either succeed or return `EBUSY` according to the documentation.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # master
 *Please add new entries at the top.*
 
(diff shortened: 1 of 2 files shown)


mdiep:
Nice work! 👏

PR #511, opened 2017-09-07
Migrate operators Vol III.

andersio:
### List of migrated operators
* `attemptMap`;
* `attempt`;
* `take(last:)`;
* `take(while:)`;
* `skip(while:)`;
* `combinePrevious`;
* `scan`;
* `reduce`;
* `skipRepeats`;
* `uniqueValues`;
* `promoteError`; and
* `promoteValue`.

### Notes
* `reduce` has a new implementation that doesn't need `Signal`.
* `skipRepeats` has a new implementation that uses one switch, instead of four operators.

### DIFF
--- a/Sources/Event.swift
+++ b/Sources/Event.swift
@@ -280,6 +280,51 @@ extension Signal.Event {
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


mdiep:
I'm liking this. 👍

Issue #227, opened 2017-01-14
Asynchronous operators and producer interruption.


andersio:
Interrupting a `SignalProducer` applied with an asynchronous operator currently delivers `interrupted` (usually) immediately and synchronously.

```swift
it("should deliver the interrupted event immediately") {
	let scheduler = TestScheduler()
	var signalInterrupted = false
	var observerInterrupted = false

	let (signal, _) = Signal<Int, NoError>.pipe()

	SignalProducer(signal)
		.observe(on: scheduler)
		.on(interrupted: { signalInterrupted = true })
		.startWithInterrupted { observerInterrupted = true }
		.dispose()

	expect(signalInterrupted) == false
	expect(observerInterrupted) == true // ⁉️

	scheduler.run()
	expect(signalInterrupted) == true
	expect(observerInterrupted) == true
}
```

I am not sure if this is ideal. Say when applying `observe(on:)` to a `SignalProducer`, my expectation would be **all** event being delivered on the specified scheduler. But it clearly is not the case.

Moreover, the current implementation also _short-circuits_ the `interrupted` event delivery. So even if there is actually a `completed` or `failed` event propagating from the upstream(s), the event would still be dropped if the interrupt is initiated before it arrives.

IMO, the `interrupted` event should respect all the applied operators. It should also originate from the ultimate upstream(s) to maintain a consistent event order.

`failed` and `interrupted` are defined to be [delivered immediately](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Documentation/DesignGuidelines.md#failures-behave-like-exceptions-and-propagate-immediately), but as implied by the context the immediacy is only with regard to time-shifting. On the other hand, asynchronous operators have a [strong implicit guarantee that all events should be delivered on the specified scheduler](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Documentation/DesignGuidelines.md#observe-events-on-a-known-scheduler).

#### API changes
This might break existing use that depends on the fraudulent behavior, e.g. `debounce` (Fixed in #228).

/cc @ReactiveCocoa/reactiveswift

mdiep:
Oof. That's tricky. I'm not sure which is right. But it seems like a 2.0 change if we make it.

mdiep:
Per #424, this has been resolved.

PR #512, opened 2017-09-09
Migrate operators Vol IV.

andersio:
Migrated operators:
1. `throttle(_:on:)`;
2. `debounce(_:on:)`;
3. `SignalProducer.negate`;
3. `SignalProducer.and`; and
3. `SignalProducer.or`.

A note on the current requirements for an operator to drop down to event transformation has been added.

### DIFF
--- a/Sources/Event.swift
+++ b/Sources/Event.swift
@@ -174,6 +174,34 @@ extension Signal.Event: EventProtocol {
 	}
 }
 
+// Event Transformations
+//
+// Operators backed by event transformations have such characteristics:
+//
+// 1. Unary
+//    The operator applies to only one stream.
+//
+// 2. Serial
+//    The outcome need not be synchronously emitted, but all events must be delivered in
+//    serial order.
+//
+// 3. No side effect upon interruption.
+//    The operator must not perform any side effect upon receving `interrupted`.
+//
+// Examples of ineligible operators (for now):
+//
+// 1. `timeout`
+//    This operator forwards the `failed` event on a different scheduler.
+//
+// 2. `combineLatest`
+//    This operator applies to two or more streams.
+//
+// 3. `SignalProducer.then`
+//    This operator starts a second stream when the first stream completes.
+//
+// 4. `on`
+//    This operator performs side effect upon interruption.
+
 extension Signal.Event {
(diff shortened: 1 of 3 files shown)


Sources/Event.swift
```diff
@@ -683,6 +711,88 @@ extension Signal.Event {
 			}
 		}
 	}
+
+	internal static func throttle(_ interval: TimeInterval, on scheduler: DateScheduler) -> Transformation<Value, Error> {
+		precondition(interval >= 0)
+
+		return { action in
+			let state: Atomic<ThrottleState<Value>> = Atomic(ThrottleState())
+			let schedulerDisposable = SerialDisposable()
```
mdiep: If this is used to create a `Signal` or `SignalProducer` and it is subsequently interrupted, it doesn't look like that interruption will propagate through to the scheduler here. (Unlike the previous implementation.)

andersio: This is the current behaviour which is part of the operator lifting change (#140), although only `observe(on:)` is mentioned in the release note. The related issue was #227.

> interrupted now respects `observe(on:)`. (#140)
>
> When a produced Signal is interrupted, if observe(on:) is the last applied operator, interrupted would now be delivered on the Scheduler passed to observe(on:) just like other events.

andersio:
```
Disposals: ** <------------------ D1
                                  ** <--------------------------- D2
Events:    P1 -> map -> filter -> P2 -> map -> take -> observe -> P3
```

Disposing of D2 causes produced signal P2 to emit `interrupted` immediately (in general). The map-take-observe chain would then receive `interrupted`. P3 would not observe `interrupted` immediately, but the asynchrony and the serial event order are respected.

andersio: Oh, I just realised I might have interpreted your question wrong.

> it doesn't look like that interruption will propagate through to the scheduler here.

We cannot interrupt an event transform specifically, so the interruption must be sourced from an upstream that can handle the interruption. In the implementation of `throttle`, all terminal events are scheduled to the given scheduler.

https://github.com/ReactiveCocoa/ReactiveSwift/pull/512/files/87592cd3f0968f35af8bd23f07cc02263ffad675#diff-c949eaeac664030204497a79420aaef3R724

mdiep: Thanks for the explanation. That helped a lot. 👍

PR #515, opened 2017-09-09
Replace `Observer.action` with `Observer.send`.

andersio:
`Signal.Observer.action` does not capture the `Signal.Observer` it derived from. This might cause issues with `Signal` inputs and transformed observer, since these rely on the lifetime of the `Signal.Observer`. This also implies a semantic inconsistency between `action` and any method reference to `send*`.

While patching `Observer.action` is sufficient, it appears to be a great opportunity to hide the closure as an implementation detail, and complete the `send` method family with a new overload taking the event sum type.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/Sources/Observer.swift
+++ b/Sources/Observer.swift
@@ -11,9 +11,16 @@ extension Signal {
 	/// (typically from a Signal).
 	public final class Observer {
 		public typealias Action = (Event) -> Void
+		private let _action: Action
 
 		/// An action that will be performed upon arrival of the event.
-		public let action: Action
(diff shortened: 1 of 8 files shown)


Sources/Observer.swift
```diff
@@ -11,9 +11,16 @@ extension Signal {
 	/// (typically from a Signal).
 	public final class Observer {
 		public typealias Action = (Event) -> Void
+		private let _action: Action
```
mdiep: This seems like it should be called `_send`.

mdiep:
👍 on the name change

Issue #423, opened 2017-06-01
A `once` flatten strategy.


andersio:
It is pretty common just to flatten at most one inner signal. For example, for APIs that promise to just send the value at once, either `latest`, `concat` or `merge` is irrelevant and might miscommunicate the intent by implying concurrency or serialisation that isn't applicable.

So the proposal is to introduce a flatten strategy that is equivalent to `.take(first: 1).flatMap(_)`.

It is marked as breaking since Swift doesn't support versioned open enum yet AFAIK, although it is part of library evolution.

mdiep:
This doesn't seem to me like it'd bring much benefit.

andersio:
You can treat it as a neat shorthand like `prefix(value:)` kind of stuff.

Say then one got a bunch of composition:
```swift
producer
    .flatMap(.once, workA)
    .flatMap(.once, workB)
    .flatMap(.once, workC)
    .flatMap(.once, workD)
```

It looks much better than:
```swift
producer
    .take(first: 1).flatMap(.latest, workA)
    .take(first: 1).flatMap(.latest, workB)
    .take(first: 1).flatMap(.latest, workC)
    .take(first: 1).flatMap(.latest, workD)
```

It is true that such strategy is useless if we for sure know at runtime `producer` would send one value. But stronger constraints are always nice to have if it is general enough IMO.

andersio:
@mdiep I think `once` has a leverage if it would propagate any `interrupted` event from inner signal. I actually ran into an complicated issue caused by existing flatten operators mimicking `once` but not propagating `interrupted`.

andersio:
It is especially important if we do treat `interrupted` [being more than user cancellation](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Documentation/APIContracts.md#interruptions-cancel-outstanding-work-and-usually-propagate-immediately). In my case, it is a custom `trigger` operator that conditionally cancels network requests. With `latest` and `concurrent`, it just completes when it isn't supposed to be.

andersio:
Well apparently `take(first: 1).flatMap(.race)` can achieve the expected semantic for a `once` strategy.

RuiAAPeres:
Closing this due to inactivity. @andersio you are free to make a PR with the `once` implementation that we are using in our codebase ([at babylon health](https://github.com/Babylonpartners))

leonid-s-usov:
[...]

Issue #201, opened 2016-12-30
Introducing `Single` type #2873


RuiAAPeres:
This refers to [ReactiveCocoa/Reactivecocoa/#2873](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2873)

liscio:
I don't see the value of having a new type to represent this.

As a consumer of an API like ReactiveSwift, I think it's nice and compact to think in terms of the two major constructs (`Signal` and `SignalProducer`). If you wanted to represent something like a Future using a `Single`, then I'd argue a `SignalProducer` makes far more sense because the caller is in control of kicking off the process that results in the value they are looking for. The same goes for `SignalProducer.init(value:)` or `SignalProducer.init(error:)`—it nicely captures the narrow scope of the created producer.

Furthermore, if the caller of an API vending a `Signal{,Producer}` is concerned about ensuring they only get a single value, then `take(1)` is a succinct way to assert exactly that in the code.

Perhaps a few examples of ReactiveSwift (pseudo-)code that'd be improved by `Single` would help sway my current 👎 attitude towards this. :)

andersio:
I'd like to see if it could bring additional compile time guarantee over `SignalProducer`. AFAIK there is no way (and no plan) to have an attribute to guarantee that a closure can only be invoked once. So in the end, in terms of the static API contract it could very well be an equivalent of `SignalProducer`.

A new type also adds more overloads to our lifting-based operator model, and the maintenance cost has to be considered too.

RuiAAPeres:
👍

NachoSoto:
I'm also a big opponent of this idea (happy to explain why).

mdiep:
I'd love to hear your thoughts @NachoSoto

NinoScript:
I'm really interested in why this may not be a good idea. @NachoSoto ?

RuiAAPeres:
ping @NachoSoto 😛

andersio:
ping @NachoSoto

Qata:
ping @NachoSoto 🎊

PR #516, opened 2017-09-11
The `once` FlattenStrategy.

andersio:
Related: #423

This is a convenience that has been using in our codebase for a while, implemented as `take(first: 1).race()`. There are a lot of circumstances that we need only one value from a continuous producer (e.g. property).

For example, we can fold this:
```swift
property.producer
    .take(first: 1)
    // Since we have just one value, any strategy is irrelevant.
    .flatMap(.latest) { value in
        // Spawn a network request from the value.
    }
```

into:
```swift
property.producer
    .flatMap(.once) { value in
        // Spawn one network request from the first value received.
    }
```

While it doesn't add substantial functionality, it offers a sensible choice for flattening single-value stream of streams, that also reduces verbosity, from an API consumer POV.

#### Checklist
- [ ] Updated CHANGELOG.md.

### DIFF
--- a/Sources/Flatten.swift
+++ b/Sources/Flatten.swift
@@ -14,6 +14,7 @@ public struct FlattenStrategy {
 		case concurrent(limit: UInt)
 		case latest
 		case race
+		case once
 	}
 
 	fileprivate let kind: Kind
@@ -98,6 +99,19 @@ public struct FlattenStrategy {
 	/// Any failure from the inner streams is propagated immediately to the flattened
 	/// stream of values.
 	public static let race = FlattenStrategy(kind: .race)
+
+	/// Forward only events from the first inner stream received. The stream of streams
+	/// would be interrupted upon the said stream being started.
+	///
+	/// The flattened stream of values completes when the stream of streams completes
+	/// without emitting any inner stream, or when the sole inner stream completes.
+	///
+	/// Any interruption of inner streams is propagated immediately to the flattened
+	/// stream of values.
+	///
+	/// Any failure from the inner streams is propagated immediately to the flattened
+	/// stream of values.
+	public static let once = FlattenStrategy(kind: .once)
 }
 
 extension Signal where Value: SignalProducerConvertible, Error == Value.Error {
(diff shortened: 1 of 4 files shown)


Sources/Flatten.swift
```diff
@@ -164,6 +181,9 @@ extension Signal where Value: SignalProducerConvertible, Error == NoError, Value
 
 		case .race:
 			return self.race()
+
+		case .once:
+			return self.take(first: 1).race()
```
RuiAAPeres: On the description you explain it's equivalent to `flatMap(.latest)`, but internally you are using `race`.

mdiep: Since there's only one signal, all the strategies are functionally equivalent. The only difference is the performance characteristics.

mdiep:
I'm pretty ambivalent about this addition. 😕

Sources/Flatten.swift
andersio: We do not have a case for performance here yet, just the room for optimization.

> Since there's only one signal, all the strategies are functionally equivalent. 

`race` propagates `interrupted` from the inner stream to the flattened stream. The other strategies do not.

andersio: Hmm, so the strategy is relevant.

Qata:
I'm a fan of this addition, the app I most use ReactiveSwift with has a lot of cases that'd benefit from this readability. 
I've had to do a workshop to explain why all of our single value streams use `.latest`.

mdiep:
Maybe it's time to reconsider adding a `Single` type of some sort. It seems like that's the real issue here: it's cumbersome to deal with a stream of _n_ events when you know there will only be 1. #201

andersio:
I am not convinced to have it as a dedicated type, since there is barely a compile time advantage. To facilitate it, we would also have to replicate the entire API of `SignalProducer` given the lack of HKT. In the end, all we gain for all these work and on going maintenance is probably just a strategy-less `flatMap`, and perhaps a specialised input observer API that accepts value or error.

andersio:
> a dedicated type name that benefits from a doc PoV

I guess this alone could be the primary driver though, since that's also how the division between `Signal` and `SignalProducer` started despite statically having minimal distinction.

mdiep:
Yeah, I think there are benefits to documentation/reasoning. Many common tasks could be represented as a observable of one value. It would be easy to have APIs to convert between them as well.

The biggest issue is coming up with a good name.

PR #517, opened 2017-09-14
Allow any value signal for trigger in takeUntil

iv-mexx:
#### Checklist
- [x] Updated CHANGELOG.md.

I was wondering why the trigger signal for `take(until:)` was constrained to only signals/signal producers of value `()`. 
If there is a specific reason for that constraint, please ignore this PR.

This change allows the trigger signal for `take(until:)` to have any value. In my experience, my trigger signals usually are "normal" signals that I currently have to map to `()`. 

So, before: 

```swift
let someSignal: Signal<Int, NoError>
let triggerSignal: Signal<Any, NoError>
...
someSignal.take(until: triggerSignal.map { _ in return })
```

After:

```swift
let someSignal: Signal<Int, NoError>
let triggerSignal: Signal<Any, NoError>
...
someSignal.take(until: triggerSignal)
```

Regarding tests, should I write additional tests with a trigger signal of another value, or should I change the value of the trigger signal in the current tests, or is it fine already?

PS: Apologies for the unnecessary changes in CHANGELOG.md. My Editor was configured to strip trailing whitespace and I did not notice until the diff here. I can revert these changes if you like, but I guess they don't hurt anyway...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # master
 *Please add new entries at the top.*
 
+1. `Signal.take(until:)` and `SignalProducer.take(until:)` now accept Triggers of any value instead of just `()`
+
 1. `Signal.Observer.action` has been deprecated. Use `Signal.Observer.send` instead. (#515)
 
 1. Workaround an unexpected EGAGIN error being returned by pthread in 32-bit ARM debug builds. (#508)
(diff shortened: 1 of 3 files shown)


NachoSoto:
There is a reason: we want users to be 100% aware that the values emitted by triggers will be ignored, and do so explicitly.
Thanks for the PR though!

brow:
You can shorten `.map { _ in return }` to `.map { _ in }` for convenience 💁

PR #533, opened 2017-10-11
Make `startWithSignal()` return the result of `setup(…)`.

Burgestrand:
Personally I'd use it to return the `Disposable` given to the block, so that I have an interruption handle for the started producer, similar to the other `start()`-methods.

By using `@discardableResult` we retain backwards-compatibility without causing warnings about unused return values.

#### Example usage

``` swift
let disposable = producer.startWithSignal { signal, disposable in
  // …
  return disposable
}
```

_Note: I have not discussed this feature with anybody beforehand, but the change was easy enough so I wrote it and submitted a PR_

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # master
 *Please add new entries at the top.*
 
+1. `SignalProducer.startWithSignal` now returns the value of the setup closure.
+
 # 2.1.0-alpha.2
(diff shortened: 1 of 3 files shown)


Burgestrand:
It looks like the last build could have been a random failure, but since I don't have write access to the repository I can't trigger a rebuild in Travis.

PR #534, opened 2017-10-14
Tweak flatten operators to take advantage of the new producer cores.

andersio:
The implementations now use `startWithInterrupter` instead of `startWithSignal`.

#### Checklist
- ~~Updated CHANGELOG.md.~~


Sources/Flatten.swift
```diff
@@ -916,10 +907,8 @@ extension Signal {
 			case let .value(value):
 				observer.send(value: value)
 			case let .failed(error):
-				handler(error).startWithSignal { signal, disposable in
-					serialDisposable.inner = disposable
-					signal.observe(observer)
-				}
+				serialDisposable.dispose()
```
mdiep: `serialDisposable` seems like it might be unnecessary? I don't see how it has any effect since the upstream doesn't need to be disposed when it has errored.

andersio:
Dropping `startWithSignal` could fail #563.

PR #535, opened 2017-10-14
Remove redundant constraints.

andersio:
#### Checklist
- ~~Updated CHANGELOG.md.~~

### DIFF
--- a/Sources/Property.swift
+++ b/Sources/Property.swift
@@ -9,9 +9,7 @@ import enum Result.NoError
 ///
 /// Only classes can conform to this protocol, because having a signal
 /// for changes over time implies the origin must have a unique identity.
-public protocol PropertyProtocol: class, BindingSource {
-	associatedtype Value
-
+public protocol PropertyProtocol: class, BindingSource where Error == NoError {
 	/// The current value of the property.
 	var value: Value { get }
 
@@ -92,7 +90,7 @@ extension PropertyProtocol {
 	}
 }
 
-extension PropertyProtocol where Error == NoError {
+extension PropertyProtocol {
 	/// Maps the current value and all subsequent values to a new property.
 	///
 	/// - parameters:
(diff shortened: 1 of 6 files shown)


Sources/Property.swift
```diff
@@ -306,55 +304,55 @@ extension PropertyProtocol {
...
 	/// `zip(with:)`.
-	public static func zip<A: PropertyProtocol, B: PropertyProtocol, C: PropertyProtocol>(_ a: A, _ b: B, _ c: C) -> Property<(Value, B.Value, C.Value)> where A.Value == Value, A.Error == NoError, A.Error == B.Error, A.Error == C.Error {
+	public static func zip<A: PropertyProtocol, B: PropertyProtocol, C: PropertyProtocol>(_ a: A, _ b: B, _ c: C) -> Property<(Value, B.Value, C.Value)> where A.Value == Value {
 		return a.lift { SignalProducer.zip($0, b, c) }
 	}
 
 	/// Zips the values of all the given properties, in the manner described by
 	/// `zip(with:)`.
-	public static func zip<A: PropertyProtocol, B: PropertyProtocol, C: PropertyProtocol, D: PropertyProtocol>(_ a: A, _ b: B, _ c: C, _ d: D) -> Property<(Value, B.Value, C.Value, D.Value)> where A.Value == Value, A.Error == NoError, A.Error == B.Error, A.Error == C.Error, A.Error == D.Error {
+	public static func zip<A: PropertyProtocol, B: PropertyProtocol, C: PropertyProtocol, D: PropertyProtocol>(_ a: A, _ b: B, _ c: C, _ d: D) -> Property<(Value, B.Value, C.Value, D.Value)> where A.Value == Value {
```
NachoSoto: These I'm okay removing since there's so much duplicating

andersio: There are occasions of redundant constraints being kept. For example, [removing this would cause the associated type inference of mutable property types to fail](https://github.com/ReactiveCocoa/ReactiveSwift/pull/535/files#diff-a912eac10c1971f50158ab4c91274879R21).

andersio: These are removed basically because `PropertyProtocol` is now constrained with `Error == NoError`.

PR #540, opened 2017-10-19
Recursive value delivery.

andersio:
While it is once said that recursive signals are wrong, unidirectional data flow is actually an example of proper use.

This PR proposes to reintroduce signal recursion with defined semantics, specifically the guarantee of _observing a signal respecting program order_. This makes `Signal` an eligible primitive for these kinds of patterns, with less surprises in deadlocks that consequentially require asynchronous workarounds.

> A signal delivers to its known observers immediately upon instruction. A future observer never sees any value happened before its attachment in program order. Note that such guarantee is applied only in a synchronized scenario — the concurrent behavior is indeterministic.
>
> A signal guarantees such order also in a recursive environment. That is, if a value is piped to a signal by a value currently processing by the said signal, the delivery of such value always respect the program order.

For example, given:
```swift
let (signal, observer) = Signal.pipe()
signal.observeValues { value in
    signal.observeValues { _ in }
    observer.send(value: value)
}
```
Since the observation happens _before_ the emission of value, the observation is guaranteed to see the value being sent.

On the other hand, given:
```swift
let (signal, observer) = Signal.pipe()
signal.observeValues { value in
    observer.send(value: value)
    signal.observeValues { _ in }
}
```
Since the observation happens _after_ the emission of value, the observation is guaranteed to never see the value being sent, but only any future value being yielded.

### What about `MutableProperty`?
Recursion support for the `MutableProperty` is hardly a brilliant idea, primarily due to its synchronous replay-latest semantic. Consider this case:
```swift
property.signal.observeValues { current in
    property = current + 1
    print(property.value)
}
```
There are two choices of semantics:

1. We defer the recursively set value after the processing of the current value (like this PR).

    This breaks the perceived program order, because writes should be seen by the writer immediately for a property.

2. We reflect the latest value regardless of the progress of observer call-out.

    This breaks the guarantee of property observers seeing the same value from its sink input and `PropertyProtocol.value`.

#### Checklist
- ~~Updated CHANGELOG.md.~~

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -56,6 +56,9 @@ public final class Signal<Value, Error: Swift.Error> {
 		/// The state of the signal.
 		private var state: State
 
+		/// The buffered values.
+		private var buffer: [(Bag<Observer>, Value)]
+
 		/// Used to ensure that all state accesses are serialized.
 		private let stateLock: Lock
 
@@ -64,6 +67,7 @@ public final class Signal<Value, Error: Swift.Error> {
 
 		fileprivate init(_ generator: (Observer, Lifetime) -> Void) {
 			state = .alive(Bag(), hasDeinitialized: false)
+			buffer = []
 
 			stateLock = Lock.make()
 			sendLock = Lock.make()
@@ -105,7 +109,17 @@ public final class Signal<Value, Error: Swift.Error> {
 
 				tryToCommitTermination()
 			} else {
-				self.sendLock.lock()
+				guard self.sendLock.try() else {
+					self.stateLock.lock()
+					defer { self.stateLock.unlock() }
+
+					if case let .alive(observers, _) = self.state {
+						self.buffer.append((observers, event.value!))
+					}
+
+					return
+				}
+
 				self.stateLock.lock()
 
 				if case let .alive(observers, _) = self.state {
@@ -116,6 +130,22 @@ public final class Signal<Value, Error: Swift.Error> {
 					}
 				} else {
 					self.stateLock.unlock()
+				}
+
+				while true {
+					self.stateLock.lock()
+
+					guard case .alive = self.state, !self.buffer.isEmpty else {
+						self.stateLock.unlock()
+						break
+					}
+
+					let (observers, value) = self.buffer.removeFirst()
+					self.stateLock.unlock()
+
+					for observer in observers {
+						observer.send(value: value)
+					}
 				}
 
 				self.sendLock.unlock()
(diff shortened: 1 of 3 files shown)


andersio:
I realize that this effectively turns `Signal` into a queue-drain model for all cases. Perhaps this should be an opt-in behavior.

mdiep:
> While it is once said that recursive signals are wrong, unidirectional data flow is actually an example of proper use.

I'd like to see more justification for why is useful and a good idea.

Sources/Signal.swift
```diff
@@ -118,6 +132,22 @@ public final class Signal<Value, Error: Swift.Error> {
 					self.stateLock.unlock()
 				}
 
+				while true {
```
NachoSoto: This doesn't seem like the best way to implement this, but I guess you can't do `while !self.buffer.isEmpty` cause you need to lock?

andersio: On top of locking, the signal needs to be alive before us proceeding.

andersio:
> I'd like to see more justification for why is useful and a good idea.

A unidirectional data flow akin to Elm, [RxFeedback](https://github.com/kzaher/RxFeedback#rxfeedback) or [ReactiveFeedback](https://github.com/babylonpartners/ReactiveFeedback) usually comprises of three components:

1. The current state
2. A reducer computing the new state from the current state and a message.
3. Multiple observers of the current state that emit messages conditionally.

```
   ,--------------------------------------------------,
   |                                                  |
   `--> ,---------, Sn+1     ,-------------------,    |
        | REDUCER | ---*---> | FEEDBACKS/UPDATES | ---'
Sn ,--> '---------'    |     '-------------------' Msg/Event
   |                   |
   '-------------------'
```

To implement it using `Signal` in ReactiveSwift, as a feedback loop exists, it is prone to deadlock because the feedback path could be triggered synchronously, effectively being a recursion.

```swift
let reducer: (State, Event) -> State
let feedbacks: [(Signal<State, NoError>) -> Signal<Event, NoError>]

let (stateRelay, stateRelayObserver) = Signal<State, NoError>.pipe()

// Construct the Feedback <-> Reducer feedback loop.
Signal.merge(feedbacks.map { $0(stateRelay) })
    .scan(into: initialState, reducer)
    .observe(stateRelayObserver)

// Bootstrap the system.
stateRelayObserver.send(value: initialState)
```

It is possible to mitigate all potentials of deadlock by forcing all feedback paths to be asynchronous. But I think we could do better than that and stick to the "synchronous by default" principle of ours.

mdiep:
I don't think this is a great idea. But I'd be curious to hear what the rest of @ReactiveCocoa/reactiveswift thinks

This adds quite a bit of additional complexity. Granted, the patch isn't that large. But I'm concerned about the additional drains on maintainability. I'm also worried about more complex interactions that may we may not be taking into account here.

On top of that, I think loops like this are generally a bad idea. They make signal chains harder to reason about and more prone to bugs. I don't want to make bad ideas easy.

But my opinion on this is definitely influenced by my relative opinions of Elm and React-style architecture. Elm's model is quite a bit different and doesn't have these issues.

andersio:
I think we might be able to approach this in a more general form without explicitly adding recursive delivery to `Signal`.

andersio:
It doesn't seem to solve the issue completely. For it to work consistently, the queue-drain needs to be available on the entire cyclic `Signal` graph.

Issue #506, opened 2017-08-29
Integer overflow with `timer` function on specifying seconds interval greater than 21474.836471


stephanecopin:
Hey guys,

I think this is just a documentation mistake at https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/FoundationExtensions.swift#L155-L157 and https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/SignalProducer.swift#L2478-L2480.
The documentation specifies 200,000 as the maximum interval, however 10% of 200,000 is 20,000 (rather than 2,000), so the `Int` conversion will overflow with any [micro/milli]seconds values greater than 21474.836471.
Not sure if just the documentation should be updated, or a more in depth change should be made. I feel like an interval like 21,600 seconds (6h00) is more likely to be set by someone than 200,000, for applications that are allowed to run in the background for example.

Thoughts?

andersio:
@liscio

RuiAAPeres:
Hello. 👋  Thanks for opening this issue. Due to inactivity, we will soft close the issue. If you feel that it should remain open, please let us know. 😄

PR #543, opened 2017-10-22
Fixed integer overflows for Foundation extensions

soxjke:
- Fixed integer overflows for Foundation extensions (#506)  …
- Added overflow test case
- Removed note on value limits for DispatchTimeInterval `*` operator

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # master
 *Please add new entries at the top.*
-Fixed integer overflow for `DispatchTimeInterval` in FoundationExtensions.swift (#506)
+1. Fixed integer overflow for `DispatchTimeInterval` in FoundationExtensions.swift (#506)
 
 # 3.0.0-alpha.1
 # 3.0.0-alpha.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # master
 *Please add new entries at the top.*
+Fixed integer overflow for `DispatchTimeInterval` in FoundationExtensions.swift (#506)
 
 # 3.0.0-alpha.1
 # 3.0.0-alpha.1
--- a/Tests/ReactiveSwiftTests/FoundationExtensionsSpec.swift
+++ b/Tests/ReactiveSwiftTests/FoundationExtensionsSpec.swift
@@ -94,6 +94,13 @@ class FoundationExtensionsSpec: QuickSpec {
 				expect((DispatchTimeInterval.seconds(5) * 0.5).timeInterval).to(beCloseTo(DispatchTimeInterval.milliseconds(2500).timeInterval))
 				expect((DispatchTimeInterval.seconds(1) * 0.25).timeInterval).to(beCloseTo(DispatchTimeInterval.milliseconds(250).timeInterval))
 			}
+			
+			it("should not introduce integer overflow upon scale") {
+				expect((DispatchTimeInterval.seconds(Int.max) * 0.01).timeInterval).to(beCloseTo(10 * DispatchTimeInterval.milliseconds(Int.max).timeInterval, within: 1))
+				expect((DispatchTimeInterval.milliseconds(Int.max) * 0.01).timeInterval).to(beCloseTo(10 * DispatchTimeInterval.microseconds(Int.max).timeInterval, within: 1))
+				expect((DispatchTimeInterval.microseconds(Int.max) * 0.01).timeInterval).to(beCloseTo(10 * DispatchTimeInterval.nanoseconds(Int.max).timeInterval, within: 1))
+				expect(DispatchTimeInterval.seconds(Int.max) * 10).to(equal(DispatchTimeInterval.never))
+			}
 
 			it("should produce the expected TimeInterval values") {
 				expect(DispatchTimeInterval.seconds(1).timeInterval).to(beCloseTo(1.0))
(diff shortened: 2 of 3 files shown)


andersio:
Seems like the test has failed to be built on Linux.

```
/home/travis/build/ReactiveCocoa/ReactiveSwift/Tests/ReactiveSwiftTests/FoundationExtensionsSpec.swift:102:59: error: cannot invoke 'equal' with an argument list of type '(DispatchTimeInterval)'
 expect(DispatchTimeInterval.seconds(Int.max) * 10).to(equal(DispatchTimeInterval.never))
 ^
/home/travis/build/ReactiveCocoa/ReactiveSwift/Tests/ReactiveSwiftTests/FoundationExtensionsSpec.swift:102:59: note: overloads for 'equal' exist with these partially matching parameter lists: (T?), ([T : C]?), ([T]?), ([T?]), (Set<T>?)
 expect(DispatchTimeInterval.seconds(Int.max) * 10).to(equal(DispatchTimeInterval.never))
 ^
```

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # master
 *Please add new entries at the top.*
+Fixed integer overflow for `DispatchTimeInterval` in FoundationExtensions.swift (#506)
```
andersio: Please start the line with `1. `.

andersio:
Thank you for your work. 😸

soxjke:
Coming up with linux fix 👍

Tests/ReactiveSwiftTests/FoundationExtensionsSpec.swift
```diff
@@ -103,15 +110,15 @@ class FoundationExtensionsSpec: QuickSpec {
...
-				expect(DispatchTimeInterval.never.timeInterval) == Double.infinity
+				expect(DispatchTimeInterval.never.timeInterval).to(equal(Double.infinity))
 			}
 
 			it("should negate as you'd hope") {
 				expect((-DispatchTimeInterval.seconds(1)).timeInterval).to(beCloseTo(-1.0))
 				expect((-DispatchTimeInterval.milliseconds(1)).timeInterval).to(beCloseTo(-0.001))
 				expect((-DispatchTimeInterval.microseconds(1)).timeInterval).to(beCloseTo(-0.000001, within: 0.0000001))
 				expect((-DispatchTimeInterval.nanoseconds(1)).timeInterval).to(beCloseTo(-0.000000001, within: 0.0000000001))
-				expect((-DispatchTimeInterval.never).timeInterval) == Double.infinity
```
soxjke: @andersio also if you don't mind I suggest changes to 3 particular test case expressions within this file. From my point of view, `expect.to.equal` syntax explicitly expresses expectation of equality when `==` operator makes people get inside Nimble and look for necessary overload to make sure whether expression is not false-positive.

andersio: I thought this is `Equatable` based, and the stdlib floating point types already imply IEEE 1754 equivalence.

soxjke: Sure, the idea is following:

Using any kind of spec + expectation/matchers framework: (Kiwi, Specta/Expecta, Quick/Nimble) I expect the following pseudo-syntax line for test assertion:
```
expect(inspectable-expression) -> kind-of-expectation(to/notTo, eventually, eventuallyBeforeTimingOutAfter..) -> matcher-expression
```
All of mentioned above frameworks provide us with some DSL to set test expectations, so as a developer I expect an expectation (sorry for tautology) to be set using DSL. With Expecta I can write `expect(someInt) == 0`, and it won't warn me because I can compare returned object with 0 in Objective-C. But the test flow will be wrong, because it won't verify this against any matcher. Instead I should write `expect(someInt).to.equal(0)`, which is correct in terms of its DSL.

Seeing `expect((-DispatchTimeInterval.never).timeInterval) == Double.infinity` in Swift makes me find whether its valid DSL syntax first. I get to https://github.com/Quick/Nimble/blob/master/Sources/Nimble/Matchers/Equal.swift#L172 and find that there's an implementation of `==` operator which makes `lhs.to(equal(rhs)` that I expected in the first place.

Maybe my confusion is connected with unsafe Objective-C background, where equality comparisons could be done mistakingly with reference arithmetics, and we should disregard it, however I would still wonder why to use operator `==` over matcher syntax for particular equality case. Especially, if this operator is implemented by calling matcher syntax.

andersio: `==` are the fluent conveniences of the matchers that are meant to help reduce the verbosity AFAIK. It could be confusing for Swift newcomers with C family language background, but IMO it does make sense given equality and inequality in Swift are implemented as operators, with compile tim overloading & type safety guarding for the problems you may otherwise encounter.

soxjke: No problem, bringing `==` syntax back. I don't have that strong opinion regarding this readability case, so knowing that's just a convenience matcher makes me fine with this approach. Also, I've checked Nimble matchers README and they show using operator `==` for equality matcher and `===` for identity matcher. Thus this point might be considered resolved?

soxjke:
@andersio  
I'm starting to receive
```
Test Case 'XCTestCase.MutableProperty, should not deadlock' started at 2017-10-23 08:37:59.778
```
for Linux build from Travis. Could you probably advice because I'm a bit stick with what's going on with it...

andersio:
@soxjke That's an intermittent issue. :(

soxjke:
@andersio thanks!

I opened #544 to keep track on it. Probably, I'll run vagrant on a local machine to simulate this over weekend.

Other than that, do you see any changes needed more for this PR?

soxjke:
@NachoSoto Sure, I've made overall feature history prettification and dropped also two "Update CHANGELOG.MD" commits as well. Hope Linux tests will succeed this time :)

soxjke:
Probably need to re-run tests again, https://travis-ci.org/ReactiveCocoa/ReactiveSwift/jobs/291677337 suffers from the same failure as #544

PR #547, opened 2017-11-08
Add bottlenecked(_:on:).

iruleonu:
This PR adds a bottlenecked function to bottleneck multiple values. It's useful for the cases where we receive a lot of values but we only want to process one value at a time within a desired time interval. If enough time as elapsed then a received value get processed without any delay otherwise it gets queued/delayed until the interval elapses.

The implementation is similar to the current throttle function with the added logic to schedule the pending value instead of replacing it.

PS: Suggestions for a different name are welcome. Initially I named it throttleThenDelay, then changed to enqueue and ended up with bottlenecked.
PS2: Any suggestion are welcome to :)

### DIFF
--- a/Sources/Event.swift
+++ b/Sources/Event.swift
@@ -788,11 +788,78 @@ extension Signal.Event {
 			}
 		}
 	}
+
+		internal static func bottlenecked(_ interval: TimeInterval, on scheduler: DateScheduler) -> Transformation<Value, Error> {
+		precondition(interval >= 0)
+		
+		return { action in
+			let state: Atomic<BottleneckedState<Value>> = Atomic(BottleneckedState<Value>())
+			let schedulerDisposable = CompositeDisposable()
+			
+			return { event in
+				guard let value = event.value else {
+					schedulerDisposable += scheduler.schedule {
+						action(event)
+					}
+					return
+				}
+				
+				var scheduleDate: Date!
+				state.modify {
+					if $0.enqueuedValues != nil {
+						$0.enqueuedValues?.append(value)
+					} else {
+						$0.enqueuedValues = [value]
+					}
+					
+					let proposedScheduleDate: Date
+					if let previousDate = $0.previousDate, previousDate.compare(scheduler.currentDate) != .orderedDescending {
+						proposedScheduleDate = previousDate.addingTimeInterval(interval)
+					} else {
+						proposedScheduleDate = scheduler.currentDate
+					}
+					
+					switch proposedScheduleDate.compare(scheduler.currentDate) {
+					case .orderedAscending:
+						scheduleDate = scheduler.currentDate
+					case .orderedSame: fallthrough
+					case .orderedDescending:
+						scheduleDate = proposedScheduleDate
+					}
+					
+					if let enqueuedValues = $0.enqueuedValues, enqueuedValues.count > 1 {
+						scheduleDate = scheduleDate.addingTimeInterval(interval * TimeInterval(enqueuedValues.count - 1))
+					}
+				}
+				
+				schedulerDisposable += scheduler.schedule(after: scheduleDate) {
+					let enqueuedValue: Value? = state.modify { state in
+						defer {
+							if state.enqueuedValues?.first != nil, let arraySlice: ArraySlice<Value> = state.enqueuedValues?.dropFirst() {
+								state.enqueuedValues = Array(arraySlice)
+								state.previousDate = scheduleDate
+							}
+						}
+						return state.enqueuedValues?.first
+					}
+					
+					if let enqueuedValue = enqueuedValue {
+						action(.value(enqueuedValue))
+					}
+				}
+			}
+		}
+	}
 }
 
 private struct ThrottleState<Value> {
 	var previousDate: Date?
 	var pendingValue: Value?
+}
+
+private struct BottleneckedState<Value> {
+	var previousDate: Date?
+	var enqueuedValues: [Value]?
 }
 
 extension Signal.Event where Error == NoError {
--- a/Sources/Event.swift
+++ b/Sources/Event.swift
@@ -789,7 +789,7 @@ extension Signal.Event {
 		}
 	}
 
-		internal static func bottlenecked(_ interval: TimeInterval, on scheduler: DateScheduler) -> Transformation<Value, Error> {
+	internal static func bottlenecked(_ interval: TimeInterval, on scheduler: DateScheduler) -> Transformation<Value, Error> {
 		precondition(interval >= 0)
 		
 		return { action in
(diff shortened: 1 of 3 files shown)


Sources/Event.swift
```diff
@@ -788,13 +788,80 @@ extension Signal.Event {
...
 }
 
 private struct ThrottleState<Value> {
 	var previousDate: Date?
 	var pendingValue: Value?
 }
 
+private struct BottleneckedState<Value> {
+	var previousDate: Date?
+	var enqueuedValues: [Value]?
```
andersio: Does it need to be an optional? An empty array seems to be semantically equivalent to `nil` in this case.

Sources/Event.swift
```diff
@@ -788,13 +788,80 @@ extension Signal.Event {
 			}
 		}
 	}
+
+		internal static func bottlenecked(_ interval: TimeInterval, on scheduler: DateScheduler) -> Transformation<Value, Error> {
```
andersio: Indentation is one level deeper than it should be.

Sources/Signal.swift
```diff
@@ -1484,6 +1484,32 @@ extension Signal {
...
+	///
+	/// - precondition: `interval` must be non-negative number.
+	///
+	/// - parameters:
+	///   - interval: Number of seconds to enqueue sent values.
+	///   - scheduler: A scheduler to send values on.
+	///
+	/// - returns: A signal that sends values with a minimun of `interval`
+	///            seconds appart from each other on a given scheduler.
+	public func bottlenecked(_ interval: TimeInterval, on scheduler: DateScheduler) -> Signal<Value, Error> {
```
andersio: It seems that it is similar to `throttle`, just that it queues all instead of just the latest occurrence. I wonder if we should extend `throttle` to have a `ThrottleStrategy` instead.

andersio:
Thanks for the PR! :)

Sources/Signal.swift
iruleonu: I don't think it fits into it. Even though it has a similar implementation, it has a different behaviour which makes it a different concept from what we're expecting from a throttle. If the implementation was completely different (actually lets forget it), would the concept still fit into throttle? 
Thoughts?

Sources/Event.swift
iruleonu: You're right it doesn't. Ill do the change.

Sources/Signal.swift
mdiep: That could make sense. We could include `debounce` at the same time. All 3 are conceptually similar, but have a different strategy. I think our grouping of operators has been very successful and the existing `debounce`/`throttle` divide is very confusing.

The tricky bit is coming up with good names for the strategies. `queue` or `buffer` could work for the `bottlenecked` case. But how do you succinctly describe the existing `throttle` and `dequeue`? I like describing them as "time since last value was received by `self`" and "time since last value was sent by `self`", but those aren't succinct enough.

**Regardless**, I think the `bottlenecked` name isn't good enough. It doesn't really capture the behavior of the operator (but it makes sense if you already understand what the operator does). I think we have the strongest argument for inclusion/acceptance when it fits into an existing name like `throttle`, but we have to find something that works.

iruleonu: Regarding the bottlenecked method name, if we play around with the name of the parameter we can use other alternatives:

>bottleneck(_ throughput: TimeInterval, on scheduler: DateScheduler)
>buffer(_ deflateOneAfter: TimeInterval, on scheduler: DateScheduler)
>constraint(_ minimumSpread: TimeInterval, on scheduler: DateScheduler)

Now finding a good names for throttle strategies that shortly describe the behaviour is more complicated. Here's a suggestion that makes sense together:
>Throttle: ThrottleStrategy.beat since the speed at which the constraint (time interval) runs sets the "beat"
>Debounce: ThrottleStrategy.fluctuate since its last value received that starts the constraint
>Bottlenecked: ThrottleStrategy.buffer or ThrottleStrategy.queue or ThrottleStrategy.enqueue or ThrottleStrategy.spread

mdiep: The best I can think of would be:

* `.queued`
* `.fromEarliest`
* `.fromLatest`

clayellis: What if the master operator was `regulate(over interval: TimeInterval, using strategy: RegulationStrategy, on scheduler: DateScheduler)`?

Where the options for `RegulationStrategy` were:
 - `.throttle`
 - `.debounce`
 - `.bottleneck`
 - `.suppress` 

(`.suppress` is a new strategy I'm proposing in an upcoming PR that will forward the latest value on a scheduler immediately then wait until an interval has passed before sending more values; during the interval, values are suppressed and will not be forwarded after the interval expires. Each new value resets the duration.)

andersio: The base name makes sense, since it is more than just time shifting. `delay` could fall under this too.

andersio: I would prefer to have the strategy first though, akin to `flatMap`.

mdiep: I think `throttle` makes the most sense as an operator name. And I dislike `throttle` and `debounce` as strategy names because I can never remember which is which.

andersio: Hmm, what about `throttle` as the base name, and the `elapse`, `defer` and `buffer`?

iruleonu: We're getting somewhere :)
To give some space for the to be implemented `suppress` strategy (and others), should we rename `all` to `queued` instead?

mdiep: I think `queue` or `queued` is a good name.

`skipPrevious` could apply to either `throttle` or `debounce`, so I'm not wild about it.

andersio: It is quite hard to be concise without bringing in new verbs. Like grace, silence, coalescing+distancing, etc.

It does make me wonder if we express them in combinations though. Say `.distance(.all)` being the bottleneck strategy proposed here, while `.distance(.latest)` being the current `throttle`. `debounce` can be `.silence(.latest)`, and `.silence(.all)` would be effectively `defer`.

clayellis: I think `throttle` would make sense as the base name. 

> Throttle: a device controlling the flow of fuel or power to an engine.

andersio: I was talking about `distance` and `silence` as strategy names with `throttle` as the base name.

clayellis: @andersio ah, I see. I didn't realize `.all`, `.latest` were parameters to the strategy.

iruleonu: I like that approach of just using throttle as the base operator with the combinations for the different strategies. 
I'll try to update this soon, although under a bit of a squeeze for time. Since there's a throttle(while:) option and only one implementation, Im going to need to add support for this option to the missing methods.

andersio:
Closing due to inactivity, and #619 having introduced `collect(every:on:)` that tackles the issue. We shall reevaluate later if generalizing `debounce` and `throttle` is worth going after. :shipit:

PR #555, opened 2017-11-12
[3.0] Make ReactiveSwift a static framework.

andersio:
CocoaPods support is enabled via https://github.com/CocoaPods/CocoaPods/pull/6811, and is only effective when using CocoaPods 1.4.0 beta.

### DIFF
--- a/ReactiveSwift.podspec
+++ b/ReactiveSwift.podspec
@@ -17,6 +17,7 @@ Pod::Spec.new do |s|
   s.source       = { :git => "https://github.com/ReactiveCocoa/ReactiveSwift.git", :tag => "#{s.version}" }
   # Directory glob for all Swift files
   s.source_files  = "Sources/*.{swift}"
+  s.static_framework = true
   s.dependency 'Result', '~> 3.2'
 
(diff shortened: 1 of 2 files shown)


ikesyo:
Carthage does not fully support static frameworks, so this would break Carthage support. See https://github.com/Carthage/Carthage/issues/1379 and https://github.com/Carthage/Carthage/pull/1935.

andersio:
@ikesyo I am not quite sure how Carthage support would be broken. The end product being yielded is still a framework bundle, isn’t it?

ikesyo:
Carthage now builds only __dynamic_ frameworks, not static ones: https://github.com/Carthage/Carthage/blob/a3573d94f97458473b2d8ae0282cd6d34e9908a2/Source/CarthageKit/Xcode.swift#L277-L279

ikesyo:
See the CI job, no frameworks were built: https://travis-ci.org/ReactiveCocoa/ReactiveSwift/jobs/300898609

andersio:
Whoops. Thought the support has already been merged. 🙈

NachoSoto:
Even when it does support it, we should keep compatibility for both types of frameworks.

PR #563, opened 2017-11-20
Test cases for synchronous start with asynchronous interruption.

andersio:


Tests/ReactiveSwiftTests/SignalProducerSpec.swift
```diff
@@ -2916,6 +2916,53 @@ class SignalProducerSpec: QuickSpec {
...
+							}
+						}
+					}
+
+				expect(count.value) == expected
+			}
+
+			it("should fail the producer") {
+				sharedTest([SignalProducer(failure.output), numbers],
+						   expected: 2)
```
mdiep: Why should this be 2 and not 1?

I think this could use a comment in the code explaining the expectation.

Tests/ReactiveSwiftTests/SignalProducerSpec.swift
```diff
@@ -2916,6 +2916,53 @@ class SignalProducerSpec: QuickSpec {
 				expect(combined is SignalProducer<(Int, Double, Float, UInt), TestError>) == true
 			}
 		}
+
+		describe("interruption") {
```
mdiep: Maybe describe this as "synchronous start with asynchronous interruption"?

andersio:
I forgot to remove the label. These are actually valid test cases written for a bug I thought to be existing. 😅

andersio:
Included as part of #534 and addressed the comments there.

PR #564, opened 2017-11-22
Concat error

nmccann:
Given that `SignalProducer` already supports concatenating a value or a signal, being able to concatenate an error seemed logical. I personally found this useful when implementing a signal that defaulted to an error if nothing matched a filter. For example:

```
filtered.concat(error: notFound).take(first: 1)
```

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/Tests/ReactiveSwiftTests/FlattenSpec.swift
+++ b/Tests/ReactiveSwiftTests/FlattenSpec.swift
@@ -1011,6 +1011,28 @@ class FlattenSpec: QuickSpec {
 			}
 		}
 
+		describe("SignalProducer.concat(error:)") {
+			it("should emit final error") {
+				let (signal, observer) = SignalProducer<Int, TestError>.pipe()
+
+				let mergedSignals = signal.concat(error: TestError.default)
+
+				var results: [Result<Int, TestError>] = []
+				mergedSignals.startWithResult { results.append($0) }
+
+				observer.send(value: 1)
+				observer.send(value: 2)
+				observer.send(value: 3)
+				observer.sendCompleted()
+
+				expect(results).to(haveCount(4))
+				expect(results[0].value) == 1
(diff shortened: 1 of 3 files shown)


nmccann:
@NachoSoto good test suggestion! Will add that shortly.

Tests/ReactiveSwiftTests/FlattenSpec.swift
```diff
@@ -1011,6 +1011,42 @@ class FlattenSpec: QuickSpec {
...
+				observer.sendCompleted()
+
+				expect(results).to(haveCount(4))
+				expect(results[0].value) == 1
+				expect(results[1].value) == 2
+				expect(results[2].value) == 3
+				expect(results[3].error) == .default
+			}
+
+			it("should emit own error when present") {
```
nmccann: I'm not quite sure of the wording of this expectation. Thoughts?

andersio: "should not concat the error", perhaps?

nmccann: I feel like that contradicts the earlier expectation - unless the reader looks at the implementation. However in my usage of Quick, I try to write expectations that could stand-alone without the implementation and still make sense.

andersio: “should not concat a subsequent error if the earlier producer has failed”?

mdiep: "should not emit concatenated error to a failed producer"?

nmccann: Thank you for the help @andersio and @mdiep - I've settled on an alternative that I'm happy with.

PR #569, opened 2017-12-05
Update APIContracts.md

Qata:
Change the switch statement to Swift 4 syntax.

#### Checklist
- [ ] Updated CHANGELOG.md.

### DIFF
--- a/Documentation/APIContracts.md
+++ b/Documentation/APIContracts.md
@@ -502,16 +502,16 @@ For example:
 ```swift
 producer.start { event in
     switch event {
-    case let .Next(value):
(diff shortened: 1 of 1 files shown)


Documentation/APIContracts.md
```diff
@@ -502,16 +502,16 @@ For example:
 ```swift
 producer.start { event in
     switch event {
-    case let .Next(value):
+    case .value(let value):
```
andersio: Please place `let` before the case name. The RAS codebase has a convention of having value binding pattern indicator "globally" regardless of the number of bound values. :smily

PR #573, opened 2017-12-12
Lower binary ops as event transforms.

andersio:
Depends on #575.

Lower binary ops like how the unary ops already do.

Two pilots `take(during:)` and `take(until:)` are lowered here, and they are particularly important given how commonly recursive terminal events happen to them.


mdiep:
Would you mind breaking up this PR into a few smaller ones? This is pretty meaty, especially with the changes to `SignalCore`, which makes it difficult to review.

andersio:
@mdiep Done as #575.

PR #575, opened 2017-12-16
Separate the serialization and multicast logic in `Signal`.

andersio:
Break part of the Signal.Core synchronisation logic out, so that we may reuse it for:

1. lowering n-ary operators as event transforms e.g. #573; and
2. (hopefully) simplifying the `combineLatest` and `zip` implementations, which currently replicates part of the `Signal.Core` to deal with recursive terminal events.

### DIFF
--- a/Sources/Event.swift
+++ b/Sources/Event.swift
@@ -174,37 +174,42 @@ extension Signal.Event: EventProtocol {
 	}
 }
 
-// Event Transformations
-//
-// Operators backed by event transformations have such characteristics:
-//
-// 1. Unary
-//    The operator applies to only one stream.
-//
-// 2. Serial
-//    The outcome need not be synchronously emitted, but all events must be delivered in
-//    serial order.
-//
-// 3. No side effect upon interruption.
-//    The operator must not perform any side effect upon receving `interrupted`.
-//
-// Examples of ineligible operators (for now):
-//
-// 1. `timeout`
-//    This operator forwards the `failed` event on a different scheduler.
-//
-// 2. `combineLatest`
-//    This operator applies to two or more streams.
-//
-// 3. `SignalProducer.then`
-//    This operator starts a second stream when the first stream completes.
-//
-// 4. `on`
-//    This operator performs side effect upon interruption.
+extension Signal.Event {
+	/// Event Transformation
+	///
+	/// Given an output sink and a upstream lifetime, an event transformation
+	/// yields an input sink which, for every event received, evaluates certain
+	/// side effects that emits zero or more events to the given output sink.
+	///
+	/// Operators are obliged to maintain:
+	///
+	/// 1. Serial event order
+	///    The outcome need not be synchronously emitted, but every event must
+	///    be delivered exclusively in serial order.
+	///
+	/// 2. No side effect upon interruption.
+	///    The operator must not perform any side effect upon receving `interrupted`.
+	///
+	/// When implementing operators with event transformations, one must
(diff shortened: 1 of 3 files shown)


mdiep:
There looks to be a real test failure.

mdiep:
It may be missing this from `Signal.send`:

```swift
			if event.isTerminating {
				self.observers = nil
			}
```

Sources/Event.swift
```diff
@@ -832,3 +837,98 @@ extension Signal.Event where Value == Never {
...
+
+				sendLock.lock()
+				action(event)
+				sendLock.unlock()
+
+				tryToCommitTermination()
+			} else if status.tryTransition(from: .alive, to: .terminated) {
+				// Gracefully ignore any terminal event other than the one
+				// transitions the event stream away from the `alive` status.
+				guard !sendLock.try() else {
```
mdiep: I would invert this `guard`. I think it reads weirdly. The common case (unblocked termination) should be what we're guarding.

andersio:
I have made the observer bag cleaning a `Disposable`, since `self.send` in this case would be running inside the send lock.

mdiep:
That was just my guess as to why the test was failing.

Sources/Signal.swift
```diff
@@ -51,96 +55,55 @@ public final class Signal<Value, Error: Swift.Error> {
...
+			}
 
 			// The generator observer retains the `Signal` core.
-			generator(Observer(action: self.send, interruptsOnDeinit: true), Lifetime(disposable))
+			let sink = Event.makeSynchronizing(self.send, disposable: disposable)
+			let inputObserver = Observer(action: sink, interruptsOnDeinit: true)
+			generator(inputObserver, Lifetime(disposable))
 		}
 
 		private func send(_ event: Event) {
```
mdiep: I think we should remove this function, inlining it into the `init`. It's no longer safe to call this outside of `sink` above, so leaving it here risks future violations of the signal contract.

mdiep:
In #573, I don't see `makeSynchronizing` reused. Is the intent to make this reusable? Where will it be reused?

andersio:
It is here:
https://github.com/ReactiveCocoa/ReactiveSwift/pull/573/files#diff-ceb9f0d4c919bbf25ca4e3c8b7036b8bR254

andersio:
Not sure if this is the most appropriate way to fit it in. If we look ahead to a HKT future, we perhaps should rely on different types of transforms, instead of enum synchronisation strategies.

andersio:
Hmm, something is off.

andersio:
Fixed in https://github.com/ReactiveCocoa/ReactiveSwift/pull/575/commits/6b252d3866e309be9c8fc53fd36d2a4a44d9678a with refinements in the test case.

mdiep:
I'm not convinced that this is a win.

When I look at this PR by itself, it definitely doesn't seem like an improve its own right. If it makes #573 easier, then it could be worth doing, so I think the question is whether we want to take the 2 PRs.

When I look at #573, it looks very complex. I don't like the look of `SynchronizationStrategy`, and I'm not convinced that it's a net benefit to move all of those operators into event transforms. Maybe it'd be better to keep the separate code paths? Event transforms for simple transforms, `Signal` operators for more complicated ones.

I guess it feels like there are 2 separate types of operators here. Treating them in the same, or mostly-same way, sounds appealing, but isn't necessarily very clean. We might be better off keeping two different approaches. The approach for the more complicated operators doesn't need to be what we have today, but I think something that uses distinct APIs might bring more clarity.

PR #579, opened 2017-12-22
Restructure `SignalProducerCore.start` for clarity.

andersio:
I think we could agree that call sites of `SignalProducerCore.start` is pretty hard to grasp due to the "observer generator" thing.

It appears that we can follow the `SignalProducerCore.Instance` approach. That is, we shift the responsibility to start the producer instance to the callee, so that we have a sane imperative API here.

```swift
 	@discardableResult
 	public func start(_ observer: Signal<Value, Error>.Observer = .init()) -> Disposable {
 		// Looks like a transform, but not exactly a transform.
-		return core.start { _ in observer }

 		// Makes better sense, perhaps?
+		let (interrupter, start) = core.bootstrap(observer)
+		start()
+		return interrupter
 	}
```

#### Checklist
- ~~Updated CHANGELOG.md.~~

### DIFF
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -251,21 +251,30 @@ internal class SignalProducerCore<Value, Error: Swift.Error> {
 	}
 
 	func makeInstance() -> Instance {
-		fatalError()
+		let disposable = SerialDisposable()
+		let (signal, observer) = Signal<Value, Error>.pipe(disposable: disposable)
+
+		let (interrupter, didStart) = start(observer)
+		disposable.inner = interrupter
+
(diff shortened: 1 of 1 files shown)


mdiep:
This doesn't really seem any clearer to me. 😕

andersio:
Is it the naming or just syntactically it feels the same? I think we have limited design space for this kind of "create graph and run later" scenarios.

mdiep:
I think `start` makes more sense to me because it's similar to `SignalProducer.start`. It's easier for me to intuitively understand its role, even though the signature/specifics is a bit different.

PR #580, opened 2017-12-22
Refactor the MutableProperty deadlock test randomly failing on Linux.

### LINKED ISSUE #544: Intermittent test deadlocks on Linux builds
Some `PropertySpec` cases are deadlocking on Linux builds. As an example one can look into https://travis-ci.org/ReactiveCocoa/ReactiveSwift/jobs/291414085 where test run is deadlocking for
```
Test Case 'XCTestCase.MutableProperty, should not deadlock' started at 2017-10-23 08:37:59.778
```

Probably, some test revision needed, because from a first glance test case seems to be OK, but still deadlocking somehow.

andersio:
Hopefully resolves #544...

### DIFF
--- a/Tests/ReactiveSwiftTests/PropertySpec.swift
+++ b/Tests/ReactiveSwiftTests/PropertySpec.swift
@@ -275,27 +275,34 @@ class PropertySpec: QuickSpec {
 					queue = DispatchQueue.global(priority: .high)
 				}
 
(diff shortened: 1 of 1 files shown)


andersio:
I have rerun the Linux instance a few times, and it seems fine (at the moment it has a high chance of failure).

PR #581, opened 2017-12-22
Implement `lazyMap` as an event transformation.

andersio:
So that it abides to the _interruption on specified scheduler_ rule exhibited by its peer async operators.

#### Checklist
- ~Updated CHANGELOG.md.~

### DIFF
--- a/Sources/Event.swift
+++ b/Sources/Event.swift
@@ -699,6 +699,53 @@ extension Signal.Event {
 		}
 	}
 
+	internal static func lazyMap<U>(on scheduler: Scheduler, transform: @escaping (Value) -> U) -> Transformation<U, Error> {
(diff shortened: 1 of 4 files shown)


NachoSoto:
Does this fix a bug? If so it might be worth adding a test. Otherwise doesn't seem worth rewriting the operator from scratch…

andersio:
@NachoSoto Done in https://github.com/ReactiveCocoa/ReactiveSwift/pull/581/commits/c8e3a8cecdaf958f1f70f0d85fce80f17060bd76.

The majority of async operators forward all events to the specified scheduler, and the downstream initiated `interrupted` event has been respecting the specified scheduler since 2.0. So the `lazyMap` here is aligned to them for consistency.

Issue #582, opened 2017-12-26
Inconsistent disposal behavior between `QueueScheduler`'s `schedule(after:action:)` and `schedule(after:interval:action:)`


jjoelson:
As [this question on StackOverflow](https://stackoverflow.com/questions/47977290/queuescheduler-not-firing-when-interval-is-set) observes, the behavior of `schedule(after:action:)` is different from `schedule(after:interval:action:)` with respect to disposal. The former returns a `Disposable` that you *may* use to cancel the action before it occurs whereas the latter returns a `Disposable` that you *must* retain in order to avoid immediate cancellation. 

`schedule(after:interval:action:)`'s behavior feels strange, as there is nowhere else within `RAS` that I can think of where merely failing to *retain* a `Disposable` causes it to, in effect, be disposed. This looks to me as though `DispatchSourceTimer`'s behavior of cancelling upon `deinit` is leaking through. 

If this is a bug, ought `schedule(after:interval:action:)`  be implemented in terms of `schedule(after:action:)` (the way `TestScheduler` appears to do it)? And if not, might it make sense to remove the `@discardableResult` annotation to signal to users that they probably want to hold on to the returned `Disposable`? It doesn't seem to me as though cancelling the scheduled action immediately is behavior users would ever want.

(Apologies for not submitting a pull request per the contribution guidelines, but I am unsure of whether this is actually a bug or not. I am happy to submit a pull request if it is agreed that this is a bug!)

andersio:
I agree that this is not the expected behavior, since disposables are generally not supposed to exercise side effect unless being explicitly disposed of. It could very well be an oversight in Swift 3 migration.

jjoelson:
I noticed that `TestScheduler` implements it recursively in terms of `schedule(after:action:)`:
```
private func schedule(after date: Date, interval: DispatchTimeInterval, disposable: SerialDisposable, action: @escaping () -> Void) {
    precondition(interval.timeInterval >= 0)

    disposable.inner = schedule(after: date) { [unowned self] in
        action()
        self.schedule(after: date.addingTimeInterval(interval), interval: interval, disposable: disposable, action: action)
    }
}

public func schedule(after date: Date, interval: DispatchTimeInterval, leeway: DispatchTimeInterval = .seconds(0), action: @escaping () -> Void) -> Disposable? {
    let disposable = SerialDisposable()
    schedule(after: date, interval: interval, disposable: disposable, action: action)
    return disposable
}
```
[...]

PR #584, opened 2017-12-28
Fix `schedule(after:interval:leeway:)` disposal

jjoelson:
Addresses #582. Retain timers in `QueueScheduler` so that they will not be cancelled if the user doesn’t retain the returned `Disposable`.

Note that the test leaks a timer, but I'm not sure how else to test that the timer continues running even when the `Disposable` is not retained. 

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # master
 *Please add new entries at the top.*
 
-1. Fixed `schedule(after:interval:leeway:)` being cancelled when the returned `Disposable` is not retained.
+1. Fixed `schedule(after:interval:leeway:)` being cancelled when the returned `Disposable` is not retained. (#584, kudos to @jjoelson)
 
 # 3.1.0-rc.1
 1. Fixed a scenario of downstream interruptions being dropped. (#577, kudos to @andersio)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # master
 *Please add new entries at the top.*
 
+1. Fixed `schedule(after:interval:leeway:)` being cancelled when the returned `Disposable` is not retained.
+
 # 3.1.0-rc.1
 1. Fixed a scenario of downstream interruptions being dropped. (#577, kudos to @andersio)
 
--- a/Sources/Scheduler.swift
+++ b/Sources/Scheduler.swift
@@ -177,12 +177,15 @@ public final class UIScheduler: Scheduler {
 	}
 }
 
-/// A `Hashable` wrapper for `DispatchSourceTimer`.
+/// A `Hashable` wrapper for `DispatchSourceTimer`. `Hashable` conformance is
+/// based on the identity of the wrapper object rather than the wrapped
+/// `DispatchSourceTimer`, so two wrappers wrapping the same timer will *not*
+/// be equal.
 private final class DispatchSourceTimerWrapper: Hashable {
 	private let value: DispatchSourceTimer
 	
 	fileprivate var hashValue: Int {
-		return ObjectIdentifier(value).hashValue
+		return ObjectIdentifier(self).hashValue
 	}
 	
 	fileprivate init(_ value: DispatchSourceTimer) {
@@ -190,7 +193,7 @@ private final class DispatchSourceTimerWrapper: Hashable {
 	}
 	
 	fileprivate static func ==(lhs: DispatchSourceTimerWrapper, rhs: DispatchSourceTimerWrapper) -> Bool {
-		return lhs.value === rhs.value
+		return lhs === rhs
 	}
 }
 
--- a/Sources/Scheduler.swift
+++ b/Sources/Scheduler.swift
@@ -182,7 +182,7 @@ private final class DispatchSourceTimerWrapper: Hashable {
 	private let value: DispatchSourceTimer
 	
 	fileprivate var hashValue: Int {
-		return value.hash
+		return ObjectIdentifier(value).hashValue
 	}
 	
 	fileprivate init(_ value: DispatchSourceTimer) {
--- a/Sources/Scheduler.swift
+++ b/Sources/Scheduler.swift
@@ -177,6 +177,23 @@ public final class UIScheduler: Scheduler {
 	}
 }
 
+/// A `Hashable` wrapper for `DispatchSourceTimer`.
+private final class DispatchSourceTimerWrapper: Hashable {
+	private let value: DispatchSourceTimer
+	
+	fileprivate var hashValue: Int {
+		return value.hash
+	}
+	
+	fileprivate init(_ value: DispatchSourceTimer) {
+		self.value = value
+	}
+	
+	fileprivate static func ==(lhs: DispatchSourceTimerWrapper, rhs: DispatchSourceTimerWrapper) -> Bool {
+		return lhs.value === rhs.value
+	}
+}
+
 /// A scheduler backed by a serial GCD queue.
 public final class QueueScheduler: DateScheduler {
 	/// A singleton `QueueScheduler` that always targets the main thread's GCD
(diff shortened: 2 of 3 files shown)


Tests/ReactiveSwiftTests/SchedulerSpec.swift
```diff
@@ -218,6 +218,34 @@ class SchedulerSpec: QuickSpec {
...
+					var count = 0
+					let timesToIncrement = 3
+					
+					// Start two repeating timers, dispose the first, and ensure only the second runs.
+					
+					disposable.inner = scheduler.schedule(after: Date(), interval: .milliseconds(10), leeway: .seconds(0)) {
+						expect(false).to(equal(true), description: "timer not cancelled on disposal")
+					}
+					
+					scheduler.schedule(after: Date(), interval: .milliseconds(10), leeway: .seconds(0)) {
```
andersio: You might scope this in a function, so that the disposable is guaranteed to be deinitialised.

Tests/ReactiveSwiftTests/SchedulerSpec.swift
```diff
@@ -218,6 +218,34 @@ class SchedulerSpec: QuickSpec {
...
 					expect{count}.toEventually(equal(timesToRun))
 				}
+				
+				it("should repeatedly run actions after a given date when the disposable is not retained") {
+					let disposable = SerialDisposable()
+					
+					var count = 0
+					let timesToIncrement = 3
+					
+					// Start two repeating timers, dispose the first, and ensure only the second runs.
```
andersio: It'd probably be better off as two separate test cases. One for cancellation, while another for not retaining the disposable.

Sources/Scheduler.swift
```diff
@@ -193,8 +193,14 @@ public final class QueueScheduler: DateScheduler {
 
 	public let queue: DispatchQueue
 
+	private let lock: Lock
+	
+	private var timers: [DispatchSourceTimer]
```
andersio: You may use `Atomic`. I'd prefer `Set` which fits the scenario here (insert/removal only).

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # master
 *Please add new entries at the top.*
 
+1. Fixed `schedule(after:interval:leeway:)` being cancelled when the returned `Disposable` is not retained.
```
andersio: Please add the PR no & your handle.

Tests/ReactiveSwiftTests/SchedulerSpec.swift
```diff
@@ -218,6 +218,34 @@ class SchedulerSpec: QuickSpec {
...
+					
+					disposable.inner = scheduler.schedule(after: Date(), interval: .milliseconds(10), leeway: .seconds(0)) {
+						expect(false).to(equal(true), description: "timer not cancelled on disposal")
```
andersio: You may replace this with `fail("message")`.

Tests/ReactiveSwiftTests/SchedulerSpec.swift
```diff
@@ -218,6 +218,34 @@ class SchedulerSpec: QuickSpec {
...
+						if count <= timesToIncrement {
+							count += 1
+						}
+					}
+					
+					disposable.dispose()
+					
+					expect(count) == 0
+					
+					scheduler.queue.resume()
```
andersio: For the cancellation test, I'd wait for 1.5-2x the interval after resuming the queue so that the test case can actually fail.

Sources/Scheduler.swift
jjoelson: `DispatchSourceTimer` isn't `Hashable`, so I think it would have to be an `NSMutableSet`. Is that an improvement over using a Swift array?

andersio:
A private wrapper for `DispatchSourceTimer` should work. The hash value & the equality can be derived from the object identity with `===` and `ObjectIdentifier`.

Tests/ReactiveSwiftTests/SchedulerSpec.swift
```diff
@@ -218,6 +218,67 @@ class SchedulerSpec: QuickSpec {
...
+							}
+						}
+					}
+					
+					scheduleAndDeinitDisposable()
+					
+					expect(count) == 0
+					
+					scheduler.queue.resume()
+					expect{count}.toEventually(equal(timesToIncrement))
```
andersio: IIRC `expect` uses @autoclosure, so it is already implicitly captured.

Tests/ReactiveSwiftTests/SchedulerSpec.swift
```diff
@@ -218,6 +218,67 @@ class SchedulerSpec: QuickSpec {
...
+
+					disposable1.dispose()
+					
+					expect(count) == 0
+					
+					scheduler.queue.resume()
+					
+					// This expectation should take about 2.0 * interval to be fulfilled, and that's
+					// enough time to ensure that the first timer was actually cancelled.
+					expect{count}.toEventually(equal(timesToRun))
```
andersio: Ditto for @autoclosure.

andersio:
Hmm

```
ReactiveSwiftTests.SchedulerSpec
  QueueScheduler__on_a_given_queue__should_repeatedly_run_actions_after_a_given_date_when_the_disposable_is_not_retained, [31mexpected to eventually equal <3>, got <4>
```

jjoelson:
I think I see the bug, but strange that it’s not failing on my machine. I guess there’s a race condition where the expectation sees that it got incremented to 3 and succeeds before it gets incremented that last time. 

I will fix when I get back to my PC.

andersio:
CI has limited concurrency (1 logical processor IIRC).

jjoelson:
Is there a test still failing on Linux? I'm having a tough time understanding the log.

jjoelson:
OK, I found the error on Linux:
```
/home/travis/build/ReactiveCocoa/ReactiveSwift/Sources/Scheduler.swift:185:10: error: cannot invoke initializer for type 'ObjectIdentifier' with an argument list of type '(DispatchSourceTimer)'
                return ObjectIdentifier(value).hashValue
                       ^
/home/travis/build/ReactiveCocoa/ReactiveSwift/Sources/Scheduler.swift:185:10: note: overloads for 'ObjectIdentifier' exist with these partially matching parameter lists: (AnyObject), (Any.Type)
                return ObjectIdentifier(value).hashValue
                       ^
/home/travis/build/ReactiveCocoa/ReactiveSwift/Sources/Scheduler.swift:193:20: error: binary operator '===' cannot be applied to two 'DispatchSourceTimer' operands
                return lhs.value === rhs.value
                       ~~~~~~~~~ ^   ~~~~~~~~~
/home/travis/build/ReactiveCocoa/ReactiveSwift/Sources/Scheduler.swift:193:20: note: overloads for '===' exist with these partially matching parameter lists: (AnyObject?, AnyObject?), (NSNull?, NSNull?), (AnyClass, AnyClass)
                return lhs.value === rhs.value
```
Will fix when I get home tonight.

jjoelson:
Stumped by this one:

```
2018-01-02 23:49:31.857 xcodebuild[2551:9128] Error 2018-01-02 23:49:31.857 xcodebuild[2551:9128] Error Domain=IDETestOperationsObserverErrorDomain Code=6 "Early unexpected exit, operation never finished bootstrapping - no restart will be attempted" UserInfo={NSLocalizedDescription=Early unexpected exit, operation never finished bootstrapping - no restart will be attempted}

** TEST EXECUTE FAILED **
```

mdiep:
Usually that's downstream from whatever the actual error was.

jjoelson:
Seems to be working now. Judging by [this Stackoverflow thread](https://stackoverflow.com/a/36217200/642233), that's an error that can occur randomly.

PR #588, opened 2018-01-11
Update the playground.

andersio:


mdiep:
This is supposed to be covered by CI. Any idea why it isn't?

andersio:
CI probably didn't verify the UI Example playground, but only the primary playground.

mdiep:
Want to add this playground to CI too?

mdiep:
🤘

PR #594, opened 2018-01-30
Rename filterMap, skipNil to compactMap, compact

ra1028:
Rename filterMap to compactMap and skipNil to compact, based on the new API of Swift 4.1.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/Sources/Deprecations+Removals.swift
+++ b/Sources/Deprecations+Removals.swift
@@ -15,3 +15,31 @@ extension Lifetime {
 }
 
 // MARK: Deprecated types
+
+extension Signal {
+	@available(*, deprecated, renamed: "compactMap(_:)")
+	public func filterMap<U>(_ transform: @escaping (Value) -> U?) -> Signal<U, Error> {
+		return compactMap(transform)
+	}
+}
+
+extension Signal where Value: OptionalProtocol {
+	@available(*, deprecated, renamed: "compact()")
+	public func skipNil() -> Signal<Value.Wrapped, Error> {
+		return compact()
+	}
+}
+
+extension SignalProducer {
+	@available(*, deprecated, renamed: "compactMap(_:)")
+	public func filterMap<U>(_ transform: @escaping (Value) -> U?) -> SignalProducer<U, Error> {
+		return compactMap(transform)
+	}
+}
+
+extension SignalProducer where Value: OptionalProtocol {
+	@available(*, deprecated, renamed: "compact()")
+	public func skipNil() -> SignalProducer<Value.Wrapped, Error> {
+		return compact()
+	}
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # master
 *Please add new entries at the top.*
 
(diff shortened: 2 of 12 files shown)


mdiep:
Thanks for the pull request!

I'm definitely 👍 on `compactMap`.

But I'm not sure we should rename `skipNil` to `compact`. While I do value consistency with the Swift stdlib, I value consistency within ReactiveSwift even more. I think our naming convention of `skip*` makes a lot of sense and makes things easier for beginners and experts alike.

What does the rest of @ReactiveCocoa/reactiveswift think?

andersio:
I am against both. `compact` makes sense in the space & mass, but time cannot be compacted.

ra1028:
Thanks for the feedback.

I think it would be better to simply deemed event stream as sequence.
Differences in naming with Swift stdlib compromise the intuitiveness of the API.

ra1028:
> I think our naming convention of skip* makes a lot of sense and makes things easier for beginners and experts alike.

`skipNil` applies optional-binding to values of the stream, but other `skip*` operators are convenience methods only to skip values by conditions.
I think they are different.

NachoSoto:
> I think our naming convention of skip* makes a lot of sense and makes things easier for beginners and experts alike.

Agreed. As opposed to the `Array` counterparts, in the case of "streams", it's actually _skipping values over time_.

Sources/Deprecations+Removals.swift
```diff
@@ -15,3 +15,31 @@ extension Lifetime {
 }
 
 // MARK: Deprecated types
+
+extension Signal {
```
NachoSoto: Thanks for these 👍

ra1028:
Should I revert `compact()` to `skipNil()` ?
What does the rest of @ReactiveCocoa/reactiveswift think about `compactMap` ?

sharplet:
I'm a little torn on this. I _really_ like the idea of aligning with the standard library where possible, but when I read `signal.compact()` it feels kind of meaningless in comparison to `signal.skipNil()`. Now that I'm thinking about it, `signal.compactMap()` also reads strangely to me. I think I prefer our current names.

However, the types _are_ similar and may very well be what some new users reach for when they compare working with a stream of optional values to working with a `Sequence` or `Collection` of them. Maybe we can get the discoverability benefits by adding empty implementations of `compact()` and `compactMap()` with fixits that suggest the correct name (and ideally hint as to why)?

mdiep:
> Maybe we can get the discoverability benefits by adding empty implementations of compact() and compactMap() with fixits that suggest the correct name (and ideally hint as to why)?

👍

ra1028:
> Maybe we can get the discoverability benefits by adding empty implementations of compact() and compactMap() with fixits that suggest the correct name (and ideally hint as to why)?

Is that something like below?
```swift
extension Signal {
    @available(*, unavailable, message: "Use filterMap(_:) instead.")
    public func compactMap<U>(_ transform: @escaping (Value) -> U?) -> Signal<U, Error> {
        fatalError()
    }
}
```

Hmm, I feel not makes sense it...
Almost everyone seems not to agreed, so I'll close this proposal.
Thanks for the all feedbacks:+1:

Issue #597, opened 2018-02-02
Propose to add frequently used new operators


ra1028:
I would like to propose some of the frequently used operators for Signal and SignalProducer.
If willing to agree with me on this, so I'll send Pull Request.

## `void()`
Map unused values to Void.
```swift
func void() -> Signal<Void, Error>
```

## `skipError()`
Create a NoError signal, drop errors.
```swift
func skipError() -> Signal<Value, NoError>
```

##  `merge(with:)`
Merges `self` and given signal into a single `Signal`.
Similar functions( `combineLatest(with:)` , `zip(with:)` ) are already exists, but `merge(with:)` is not exist.
```swift
func merge(with other: Signal<Value, Error>) -> Signal<Value, Error>
```

NachoSoto:
- `void()`: I support this. I don’t like that name in particular, but I’m sure we can find a more precise one. In my codebase I have this as `var voidValues: Signal<Void, E> {}`, but I don’t love that name either.
- `skipError`: this has been suggested many times, but I’m strongly against it. It encourages bad practices by simply ignoring errors. Instead, each user should implement this as they see fit. For example, in my codebase I have `logAndIgnoreErrors()`, which is only exposed for `Error: ReportableError` (my own type), and it sends _non-fatal_ errors to Crashlytics. Ignoring them without any consequence is a bad practice IMO.
- `merge(with:)`: I have this same one in my codebase, so 👍

mdiep:
> `void()`: I support this. I don’t like that name in particular, but I’m sure we can find a more precise one. In my codebase I have this as `var voidValues: Signal<Void, E> {}`, but I don’t love that name either.

Agreed. I think we just need the right name. We've previously referred to `Signal<(), E>` as a _trigger_, so that might be a clue towards a good name.

I've been wanting to add a `func map<NewValue>(value: NewValue) -> Signal<NewValue, Error>` to map to constant values. Maybe that'd be enough for this? `.map(value: ())`

> `skipError`: this has been suggested many times, but I’m strongly against it.

Agreed. 👎 

> `merge(with:)`: I have this same one in my codebase, so 👍

👍

NachoSoto:
> I've been wanting to add a func map<NewValue>(value: NewValue) -> Signal<NewValue, Error> to map to constant values. Maybe that'd be enough for this? .map(value: ())

[...]

PR #600, opened 2018-02-06
Add merge(with:) operator

ra1028:
Refer https://github.com/ReactiveCocoa/ReactiveSwift/issues/597

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/Sources/Flatten.swift
+++ b/Sources/Flatten.swift
@@ -481,11 +481,8 @@ extension UnsafeAtomicState where State == ProducerState {
 }
 
 extension Signal {
-	/// Merges the given signals into a single `Signal` that will emit all
-	/// values from each of them, and complete when all of them have completed.
-	///
-	/// - parameters:
-	///   - signals: A sequence of signals to merge.
+	/// Merges the values of all the given signals, in the manner described
+	/// by `merge(with:)`.
 	public static func merge<Seq: Sequence>(_ signals: Seq) -> Signal<Value, Error> where Seq.Iterator.Element == Signal<Value, Error>
 	{
 		return SignalProducer<Signal<Value, Error>, Error>(signals)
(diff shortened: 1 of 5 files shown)


Sources/Signal.swift
```diff
@@ -788,6 +788,17 @@ extension Signal {
...
+	/// Merge the given signal into a single `Signal` that will emit all
+	/// values from both of them, and complete when all of them have completed.
+	///
+	/// - parameters:
+	///   - other: A signal to merge `self`'s value with.
+	///
+	/// - returns: A signal that sends all values of `self` and given signal.
+	public func merge(with other: Signal<Value, Error>) -> Signal<Value, Error> {
+		return Signal.merge(self, other)
+	}
```
mdiep: I think we still ought to have a test for this (and the `SignalProducer` variant below). Would you mind adding one?

Sources/Flatten.swift
```diff
@@ -481,46 +481,32 @@ extension UnsafeAtomicState where State == ProducerState {
 }
 
 extension Signal {
-	/// Merges the given signals into a single `Signal` that will emit all
-	/// values from each of them, and complete when all of them have completed.
-	///
-	/// - parameters:
-	///   - signals: A sequence of signals to merge.
+	/// Merges the values of all the given signals, in the manner described
+	/// by `merge(with:)`.
```
NachoSoto: Normally I wouldn't mind this, but Xcode does not make it easy to find this documentation, so I'd keep it duplicated.

ra1028: hmm, I was according to `zip`, `combineLatest` operator, though should revert?

ra1028:
I added tests

Sources/Flatten.swift
NachoSoto: We should duplicate it in those too IMO (at least until SourceKit & CMD+Click actually works consistently in Xcode). For now can you keep these how they were?

Tests/ReactiveSwiftTests/FlattenSpec.swift
```diff
@@ -820,8 +820,9 @@ class FlattenSpec: QuickSpec {
 			it("should emit values from all signals") {
 				let (signal1, observer1) = Signal<Int, NoError>.pipe()
 				let (signal2, observer2) = Signal<Int, NoError>.pipe()
+				let (signal3, observer3) = Signal<Int, NoError>.pipe()
 
-				let mergedSignals = Signal.merge([signal1, signal2])
+				let mergedSignals = Signal.merge([signal1, signal2, signal3])
```
NachoSoto: Thanks for making this more complex now 👍

Sources/Flatten.swift
ra1028: gotcha, will revert them:+1:
Write the doc for `zip` and` combineLatest`, will too much lines.
I think there are pros and cons, so not to write now.

mdiep:
Awesome! Thanks for adding this. 🤘

PR #601, opened 2018-02-06
Add map(to:) operator

ra1028:
Refer #597

For example, use as below.

```swift
let (signal, observer) = Signal<String, NoError>.pipe()
let triggerSignal = signal.map(to: ())

triggerSignal.observeValues {
    print("trigger `value` events")
}
```

#### Discassion
I want to add `@autoclosure @escaping`, but occur an error(`Ambiguous reference to member 'map`) on overload `map`.
So, I propose to name it `replace(to:)`, like below.

```swift
public func replace<U>(to value: @autoclosure @escaping () -> U) -> Signal<U, Error> {
    return map { _ in value() }
}
```

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/Sources/Property.swift
+++ b/Sources/Property.swift
@@ -101,6 +101,16 @@ extension PropertyProtocol {
 	public func map<U>(_ transform: @escaping (Value) -> U) -> Property<U> {
 		return lift { $0.map(transform) }
 	}
+	
+	/// Map the current value and all susequent values to a new constant property.
+	///
+	/// - parameters:
+	///   - value: A new value.
+	///
+	/// - returns: A property that holds a mapped value from `self`.
+	public func map<U>(value: U) -> Property<U> {
+		return map { _ in value }
+	}
 
 	/// Maps the current value and all subsequent values to a new property
 	/// by applying a key path.
--- a/Sources/Property.swift
+++ b/Sources/Property.swift
@@ -109,7 +109,7 @@ extension PropertyProtocol {
 	///
 	/// - returns: A property that holds a mapped value from `self`.
 	public func map<U>(value: U) -> Property<U> {
-		return map { _ in value }
+		return lift { $0.map(value: value) }
 	}
 
 	/// Maps the current value and all subsequent values to a new property
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -564,7 +564,7 @@ extension Signal {
 	///   - value: A new value.
 	///
 	/// - returns: A signal that will send new values.
-	public func map<U>(to value: U) -> Signal<U, Error> {
+	public func map<U>(value: U) -> Signal<U, Error> {
 		return map { _ in value }
 	}
 
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -557,6 +557,16 @@ extension Signal {
 	public func map<U>(_ transform: @escaping (Value) -> U) -> Signal<U, Error> {
 		return flatMapEvent(Signal.Event.map(transform))
 	}
+	
+	/// Map each value in the signal to a new constant value.
+	///
+	/// - parameters:
+	///   - value: A new value.
+	///
+	/// - returns: A signal that will send new values.
+	public func map<U>(to value: U) -> Signal<U, Error> {
+		return map { _ in value }
+	}
 
 	/// Map each value in the signal to a new value by applying a key path.
 	///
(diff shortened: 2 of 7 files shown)


Sources/Signal.swift
```diff
@@ -557,6 +557,16 @@ extension Signal {
...
 		return flatMapEvent(Signal.Event.map(transform))
 	}
+	
+	/// Map each value in the signal to a new constant value.
+	///
+	/// - parameters:
+	///   - value: A new value.
+	///
+	/// - returns: A signal that will send new values.
+	public func map<U>(to value: U) -> Signal<U, Error> {
```
NachoSoto: Can you add this to `SignalProducer` too?

NachoSoto:
I think eager evaluation is precisely what we want. If somebody needs lazy evaluation of the value they can make that explicit by using the closure variant.

Sources/Signal.swift
mdiep: I think we should call this `map(value:)` for consistency. This would match `SignalProducer(value:)`, but it would also match things like `skip(value:)`, `filter(value:)`, which wouldn't make sense with `to:`. We could also conceivably add `mapError(value:)`, `map(error:)` and other things.

ra1028: oh, I forgot
will add it.

Sources/Property.swift
```diff
@@ -101,6 +101,16 @@ extension PropertyProtocol {
...
 	}
+	
+	/// Map the current value and all susequent values to a new constant property.
+	///
+	/// - parameters:
+	///   - value: A new value.
+	///
+	/// - returns: A property that holds a mapped value from `self`.
+	public func map<U>(value: U) -> Property<U> {
+		return map { _ in value }
```
NachoSoto: Can you implement this using `lift` as well?

Sources/SignalProducer.swift
```diff
@@ -857,6 +857,17 @@ extension SignalProducer {
...
+	
+	/// Map each value in the producer to a new constant value.
+	///
+	/// - parameters:
+	///   - value: A new value.
+	///
+	/// - returns: A signal producer that, when started, will send a mapped
+	///            value of `self`.
+	public func map<U>(value: U) -> SignalProducer<U, Error> {
+		return map { _ in value }
```
NachoSoto: Can you implement this using `lift`?

ra1028: done

NachoSoto:
🎉

NachoSoto:
@mishagray realized that this produced a regression. I wrote a test case that reproduces it:
```swift
let mappedSignal = signal.map { $0 as Int? }
	
let expectedType: Signal<Int?, NoError> = mappedSignal
```

I'm sending a PR to fix this.

mishagray:
This PR breaks some of my existing 'map' code.
```
        // I have a Signal Producer of UIImage values
	let fetchImage = SignalProducer<UIImage, NoError>(value: UIImage())

        // optionalFetch SHOULD resolve as  SignalProducer<UIImage?, NoError>
	let optionalFetch = fetchImage.map { $0 as UIImage? }

	assert(optionalFetch is SignalProducer<UIImage?, NoError>, "did I get the right type?") 
        // The line above actually gives the compiler warning: "Cast from 'SignalProducer<(UIImage?) -> UIImage?, NoError>' to unrelated type 'SignalProducer<UIImage?, NoError>' always fails"
	assert(!(optionalFetch is SignalProducer<(UIImage?) -> UIImage?, NoError>), "did I get the wrong type?")
       // line above generates compiler warning: "'is' test is always true"

       // Defining the type seems to get the compiler to 'pick' the right version of 'map'.   But this kind of ugly.
	let optionalFetch1: SignalProducer<UIImage?, NoError>  = fetchImage.map { $0 as UIImage? }
	let optionalFetch2: SignalProducer<(UIImage?) -> UIImage?, NoError>  = fetchImage.map { $0 as UIImage? }
       	assert(optionalFetch1 is SignalProducer<UIImage?, NoError>, "did I get the right type?") 
 


```

NachoSoto:
I'm proposing renaming this `replacingValues(to:)` to avoid the bad type inference due to the overload.

mdiep:
Wat! That does _not_ seem like a valid use of trailing closure syntax. 😕

NachoSoto:
But Swift doesn't know that :(

NachoSoto:
Oh wait, I see what you're saying! That seems like a Swift bug then.

mdiep:
Yes, about to file an issue.

NachoSoto:
screen shot 2018-02-12 at 4 42 37 pm

mdiep:
https://bugs.swift.org/browse/SR-6989

ra1028:
Oh... 🙇

PR #602, opened 2018-02-09
Bag performance

ra1028:
1. Simplify nested code of `Bag.remove(using:)`
  Use `Sequence.first(where:)`
  Can be expected to gain a benefit from future Swift performance improvements by using the standard API.

2. Not use `revert()` for removing elements from bag.
  Improve dispose efficiency in `CompositeDisposable`.

#### Checklist
~- [ ] Updated CHANGELOG.md.~

### DIFF
--- a/Sources/Bag.swift
+++ b/Sources/Bag.swift
@@ -67,36 +67,12 @@ extension Bag: RandomAccessCollection {
 	public var endIndex: Int {
 		return elements.endIndex
 	}
-
-	public subscript(index: Int) -> Element {
-		return elements[index]
+	
+	public func index(after i: Int) -> Int {
+		return i + 1
 	}
-
-	public func makeIterator() -> Iterator {
-		return Iterator(elements)
-	}
-
-	/// An iterator of `Bag`.
-	public struct Iterator: IteratorProtocol {
-		private let base: ContiguousArray<Element>
-		private var nextIndex: Int
-		private let endIndex: Int
-
-		fileprivate init(_ base: ContiguousArray<Element>) {
-			self.base = base
-			nextIndex = base.startIndex
-			endIndex = base.endIndex
-		}
-
-		public mutating func next() -> Element? {
-			let currentIndex = nextIndex
-
-			if currentIndex < endIndex {
-				nextIndex = currentIndex + 1
-				return base[currentIndex]
-			}
-
-			return nil
-		}
+	
+	public subscript(position: Int) -> Element {
+		return elements[position]
 	}
 }
--- a/Sources/Bag.swift
+++ b/Sources/Bag.swift
@@ -8,46 +8,46 @@
 
 /// An unordered, non-unique collection of values of type `Element`.
 public struct Bag<Element> {
-    /// A uniquely identifying token for removing a value that was inserted into a
-    /// Bag.
-    public struct Token {
-        fileprivate let value: UInt64
-    }
-
-    fileprivate var elements: ContiguousArray<Element> = []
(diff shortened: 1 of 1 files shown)


ra1028:
@andersio 

Sorry, my test was lacked🙇 
I reverted it to use `Iterator`.

And I have the one more proposal.
I've suspected the necessity of `reversed()` in [this line](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/Bag.swift#L51).
I think that `Disposable` is disposed by `CompositeDisposable`(mostly `Lifetime`) in many cases.
In that cases, disposables will disposed in order from the beginning, so higher performance maybe expected if not to use `reversed()`.
I demonstrated it with a simple benchmark such as below.

```swift
func testRemovingByCompositeDisposable() {
    let count = 10_000

    measure {
        var bag = Bag<Int>()
        let disposable = CompositeDisposable()

        for i in 0..<count {
            let token = bag.insert(i)
            disposable += { bag.remove(using: token) }
        }

        XCTAssertEqual(bag.count, count)

        disposable.dispose()

        XCTAssertEqual(bag.count, 0)
    }
}
```

In this case, it has improved about 3x.
screen shot 2018-02-13 at 2 04 05

Also, I demonstrated that using `first(where:)` does not affect performance by the tests such as below.

```swift
func testRemoving() {
    let count = 10_000
    
    measure {
        var bag = Bag<Int>()
        var tokens = ContiguousArray<Bag<Int>.Token>()
        tokens.reserveCapacity(count)
        
        for i in 0..<count {
            let token = bag.insert(i)
            tokens.append(token)
        }
        
        XCTAssertEqual(bag.count, count)
        
        
        for token in tokens {
            bag.remove(using: token)
        }
        
        XCTAssertEqual(bag.count, 0)
    }
}

func testIteration() {
    let count = 10_000_000
    var bag = Bag<Int>()
    
    for i in 0..<count {
        bag.insert(i)
    }
    
    XCTAssertEqual(bag.count, count)
    
    measure {
        var trash = [Int]()
        trash.reserveCapacity(count)
        
        for i in bag {
            trash.append(i)
        }
        
        XCTAssertEqual(bag.count, count)
        XCTAssertEqual(trash.count, count)
    }
}
```
screen shot 2018-02-13 at 2 27 04

andersio:
> In that cases, disposables will disposed in order from the beginning, so higher performance maybe expected if not to use reversed().

Is the case still hold if you more or less randomise the insertion?

ra1028:
@andersio 
Thanks👍  It's already reflected.

Sources/Bag.swift
```diff
@@ -13,14 +13,14 @@ public struct Bag<Element> {
 	public struct Token {
 		fileprivate let value: UInt64
 	}
-
+
```
ikesyo: Could you please remove those white space changes?

ra1028: done

Sources/Bag.swift
```diff
@@ -48,14 +48,12 @@ public struct Bag<Element> {
 	///   - token: A token returned from a call to `insert()`.
 	@discardableResult
 	public mutating func remove(using token: Token) -> Element? {
-		for i in elements.indices.reversed() {
-			if tokens[i] == token.value {
-				tokens.remove(at: i)
-				return elements.remove(at: i)
-			}
+		guard let i = indices.first(where: { tokens[$0] == token.value }) else {
```
andersio: [Nit] Could you replace `i` with `index`?

ra1028: 👍

PR #603, opened 2018-02-12
Added back in swift 4.0 compatibility. 

mishagray:
This was my swift 4.0 compatibility strategy... 
Haven't done much more than compile in box XCode 9.2 and XCode 9.3 and some small test of my app.   But if any of this code is useful for this branch... here it is..

#### Checklist
- [ ] Updated CHANGELOG.md.

### DIFF
--- a/Sources/Atomic.swift
+++ b/Sources/Atomic.swift
@@ -11,6 +11,16 @@ import Foundation
 import MachO
 #endif
 
+#if swift(>=4.1)
+#else
+    extension UnsafeMutablePointer {
+        private func deallocate() {
+            self.deallocate(capacity: 1)
+        }
+    }
+#endif
+
+
 /// A simple, generic lock-free finite state machine.
 ///
 /// - warning: `deinitialize` must be called to dispose of the consumed memory.
--- a/Sources/Event.swift
+++ b/Sources/Event.swift
@@ -118,7 +118,7 @@ extension Signal {
 	}
 }
 
-extension Signal.Event: Equatable where Value: Equatable, Error: Equatable {
+extension Signal.Event where Value: Equatable, Error: Equatable {
 	public static func == (lhs: Signal<Value, Error>.Event, rhs: Signal<Value, Error>.Event) -> Bool {
 		switch (lhs, rhs) {
 		case let (.value(left), .value(right)):
@@ -138,6 +138,9 @@ extension Signal.Event: Equatable where Value: Equatable, Error: Equatable {
 		}
 	}
 }
+#if swift(>=4.1)
+extension Signal.Event: Equatable where Value: Equatable, Error: Equatable {}
+#endif
 
 extension Signal.Event: CustomStringConvertible {
 	public var description: String {
--- a/Sources/Property.swift
+++ b/Sources/Property.swift
@@ -9,7 +9,31 @@ import enum Result.NoError
 ///
 /// Only classes can conform to this protocol, because having a signal
 /// for changes over time implies the origin must have a unique identity.
-public protocol PropertyProtocol: class, BindingSource {
+#if swift(>=4.1)
+	public protocol PropertyProtocol: class, BindingSource {
+		/// The current value of the property.
+		var value: Value { get }
+
+		/// The values producer of the property.
+		///
+		/// It produces a signal that sends the property's current value,
+		/// followed by all changes over time. It completes when the property
+		/// has deinitialized, or has no further change.
+		///
+		/// - note: If `self` is a composed property, the producer would be
+		///         bound to the lifetime of its sources.
+		var producer: SignalProducer<Value, NoError> { get }
+
+		/// A signal that will send the property's changes over time. It
+		/// completes when the property has deinitialized, or has no further
+		/// change.
+		///
+		/// - note: If `self` is a composed property, the signal would be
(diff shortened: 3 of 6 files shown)


Sources/Atomic.swift
```diff
@@ -11,6 +11,16 @@ import Foundation
 import MachO
 #endif
 
+#if swift(>=4.1)
+#else
```
NachoSoto: You can do `#if !swift(>=4.1)`.

Sources/Event.swift
```diff
@@ -138,6 +138,9 @@ extension Signal.Event: Equatable where Value: Equatable, Error: Equatable {
 		}
 	}
 }
+#if swift(>=4.1)
+extension Signal.Event: Equatable where Value: Equatable, Error: Equatable {}
```
NachoSoto: 👍

Sources/Property.swift
```diff
@@ -9,7 +9,31 @@ import enum Result.NoError
...
+		/// A signal that will send the property's changes over time. It
+		/// completes when the property has deinitialized, or has no further
+		/// change.
+		///
+		/// - note: If `self` is a composed property, the signal would be
+		///         bound to the lifetime of its sources.
+		var signal: Signal<Value, NoError> { get }
+	}
+#else
+	public protocol PropertyProtocol: class, BindingSource where Error == NoError {
```
NachoSoto: This is annoying, but I guess we have no alternative.

Sources/Property.swift
```diff
@@ -9,7 +9,31 @@ import enum Result.NoError
 ///
 /// Only classes can conform to this protocol, because having a signal
 /// for changes over time implies the origin must have a unique identity.
-public protocol PropertyProtocol: class, BindingSource {
+#if swift(>=4.1)
+	public protocol PropertyProtocol: class, BindingSource {
```
NachoSoto: Nit: I would avoid this indentation to simplify the diff.

Sources/UnidirectionalBinding.swift
```diff
@@ -52,16 +70,29 @@ extension BindingTargetProvider {
...
-		<Source: BindingSource>
-		(provider: Self, source: Source) -> Disposable?
-		where Source.Value == Value, Source.Error == NoError
-	{
-		return source.producer
-			.take(during: provider.bindingTarget.lifetime)
-			.startWithValues(provider.bindingTarget.action)
-	}
+	#if swift(>=4.1)
+		@discardableResult
```
NachoSoto: Ditto indentation.

Sources/Property.swift
mishagray: agreed...

Sources/Property.swift
mishagray: yeah... I checked other places to see if we wanted to 'indent' around a #if... and it's not consistent.  I think you indent inside a class/struct definition, but idk.  I DID try and match the style here!

Sources/Property.swift
```diff
@@ -5,11 +5,8 @@ import Glibc
 #endif
 import enum Result.NoError
 
-/// Represents a property that allows observation of its changes.
-///
-/// Only classes can conform to this protocol, because having a signal
-/// for changes over time implies the origin must have a unique identity.
-public protocol PropertyProtocol: class, BindingSource {
+
+public protocol BasePropertyProtocol: class, BindingSource {
```
mishagray: So I thought using some intermediate 'base' protocol would help reduce the difference between swift 4.0 and swift 4.1.  But it introduces a new public Protocol name.  Not sure if this adds complexity or reduces it in the whole scope of things.

Sources/Property.swift
```diff
@@ -32,6 +29,16 @@ public protocol PropertyProtocol: class, BindingSource {
 	var signal: Signal<Value, NoError> { get }
 }
 
+/// Represents a property that allows observation of its changes.
+///
+/// Only classes can conform to this protocol, because having a signal
+/// for changes over time implies the origin must have a unique identity.
```
NachoSoto: I think you need this docstring next to the type so that it gets exposed to clients (and duplicated).

Sources/UnidirectionalBinding.swift
```diff
@@ -11,9 +11,28 @@ precedencegroup BindingPrecedence {
...
+#if swift(>=4.1)
+/// Describes a source which can be bound.
+public protocol BindingSource: SignalProducerConvertible where Error == NoError {}
+extension Signal: BindingSource where Error == NoError {}
+extension SignalProducer: BindingSource where Error == NoError {}
+
+#else
+/// Describes a source which can be bound.
+public protocol BindingSource: SignalProducerConvertible {
+	// FIXME: Swift 4 compiler regression.
```
NachoSoto: Please remove this FIXME since it's already addressed.

Sources/Property.swift
NachoSoto: I think this makes sense 👍

NachoSoto: I think this makes sense, but I would probably name this `_BasePropertyProtocol` to indicate that it's not a type that's meant to be used outside the codebase (just like Swift's stdlib does when they hit language limitations), especially considering that we'll be removing this workaround when we drop `Swift 4.0.x` support.

PR #604, opened 2018-02-13
Fix Signal.map(value:) regression: changed to .replacingValues(with:) to avoid potentially incorrect type inference

NachoSoto:
See conversation in #601. The `.map` overload lead to potentially bad type inference:
```swift

let mappedSignal = signal.map { $0 as Int? }
// // Cannot convert value of type 'Signal<(Int?) -> Int?, NoError>' to specified type 'Signal<Int?, NoError>'
let expectedType: Signal<Int?, NoError> = mappedSignal
```

#### Checklist
- [x] Updated CHANGELOG.md.
- [x] Add regression test
- [x] Fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 1. New property operator: `filter` (#586, kudos to @iv-mexx)
 1. New operator `merge(with:)` (#600, kudos to @ra1028)
-1. New operator `map(value:)` (#601, kudos to @ra1028)
+1. New operator `replacingValues(with:)` (#601, #604, kudos to @ra1028, @NachoSoto)
 
 # 3.1.0
 1. Fixed `schedule(after:interval:leeway:)` being cancelled when the returned `Disposable` is not retained. (#584, kudos to @jjoelson)
--- a/Sources/Property.swift
+++ b/Sources/Property.swift
@@ -108,8 +108,8 @@ extension PropertyProtocol {
 	///   - value: A new value.
 	///
 	/// - returns: A property that holds a mapped value from `self`.
-	public func map<U>(value: U) -> Property<U> {
-		return lift { $0.map(value: value) }
+	public func replacingValues<U>(with value: U) -> Property<U> {
+		return lift { $0.replacingValues(with: value) }
 	}
 
 	/// Maps the current value and all subsequent values to a new property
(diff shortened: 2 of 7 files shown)


NachoSoto:
This seems like a Swift bug btw:
screen shot 2018-02-12 at 4 42 37 pm

I believe @mdiep is filing a Jira?

mdiep:
This seems to specifically be an issue with the block `{ $0 as Int? }`. This works fine:

```swift

let mappedSignal = signal.map { Optional($0) }
// // Cannot convert value of type 'Signal<(Int?) -> Int?, NoError>' to specified type 'Signal<Int?, NoError>'
let expectedType: Signal<Int?, NoError> = mappedSignal
```

So depending on what response we get to SR-6989, I might favor leaving it as `map(value:)`.

ra1028:
Thanks @NachoSoto.

> So depending on what response we get to SR-6989, I might favor leaving it as map(value:).

I Agreed.
But, If you will rename it, I think `repace(value:)` is better.

NachoSoto:
Whatever we do, it's important that we don't release a `RAS` update though, unless we workaround it. Based on the discussion on the Jira so far, I'd be surprised it even gets fixed in `Swift 4.1`.

NachoSoto:
Although, if this only happens with that particular type of closure (`{ $0 as Type?}`), we might want to ignore it, since it's "workaroundable" for clients by giving it an explicit type?

mdiep:
I'm in favor of ignoring it since (1) it only relates to specific closures, (2) it results in an error, not incorrect behavior, and (3) clients can work around it by adding explicit types.

NachoSoto:
Sounds good. I'll update the PR to keep the "negative" test so we can detect when (if?) this is fixed, and at least keep the behaviour documented.

andersio:
ping @NachoSoto

andersio:
I believe this could be mitigated by adding a `map` overload that takes `(Value) -> U?` for now. cc @mdiep @NachoSoto

andersio:
I do have to pose an objection to ignoring the issue, because giving explicit types is unergonomic for composition.

andersio:
Closing due to inactivity, which seems a legit sign of the issue having been mitigated successfully by the type checker fixes.

Issue #287, opened 2017-02-28
debounce/throttle operator that always emits the last value


NachoSoto:
I just found myself needing such an operator. I have a `SignalProducer<[Data], NoError>` created with this custom operator:
```swift
public func collectIncrementally() -> Signal<[Value], Error> {
    return self
        .map { [$0] }
        .scan([], +)
}
```

I want to limit how often I get updated arrays, but I do want to make sure that the last value (which will contain all the elements), is always emitted.

NachoSoto:
Actually I misread the documentation, but I'm still not 100% sure either `debounce` or `throttle` will behave this way, so I guess this is a question now, not a feature request :P

NachoSoto:
Yeah, it does look like if the source signal terminates **before** the given interval, no more cached values are sent. This is not documented, and I'm not sure if it's desired behavior. Either way, it would be nice to have an operator that doesn't behave this way.
I looked at `Rx`'s versions and they don't seem to have this either.

andersio:
It was documented before the revision. Guess we should put it back, alongside an option. A `discardsWhenTerminated` label, perhaps?

https://github.com/ReactiveCocoa/ReactiveSwift/commit/5f7109d57306f9204030172f3e8b6df37b1e5f1a#diff-09166f905ad1f830ac2dc91f2663c1a5L1728

NachoSoto:
> Guess we should put it back, alongside an option. A discardsWhenTerminated parameter with a default of true, perhaps?

That makes sense to me 👍

Qata:
@andersio Would it be welcome if I picked up this issue and created an implementation?

andersio:
@Qata Sure. 👍

RuiAAPeres:
Hello. 👋  Thanks for opening this issue. Due to inactivity, we will soft close the issue. If you feel that it should remain open, please let us know. 😄

PR #605, opened 2018-02-13
Add `discardWhenCompleted` to `debounce`

Qata:
#### Checklist
- [X] Updated CHANGELOG.md.

This PR adds a new option to `debounce` to allow sending values upon completion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # master
 *Please add new entries at the top.*
 
+1. New `debounce` parameter: `discardsWhenTerminated` (#287, kudos to @Qata)
 1. New property operator: `filter` (#586, kudos to @iv-mexx)
 1. New operator `merge(with:)` (#600, kudos to @ra1028)
 1. New operator `map(value:)` (#601, kudos to @ra1028)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # master
 *Please add new entries at the top.*
 
-1. New `debounce` parameter: `discardsWhenTerminated` (#287, kudos to @Qata)
+1. `debounce` now offers an opt-in behaviour to preserve the pending value when the signal or producer completes. You may enable it by specifying `discardsWhenCompleted` as false (#287, kudos to @Qata)
 1. Result now interoperates with SignalProducer n-ary operators as a constant producer (#606, kudos to @Qata)
 1. New property operator: `filter` (#586, kudos to @iv-mexx)
 1. New operator `merge(with:)` (#600, kudos to @ra1028)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # master
 *Please add new entries at the top.*
 
-1. `debounce` now offers an opt-in behaviour to preserve the pending value when the signal or producer completes. You may enable it by specifying `discardsWhenCompleted` as false (#287, kudos to @Qata)
+1. `debounce` now offers an opt-in behaviour to preserve the pending value when the signal or producer completes. You may enable it by specifying `discardWhenCompleted` as false (#287, kudos to @Qata)
 1. Result now interoperates with SignalProducer n-ary operators as a constant producer (#606, kudos to @Qata)
 1. New property operator: `filter` (#586, kudos to @iv-mexx)
 1. New operator `merge(with:)` (#600, kudos to @ra1028)
--- a/Sources/Event.swift
+++ b/Sources/Event.swift
@@ -816,17 +816,7 @@ extension Signal.Event {
 				}
 
 				schedulerDisposable.inner = scheduler.schedule(after: scheduleDate) {
-					let pendingValue: Value? = state.modify { state in
-						defer {
-							if state.pendingValue != nil {
-								state.pendingValue = nil
-								state.previousDate = scheduleDate
-							}
-						}
-						return state.pendingValue
-					}
-
-					if let pendingValue = pendingValue {
+					if let pendingValue = state.modify({ $0.retrieveValue(date: scheduleDate) }) {
 						action(.value(pendingValue))
 					}
 				}
@@ -834,7 +824,7 @@ extension Signal.Event {
 		}
 	}
 
-	internal static func debounce(_ interval: TimeInterval, on scheduler: DateScheduler, discardsWhenTerminated: Bool) -> Transformation<Value, Error> {
+	internal static func debounce(_ interval: TimeInterval, on scheduler: DateScheduler, discardsWhenCompleted: Bool) -> Transformation<Value, Error> {
 		precondition(interval >= 0)
 		
 		let state: Atomic<ThrottleState<Value>> = Atomic(ThrottleState(previousDate: scheduler.currentDate, pendingValue: nil))
@@ -855,20 +845,12 @@ extension Signal.Event {
 					}
 					let date = scheduler.currentDate.addingTimeInterval(interval)
 					d.inner = scheduler.schedule(after: date) {
-						state.modify { state in
-							defer {
-								if state.pendingValue != nil {
-									state.pendingValue = nil
-									state.previousDate = date
-								}
-							}
-							if let pendingValue = state.pendingValue {
-								action(.value(pendingValue))
-							}
+						if let pendingValue = state.modify({ $0.retrieveValue(date: date) }) {
+							action(.value(pendingValue))
 						}
 					}
-
-				case .completed, .failed, .interrupted:
+					
+				case .completed:
 					d.inner = scheduler.schedule {
 						let pending: (value: Value, previousDate: Date)? = state.modify { state in
 							defer {
@@ -879,14 +861,19 @@ extension Signal.Event {
 							guard let pendingValue = state.pendingValue, let previousDate = state.previousDate else { return nil }
 							return (pendingValue, previousDate)
 						}
-						if !discardsWhenTerminated, let (pendingValue, pendingDate) = pending {
+						if !discardsWhenCompleted, let (pendingValue, pendingDate) = pending {
 							scheduler.schedule(after: pendingDate.addingTimeInterval(interval)) {
 								action(.value(pendingValue))
-								action(event)
+								action(.completed)
 							}
 						} else {
-							action(event)
+							action(.completed)
 						}
+					}
+
+				case .failed, .interrupted:
+					d.inner = scheduler.schedule {
+						action(event)
 					}
 				}
 			}
@@ -897,6 +884,16 @@ extension Signal.Event {
 private struct ThrottleState<Value> {
 	var previousDate: Date?
 	var pendingValue: Value?
+	
+	mutating func retrieveValue(date: Date) -> Value? {
+		defer {
+			if pendingValue != nil {
+				pendingValue = nil
+				previousDate = date
+			}
+		}
+		return pendingValue
+	}
 }
 
 extension Signal.Event where Error == NoError {
--- a/Sources/Event.swift
+++ b/Sources/Event.swift
@@ -853,11 +853,7 @@ extension Signal.Event {
 				case .completed:
 					d.inner = scheduler.schedule {
 						let pending: (value: Value, previousDate: Date)? = state.modify { state in
-							defer {
-								if state.pendingValue != nil {
-									state.pendingValue = nil
-								}
-							}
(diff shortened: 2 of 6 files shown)


Qata:
Solves #287

Qata:
@ikesyo is this API breaking? The default parameter should just make it additive.

andersio:
@Qata It is source compatible but binary incompatible.

ikesyo:
If one uses the method with method reference, the addition of a default arguments would break that usages (theoretically).

andersio:
It is fine. We are going to do a major release anyway since #610 strictly speaking is API breaking.

Sources/Event.swift
```diff
@@ -834,28 +834,59 @@ extension Signal.Event {
...
+	internal static func debounce(_ interval: TimeInterval, on scheduler: DateScheduler, discardsWhenTerminated: Bool) -> Transformation<Value, Error> {
 		precondition(interval >= 0)
-
+		
+		let state: Atomic<ThrottleState<Value>> = Atomic(ThrottleState(previousDate: scheduler.currentDate, pendingValue: nil))
+		
 		return { action, lifetime in
 			let d = SerialDisposable()
-
+
```
andersio: [Nit] Whitespace

Sources/Event.swift
```diff
@@ -834,28 +834,59 @@ extension Signal.Event {
...
 		return { action, lifetime in
 			let d = SerialDisposable()
-
+			
 			lifetime.observeEnded {
 				d.dispose()
 				scheduler.schedule { action(.interrupted) }
 			}
-
+
```
andersio: [Nit] Whitespace

Sources/Signal.swift
```diff
@@ -1491,20 +1491,25 @@ extension Signal {
...
+	///			If `false`, that value will be delivered at the next debounce
+	///			interval.
 	///
 	/// - precondition: `interval` must be non-negative number.
 	///
 	/// - parameters:
 	///   - interval: A number of seconds to wait before sending a value.
 	///   - scheduler: A scheduler to send values on.
+	///	  - discardsWhenTerminated: A boolean to indicate if the latest value
+	///                             should be discarded on termination.
```
andersio: [Nit] Indentation.

Sources/SignalProducer.swift
```diff
@@ -1514,17 +1514,22 @@ extension SignalProducer {
...
+	///         If `true`, that value will be discarded and the returned producer
+	///         will terminate immediately.
+	///			If `false`, that value will be delivered at the next debounce
+	///			interval.
 	///
 	/// - parameters:
 	///   - interval: A number of seconds to wait before sending a value.
 	///   - scheduler: A scheduler to send values on.
+	///	  - discardsWhenTerminated: A boolean to indicate if the latest value
+	///                             should be discarded on termination.
```
andersio: [Nit] Indentation.

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
```diff
@@ -1341,10 +1341,18 @@ class SignalProducerLiftingSpec: QuickSpec {
 			it("should interrupt ASAP and discard outstanding events") {
 				testAsyncASAPInterruption(op: "debounce") { $0.debounce(10.0, on: $1) }
```
andersio: Could we specify `discardsWhenTerminated: true` here explicitly?

Tests/ReactiveSwiftTests/SignalProducerLiftingSpec.swift
```diff
@@ -1341,10 +1341,18 @@ class SignalProducerLiftingSpec: QuickSpec {
...
 			}
+			
+			it("should interrupt ASAP and discard outstanding events") {
+				testAsyncASAPInterruption(op: "debounce") { $0.debounce(10.0, on: $1, discardsWhenTerminated: false) }
+			}
 
 			it("should interrupt on the given scheduler") {
 				testAsyncInterruptionScheduler(op: "debounce") { $0.debounce(10.0, on: $1) }
```
andersio: Ditto.

Tests/ReactiveSwiftTests/SignalSpec.swift
```diff
@@ -1953,6 +1953,114 @@ class SignalSpec: QuickSpec {
 				expect(completed) == true
 			}
 		}
+		
+		describe("debounce without discarding when terminated") {
```
andersio: The test case group above should be renamed likewise, and perhaps:
> discarding the latest value when terminated

Tests/ReactiveSwiftTests/SignalSpec.swift
```diff
@@ -1953,6 +1953,114 @@ class SignalSpec: QuickSpec {
...
+				scheduler.advance()
+				expect(values) == []
+				expect(completed) == false
+				
+				scheduler.run()
+				expect(values) == [1]
+				expect(completed) == true
+			}
+			
+			it("should schedule completion immediately") {
```
andersio: > if there is no pending value

Qata:
@andersio Comments addressed

Sources/Event.swift
```diff
@@ -848,14 +850,43 @@ extension Signal.Event {
...
-						action(.value(value))
+						state.modify { state in
+							defer {
+								if state.pendingValue != nil {
+									state.pendingValue = nil
+									state.previousDate = date
+								}
+							}
+							if let pendingValue = state.pendingValue {
+								action(.value(pendingValue))
```
mdiep: `action` should be called outside of `modify`

Sources/Event.swift
```diff
@@ -848,14 +850,43 @@ extension Signal.Event {
...
+					}
 					let date = scheduler.currentDate.addingTimeInterval(interval)
 					d.inner = scheduler.schedule(after: date) {
-						action(.value(value))
+						state.modify { state in
+							defer {
+								if state.pendingValue != nil {
+									state.pendingValue = nil
+									state.previousDate = date
+								}
```
mdiep: I think this logic (repeated in `throttle`) should be extracted into a method on `ThrottleState`.

Sources/Event.swift
```diff
@@ -848,14 +850,43 @@ extension Signal.Event {
...
+							}
+							if let pendingValue = state.pendingValue {
+								action(.value(pendingValue))
+							}
+						}
 					}
 
 				case .completed, .failed, .interrupted:
```
mdiep: Interruptions are generally immediate. I think introducing a delay here might be a problem. (I also think it's generally unexpected that interrupting would send another value. If you've canceled the work, then you're not interested in any other results.)

@andersio?

andersio: I agree, and `failed` should follow suit since by contract [`failures` behave like exceptions and propagate immediately](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Documentation/APIContracts.md#failures-behave-like-exceptions-and-propagate-immediately).

andersio: > If you've canceled the work, then you're not interested in any other results.

If the stream has already failed, I don't think emitting the pending value would make any difference.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # master
 *Please add new entries at the top.*
 
+1. New `debounce` parameter: `discardsWhenTerminated` (#287, kudos to @Qata)
```
andersio: > `debounce` now offers an opt-in behaviour to preserve the pending value when the signal or producer completes. You may enable it by specifying `discardsWhenCompleted` as `false`.

Sources/Event.swift
```diff
@@ -848,12 +840,38 @@ extension Signal.Event {
...
+						let pending: (value: Value, previousDate: Date)? = state.modify { state in
+							defer {
+								if state.pendingValue != nil {
+									state.pendingValue = nil
+								}
+							}
+							guard let pendingValue = state.pendingValue, let previousDate = state.previousDate else { return nil }
+							return (pendingValue, previousDate)
+						}
+						if !discardsWhenCompleted, let (pendingValue, pendingDate) = pending {
```
andersio: `/pendingDate/previousDate/`

Sources/Signal.swift
```diff
@@ -1491,20 +1491,25 @@ extension Signal {
...
+	///			If `false`, that value will be delivered at the next debounce
+	///			interval.
 	///
 	/// - precondition: `interval` must be non-negative number.
 	///
 	/// - parameters:
 	///   - interval: A number of seconds to wait before sending a value.
 	///   - scheduler: A scheduler to send values on.
+	///   - discardsWhenCompleted: A boolean to indicate if the latest value
+	///                             should be discarded on termination.
```
andersio: /termination/completion/

Sources/Signal.swift
```diff
@@ -1491,20 +1491,25 @@ extension Signal {
...
 	///         the latest value is the one that will be passed on.
 	///
-	/// - note: If the input signal terminates while a value is being debounced, 
-	///         that value will be discarded and the returned signal will 
-	///         terminate immediately.
+	/// - note: If `self` terminates while a value is being debounced,
+	///         the behaviour will be determined by `discardsWhenCompleted`.
+	///         If `true`, that value will be discarded and the returned producer
+	///         will terminate immediately.
+	///			If `false`, that value will be delivered at the next debounce
```
andersio: Indentation.

Sources/SignalProducer.swift
```diff
@@ -1514,17 +1514,22 @@ extension SignalProducer {
 	///         the latest value is the one that will be passed on.
 	///
 	/// - note: If `self` terminates while a value is being debounced,
-	///         that value will be discarded and the returned producer will
-	///         terminate immediately.
+	///         the behaviour will be determined by `discardsWhenCompleted`.
+	///         If `true`, that value will be discarded and the returned producer
+	///         will terminate immediately.
+	///			If `false`, that value will be delivered at the next debounce
```
andersio: Use spaces instead of tabs for comments.

Sources/Signal.swift
andersio: Use spaces instead of tabs for comments.

Sources/Signal.swift
andersio: Use spaces instead of tabs for comments.

Qata:
@andersio Comments have been addressed.

Sources/Event.swift
```diff
@@ -816,26 +816,18 @@ extension Signal.Event {
...
+					if let pendingValue = state.modify({ $0.retrieveValue(date: scheduleDate) }) {
 						action(.value(pendingValue))
 					}
 				}
 			}
 		}
 	}
 
-	internal static func debounce(_ interval: TimeInterval, on scheduler: DateScheduler) -> Transformation<Value, Error> {
+	internal static func debounce(_ interval: TimeInterval, on scheduler: DateScheduler, discardsWhenCompleted: Bool) -> Transformation<Value, Error> {
```
mdiep: I think this should probably be `discardWhenCompleted` (no _s_).

Sources/Event.swift
```diff
@@ -848,12 +840,38 @@ extension Signal.Event {
...
+						if let pendingValue = state.modify({ $0.retrieveValue(date: date) }) {
+							action(.value(pendingValue))
+						}
+					}
+					
+				case .completed:
+					d.inner = scheduler.schedule {
+						let pending: (value: Value, previousDate: Date)? = state.modify { state in
+							defer {
+								if state.pendingValue != nil {
```
mdiep: This `if` isn't needed. We can unconditionally assign `nil`.

Sources/Event.swift
Qata: I feel like that gives the implication that it will always discard *something*. Might be worth bikeshedding.

mdiep: I don't see how the _s_ changes that.

The _s_ seems inconsistent the base operator name and other operators. Since it's `debounce`, not `debounces`, I think `discardWithCompleted` is more appropriate.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # master
 *Please add new entries at the top.*
 
+1. `debounce` now offers an opt-in behaviour to preserve the pending value when the signal or producer completes. You may enable it by specifying `discardsWhenCompleted` as false (#287, kudos to @Qata)
```
mdiep: This _s_ needs to be removed too. 🙈

mdiep:
Thanks for your work on this! 🎉

PR #607, opened 2018-02-15
Bag optimization

ra1028:
This is continuation of https://github.com/ReactiveCocoa/ReactiveSwift/pull/602

### Use `IndexingIterator` implemented in `ContiguousArray`
This simplifies the implementation and gave the performance improves 2~3x.
Benchmark:
```swift
func testIteration() {
    let count = 10_000_000
    var bag = Bag<Int>()
    
    for i in 0..<count {
        bag.insert(i)
    }
    
    XCTAssertEqual(bag.count, count)
    
    measure {
        var trash = [Int]()
        trash.reserveCapacity(count)
        
        for i in bag {
            trash.append(i)
        }
        
        XCTAssertEqual(bag.count, count)
        XCTAssertEqual(trash.count, count)
    }
}
```
Result:
performance

### Use overflow addition operator `&+`
Higher performance can be expected by dropping the overflow checking.
There is no much impact on performance, but this would be better, because `Token.value` doesn't assume overflow.
Benchmark:
```swift
func testRemoving() {
    let count = 10_000
    
    measure {
        var bag = Bag<Int>()
        var tokens = ContiguousArray<Bag<Int>.Token>()
        tokens.reserveCapacity(count)
        
        for i in 0..<count {
            let token = bag.insert(i)
            tokens.append(token)
        }
        
        XCTAssertEqual(bag.count, count)
        
        
        for token in tokens {
            bag.remove(using: token)
        }
        
        XCTAssertEqual(bag.count, 0)
    }
}
```
Result:
performance2

#### Checklist
~- [ ] Updated CHANGELOG.md.~

### DIFF
--- a/Sources/Bag.swift
+++ b/Sources/Bag.swift
@@ -32,7 +32,7 @@ public struct Bag<Element> {
 
 		// Practically speaking, this would overflow only if we have 101% uptime and we
 		// manage to call `insert(_:)` every 1 ns for 500+ years non-stop.
-		nextToken = Token(value: token.value + 1)
+		nextToken = Token(value: token.value &+ 1)
 
 		elements.append(value)
 		tokens.append(token.value)
--- a/Sources/Bag.swift
+++ b/Sources/Bag.swift
@@ -70,31 +70,7 @@ extension Bag: RandomAccessCollection {
 		return elements[index]
 	}
 
-	public func makeIterator() -> Iterator {
-		return Iterator(elements)
-	}
-
-	/// An iterator of `Bag`.
-	public struct Iterator: IteratorProtocol {
-		private let base: ContiguousArray<Element>
-		private var nextIndex: Int
-		private let endIndex: Int
-
-		fileprivate init(_ base: ContiguousArray<Element>) {
-			self.base = base
-			nextIndex = base.startIndex
-			endIndex = base.endIndex
-		}
-
-		public mutating func next() -> Element? {
-			let currentIndex = nextIndex
-
-			if currentIndex < endIndex {
-				nextIndex = currentIndex + 1
-				return base[currentIndex]
-			}
-
-			return nil
-		}
+	public func makeIterator() -> IndexingIterator<ContiguousArray<Element>> {
+		return elements.makeIterator()
 	}
 }
--- a/Sources/Bag.swift
+++ b/Sources/Bag.swift
@@ -70,7 +70,20 @@ extension Bag: RandomAccessCollection {
 		return elements[index]
 	}
 
-	public func makeIterator() -> IndexingIterator<ContiguousArray<Element>> {
-		return elements.makeIterator()
+	public func makeIterator() -> Iterator {
+		return Iterator(elements.makeIterator())
+	}
+
+	/// An iterator of `Bag`.
+	public struct Iterator: IteratorProtocol {
+		private var base: ContiguousArray<Element>.Iterator
+
+		fileprivate init(_ base: ContiguousArray<Element>.Iterator) {
+			self.base = base
+		}
+
+		public mutating func next() -> Element? {
+			return base.next()
+		}
 	}
 }


andersio:
The performance gain is likely due to the special status of the Standard Library regarding cross-module optimisation. Only if you write the test as part of the ReactiveSwift module & the gain is maintained, then it can be concluded as an improvement.

It also leaks an implementation detail of `Bag`.

ra1028:
Thanks @andersio 
I defined the following method in ReactiveSwift/`Bag.swift` and benchmarked again.
```swift
public extension Bag where Element == Int {
    mutating func iterationTest() -> [Int] {
        var trash = [Int]()
        trash.reserveCapacity(count)

        for i in self {
            trash.append(i)
        }

        return trash
    }
}
```

Benchmark:
```swift
func testIteration() {
	let count = 10_000_000
	var bag = Bag<Int>()
	for i in 0..<count {
		bag.insert(i)
	}
	measure {
		let trash = bag.iterationTest()
		XCTAssertEqual(trash.count, count)
	}
}

func testIteration2() {
	let count = 10_000_000
	var bag = Bag<Int>()
	for i in 0..<count {
		bag.insert(i)
	}
	measure {
		var trash = [Int]()
		trash.reserveCapacity(count)
		
		for i in bag {
			trash.append(i)
		}
		
		XCTAssertEqual(trash.count, count)
	}
}
```
Result(-Owholemodule):
performance1

Result(-Onone):
performance2

The reason why it got worse performance with `-Onone`, maybe related to optimizations of inlining, specialization and witness table in SIL.

andersio:
I did a quick test, and apparently a wrapper struct around `IndexingIterator<ContiguousArray<Element>>` is as performant as using it directly with `-O -whole-module-optimization`.

ra1028:
got it.
thanks for reviews @andersio

andersio:
You could still replace the internal of `Bag.Iterator` with `IndexingIterator<ContiguousArray<Element>>` though.

ra1028:
oops I misunderstood.
will do.

ra1028:
@andersio 
Did you intend `ContiguousArray<Element>.Iterator`'s wrapper struct like this?
```swift
public struct Iterator: IteratorProtocol {
    private var base: ContiguousArray<Element>.Iterator

    fileprivate init(_ base: ContiguousArray<Element>.Iterator) {
        self.base = base
    }

    public mutating func next() -> Element? {
        return base.next()
    }
}
```

I benchmarked the same for it.
Result:
screen shot 2018-02-24 at 4 01 19
It's about the same performance with using `ContiguousArray<Element>.Iterator` in ReactiveSwift module, but it got worse in cross-module.

In addition, I also benchmarked the following `Iterator`.
```swift
/// An iterator of `Bag`.
public struct Iterator: IteratorProtocol {
    @_versioned
    internal private(set) var base: ContiguousArray<Element>.Iterator

    fileprivate init(_ base: ContiguousArray<Element>.Iterator) {
        self.base = base
    }

    @inline(__always)
    public mutating func next() -> Element? {
        return base.next()
    }
}
```
Result:
screen shot 2018-02-24 at 3 53 36
It performant as same as using `ContiguousArray<Element>.Iterator` directly.
Apparently the difference in performance seems to be due to inlining.
It's not leaking an implementation detail, and better performance, but I doesn't recommended to use `@_versioned internal`.

What do you think about these results?

andersio:
> Did you intend ContiguousArray<Element>.Iterator's wrapper struct like this?

Yes

> It's about the same performance with using ContiguousArray<Element>.Iterator directly in ReactiveSwift module, but it got worse in cross-module.

Performance of external use is not a primary concern of any primitive we offer here, because it is a well acknowledged issue with a solution materialising in Swift 5.

> In addition, I also benchmarked the following Iterator.

We are not committed to use any compiler private attribute at this point. These attributes _will_ show up in Swift 5 as part of the official language spec AFAIK, so until then...

ra1028:
OK, I'll implement wrapper struct around IndexingIterator 👍

PR #608, opened 2018-02-19
[Gardening] Disposable

ra1028:
#### Checklist
~- [ ] Updated CHANGELOG.md.~

### DIFF
--- a/Sources/Disposable.swift
+++ b/Sources/Disposable.swift
@@ -145,7 +145,7 @@ public final class CompositeDisposable: Disposable {
 		}
 
 		self.disposables = Atomic(bag)
-		self.state = UnsafeAtomicState(DisposableState.active)
+		self.state = UnsafeAtomicState(.active)
 	}
 
 	/// Initialize a `CompositeDisposable` containing the given sequence of
--- a/Sources/Disposable.swift
+++ b/Sources/Disposable.swift
@@ -169,9 +169,9 @@ public final class CompositeDisposable: Disposable {
 	}
 
 	public func dispose() {
-		if state.tryDispose(), let ds = disposables.swap(nil) {
-			for d in ds {
-				d.dispose()
+		if state.tryDispose(), let disposables = disposables.swap(nil) {
+			for disposable in disposables {
+				disposable.dispose()
 			}
 		}
 	}
@@ -186,7 +186,7 @@ public final class CompositeDisposable: Disposable {
 	///            `disposable` is `nil`.
 	@discardableResult
 	public func add(_ disposable: Disposable?) -> Disposable? {
-		guard let d = disposable, !d.isDisposed, !isDisposed else {
+		guard let _disposable = disposable, !_disposable.isDisposed, !isDisposed else {
 			disposable?.dispose()
 			return nil
 		}
@@ -194,7 +194,7 @@ public final class CompositeDisposable: Disposable {
 		return disposables.modify { disposables in
 			guard disposables != nil else { return nil }
 
-			let token = disposables!.insert(d)
+			let token = disposables!.insert(_disposable)
 			return AnyDisposable { [weak self] in
 				self?.disposables.modify {
 					$0?.remove(using: token)
@@ -358,10 +358,10 @@ public final class SerialDisposable: Disposable {
 			return _inner.value
 		}
 
-		set(d) {
-			_inner.swap(d)?.dispose()
-			if let d = d, isDisposed {
-				d.dispose()
+		set(disposable) {
+			_inner.swap(disposable)?.dispose()
+			if let disposable = disposable, isDisposed {
+				disposable.dispose()
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


Sources/Disposable.swift
```diff
@@ -33,6 +33,7 @@ extension UnsafeAtomicState where State == DisposableState {
 	///
 	/// - returns: `true` if the transition succeeds. `false` otherwise.
 	@inline(__always)
+	@discardableResult
```
mdiep: I don't think this result should be discardable.  It's important that we check the result most of the time.

Sources/Disposable.swift
```diff
@@ -83,10 +84,9 @@ public final class AnyDisposable: Disposable {
 		}
 
 		func dispose() {
-			if state.tryDispose() {
-				action?()
-				action = nil
-			}
+			guard state.tryDispose() else { return }
+			action?()
+			action = nil
```
mdiep: I prefer `if` over `guard` here. I think it communicates the intent more clearly.

Sources/Disposable.swift
```diff
@@ -163,12 +159,10 @@ public final class CompositeDisposable: Disposable {
...
-		if state.tryDispose() {
-			if let ds = disposables.swap(nil) {
-				for d in ds {
-					d.dispose()
-				}
-			}
+		guard state.tryDispose(), let disposables = disposables.swap(nil) else { return }
+
+		for disposable in disposables {
+			disposable.dispose()
```
mdiep: Same here: I think `if` is clearer than `guard`.

Sources/Disposable.swift
```diff
@@ -182,15 +176,14 @@ public final class CompositeDisposable: Disposable {
 	///            `disposable` is `nil`.
 	@discardableResult
 	public func add(_ disposable: Disposable?) -> Disposable? {
-		guard let d = disposable, !d.isDisposed, !isDisposed else {
+		guard let _disposable = disposable, !_disposable.isDisposed, !isDisposed else {
```
mdiep: I think `d` is fine here; we don't need a longer name.

Sources/Disposable.swift
```diff
@@ -182,15 +176,14 @@ public final class CompositeDisposable: Disposable {
...
 			disposable?.dispose()
 			return nil
 		}
 
 		return disposables.modify { disposables in
-			guard disposables != nil else { return nil }
+			guard let token = disposables?.insert(_disposable) else { return nil }
 
-			let token = disposables!.insert(d)
```
mdiep: Nice. 👍

Sources/Disposable.swift
```diff
@@ -373,9 +367,8 @@ public final class SerialDisposable: Disposable {
 	}
 
 	public func dispose() {
-		if state.tryDispose() {
-			_inner.swap(nil)?.dispose()
-		}
+		guard state.tryDispose() else { return }
+		_inner.swap(nil)?.dispose()
```
mdiep: Same here. I prefer the `if` version.

mdiep:
Thanks for the PR! I always appreciate efforts to clean up the codebase.

I don't think all of the changes are better, but there are some good things in here. 👍

Sources/Disposable.swift
ra1028: `tryDispose()` is a method for transitioning the state of `Disposable` and thinks that the result is a supplementary thing.
How is it?

Sources/Disposable.swift
ra1028: Hmm, disagreed.
By returning asap, the intention of do nothing if already disposed would be clarified.

Sources/Disposable.swift
ra1028: I prefer `guard`.
However, `swap` may be better to use` if`.

Sources/Disposable.swift
ra1028: I don't prefer to give meaningless variable names like `d`, but if the RAS naming conventions are are prefer `d`, I'll follow it.

Sources/Disposable.swift
ra1028: > By returning asap, the intention of do nothing if already disposed would be clarified.

This also.

Sources/Disposable.swift
andersio: The returned Boolean indicates whether the caller is the one uniquely transitioned the state, which is critical to most use cases.

Sources/Disposable.swift
mdiep: Thinking about it a bit more, I dislike `guard` here because `state.tryDispose()` is effectful. `guard` strongly suggests a test that's free of effects IMO. I still think it should stay as `if` here and below.

ra1028: Got it 👍

mdiep:
Thanks for this! Clean up is always appreciated!

PR #609, opened 2018-02-19
Create Bag with initial elements

ra1028:
Hi✋ 
I made it possible to create Bag with initial elements, and CompositeDisposable also improved accordingly.
Probably conflicts with #608, but I fix after either is merged.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/Sources/Bag.swift
+++ b/Sources/Bag.swift
@@ -14,12 +14,22 @@ public struct Bag<Element> {
 		fileprivate let value: UInt64
 	}
 
-	fileprivate var elements: ContiguousArray<Element> = []
-	fileprivate var tokens: ContiguousArray<UInt64> = []
+	fileprivate var elements: ContiguousArray<Element>
+	fileprivate var tokens: ContiguousArray<UInt64>
 
-	private var nextToken = Token(value: 0)
+	private var nextToken: Token
 
-	public init() {}
+	public init() {
+		elements = ContiguousArray()
+		tokens = ContiguousArray()
+		nextToken = Token(value: 0)
+	}
+
+	public init<S: Sequence>(_ elements: S) where S.Iterator.Element == Element {
+		nextToken = Token(value: UInt64(elements.underestimatedCount))
+		tokens = ContiguousArray(0..<nextToken.value)
+		self.elements = ContiguousArray(elements)
+	}
 
 	/// Insert the given value into `self`, and return a token that can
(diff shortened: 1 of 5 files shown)


Sources/Bag.swift
```diff
@@ -14,12 +14,22 @@ public struct Bag<Element> {
...
+	public init() {
+		elements = ContiguousArray()
+		tokens = ContiguousArray()
+		nextToken = Token(value: 0)
+	}
+
+	public init<S: Sequence>(_ elements: S) where S.Iterator.Element == Element {
+		nextToken = Token(value: UInt64(elements.underestimatedCount))
+		tokens = ContiguousArray(0..<nextToken.value)
+		self.elements = ContiguousArray(elements)
```
NachoSoto: Be consistent here: either self for all or none.

Sources/Disposable.swift
```diff
@@ -140,12 +140,7 @@ public final class CompositeDisposable: Disposable {
 	public init<S: Sequence>(_ disposables: S)
 		where S.Iterator.Element == Disposable
 	{
-		var bag: Bag<Disposable> = Bag()
-
-		for disposable in disposables {
-			bag.insert(disposable)
-		}
-
+		let bag = Bag(disposables)
```
NachoSoto: 👍

Sources/Bag.swift
ra1028: will do.
`elements` is used as a parameter name, so I'll add` self` to all.

Sources/Bag.swift
```diff
@@ -14,12 +14,22 @@ public struct Bag<Element> {
...
 
-	public init() {}
+	public init() {
+		elements = ContiguousArray()
+		tokens = ContiguousArray()
+		nextToken = Token(value: 0)
+	}
+
+	public init<S: Sequence>(_ elements: S) where S.Iterator.Element == Element {
+		self.nextToken = Token(value: UInt64(elements.underestimatedCount))
```
andersio: `underestimatedCount` is not guaranteed to equal the exact number of elements. We should derive this from the `ContiguousArray` constructed from `elements`.

ra1028: Ah, true.

ra1028: Amended:+1:

mdiep:
Thanks for the PR!

Issue #598, opened 2018-02-02
Ambiguous reference to 'observeValues'


danielaRiesgo:
To any signal with an error that is not NoError, every time I want to use `.observeValues` I get a compiler error saying:
    Ambiguous reference to member 'observeValues'
    Found this candidate ReactiveSwift.Signal<Value, NoError>
    Found this candidate ReactiveSwift.Signal<Never, NoError>

ReactiveSwift version: 3.1.0
ReactiveCocoa version: 7.1.0

It can be reproduce with the following code:
```
let signal = UIButton().reactive.controlEvents(.touchUpInside).mapError { _ in NSError(domain: "My domain", code: 3, userInfo: .none) }

signal.observeValues { _ in print("error") } 
```

jjoelson:
In order to discourage users from ignoring errors, [`observeValues` is only defined for `NoError` Signals](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/Signal.swift#L515). You might want to take a look at `observeResult` which has the same signature as `observeValues` except it's closure takes a `Result<Value, Error>` that you can switch over in order to handle errors.

NachoSoto:
Yeah this is by design. In fact I think the problem is that Swift is providing a bogus diagnostic, [as usual](https://bugs.swift.org/browse/SR-5198).

mdiep:
Could we provide an unavailable version that explains it while leaving the `NoError` version in tact?

danielaRiesgo:
Yeah, I know about `observeResults` but the error I think is not good. I think in the previous versions you got a warning saying you should better use `observeResult`, but know it just tells you that there is ambiguity for signals that don't even match yours. It's strange.

RuiAAPeres:
Hello. 👋  Thanks for opening this issue. Due to inactivity, we will soft close the issue. If you feel that it should remain open, please let us know. 😄

PR #618, opened 2018-03-03
Explicitly warning that cannot observe with ignoring the errors.

ra1028:
Currently, compiler warns `Ambiguous reference to observeValues` when use `observeValues` of `Signal` that with an error type which is not `NoError`.
This is a confusing API.
I think it should better warn to use observeResult.
relevant issue: https://github.com/ReactiveCocoa/ReactiveSwift/issues/598

screen shot 2018-03-03 at 11 40 10
screen shot 2018-03-03 at 11 39 54

#### Checklist
~- [ ] Updated CHANGELOG.md.~

### DIFF
--- a/Sources/UninhabitedTypeGuards.swift
+++ b/Sources/UninhabitedTypeGuards.swift
@@ -118,6 +118,9 @@ private func observingUninhabitedTypeError() -> Never {
 
 /*
 func test() {
+	SignalProducer<Any, AnyError>.never.startWithValues { _ in }
+	Signal<Any, AnyError>.never.observeValues { _ in }
+
 	SignalProducer<Never, AnyError>.never.startWithResult { _ in }
 	SignalProducer<Never, NoError>.never.startWithResult { _ in }
 	SignalProducer<Any, NoError>.never.startWithFailed { _ in }
--- a/Sources/UninhabitedTypeGuards.swift
+++ b/Sources/UninhabitedTypeGuards.swift
@@ -1,6 +1,18 @@
 import Result
 
 // Observation
+extension SignalProducer {
+	@available(*, unavailable, message:"Use `startWithResult` instead - cannot ignore errors that may occur")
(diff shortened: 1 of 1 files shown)


Sources/UninhabitedTypeGuards.swift
```diff
@@ -1,6 +1,18 @@
...
+extension SignalProducer {
+	@available(*, unavailable, message:"Use `startWithResult` instead - cannot ignore errors that may occur")
+	@discardableResult
+	public func startWithValues(_ action: @escaping (Value) -> Void) -> Disposable { observingUninhabitedTypeError() }
+}
+
+extension Signal {
+	@available(*, unavailable, message:"Use `observeResult` instead - cannot ignore errors that may occur")
+	@discardableResult
+	public func observeValues(_ action: @escaping (Value) -> Void) -> Disposable { observingUninhabitedTypeError() }
```
andersio: `-> Disposable?`.

Sources/UninhabitedTypeGuards.swift
```diff
@@ -1,6 +1,18 @@
 import Result
 
 // Observation
+extension SignalProducer {
+	@available(*, unavailable, message:"Use `startWithResult` instead - cannot ignore errors that may occur")
+	@discardableResult
+	public func startWithValues(_ action: @escaping (Value) -> Void) -> Disposable { observingUninhabitedTypeError() }
```
andersio: > Transform the error beforehand, or use `startWithResult` instead

Sources/UninhabitedTypeGuards.swift
```diff
@@ -1,6 +1,18 @@
...
+}
+
+extension Signal {
+	@available(*, unavailable, message:"Use `observeResult` instead - cannot ignore errors that may occur")
```
andersio: Ditto

Issue #614, opened 2018-03-02
A throttle-like function that accumulates values


Qata:
We currently have no way to see all values that arrived during a given time period.

mdiep:
Something like that was proposed in https://github.com/ReactiveCocoa/ReactiveSwift/pull/547, but was never finished.

Qata:
Ah that looks quite different though. You could actually implement that function with this one and flatMap I think.

PR #619, opened 2018-03-04
Add a recurring collect function and associated unit tests

Qata:
#### Checklist
- [X] Updated CHANGELOG.md.

This PR introduces a new function that collects values from `self` and delivers them to the subscriber every `interval`.
Solves issue #614.


Sources/Signal.swift
```diff
@@ -1506,6 +1506,25 @@ extension Signal {
...
 	}
+
+	/// Forward the latest values on `scheduler` every `interval`.
+	///
+	/// - seealso: `throttle`
+	/// - seealso: `debounce`
+	///
+	/// - note: If the input signal terminates, the returned signal will
+	///         terminate immediately without forwarding the values
+	///	        currently being accumulated.
```
mdiep: In the spirit of #605, maybe there should be an option to forward the values first?

Sources/Signal.swift
```diff
@@ -1506,6 +1506,25 @@ extension Signal {
...
+	///
+	/// - parameters:
+	///   - interval: A repetition interval.
+	///   - scheduler: A scheduler to send values on.
+	///
+	/// - returns: A signal that sends all values that are sent from `self` at
+	///            `interval` seconds apart.
+	public func chunk(_ interval: DispatchTimeInterval, on scheduler: DateScheduler, ignoreEmptyChunks: Bool = false) -> Signal<[Value], Error> {
```
mdiep: I'm very hesitant to introduce complete new operator names. I think our grouping of operators by base name has been very successful and would like for us to continue it.

While this is similar to `debounce`/`throttle`, I'm not sure it's really throttling them. I don't think a throttle is always regular.

I think this could fit under `collect` though.

```swift
func collect(every interval: DispatchTimeInterval, on scheduler: DataScheduler) -> Signal<[Value], Error> {
```

`collect` serves as the base name for operators that go from `A` to `[A]`.

Thoughts @ReactiveCocoa/reactiveswift?

Qata: Oh yep I forgot to say that bikeshedding the name is very welcome, I have no attachments to the one I chose.

Sources/Signal.swift
Qata: Good point!

Qata: After looking into this I can't find a way to make it work consistently, given:
When you complete the Signal when you'd expect is for it to still delay up to the supplied interval, and only then deliver the values followed immediately by a completed event. 
In the event of `interrupted` or `failed`, you'd need to forward those events immediately, so the values would also need to be delivered immediately. 
This would probably create tricky behaviour for a lot of people consuming this function, since you'd expect values to be delivered exactly `interval` apart (or a multiple of `interval` if you're ignoring when empty).

So should it only not discard in the case of `.completed`?

Sources/Event.swift
```diff
@@ -861,6 +861,42 @@ extension Signal.Event {
...
+			d.inner = scheduler.schedule(after: scheduler.currentDate.addingTimeInterval(interval), interval: interval, leeway: interval * 0.1, action: {
+				let currentValues: [Value]? = values.modify { values in
+					guard !(values.isEmpty && ignoreWhenEmpty) else { return nil }
+					defer { values = [] }
+					return values
+				}
+				if let values = currentValues {
+					action(.value(values))
+				}
+			})
```
mdiep: This should use trailing closure syntax.

Qata:
@mdiep Issues addressed 👍

Sources/Event.swift
```diff
@@ -861,6 +861,48 @@ extension Signal.Event {
...
+			return { event in
+				switch event {
+				case let .value(value):
+					values.modify { $0.append(value) }
+				case let .failed(error):
+					d.inner = scheduler.schedule { action(.failed(error)) }
+				case .completed:
+					let currentValues = values.swap([])
+					if !discardsWhenCompleted, !(currentValues.isEmpty && skipEmpty) {
+						d.inner = scheduler.schedule(after: lastDelivered.addingTimeInterval(interval)) {
```
mdiep: This `schedule` call makes me a little uneasy. It's difficult to reason about how this will interact with the other scheduling call. (e.g. it's possible that the above schedule could run in-between this `values.swap` and this `scheduler.schedule`).

I think we should solve this by removing this `schedule` and sending the complete through the top `schedule`. This would be best done by adding a `CollectEveryState` struct and putting that inside an `Atomic` instead of a `Atomic<[Value]>`.

Something like this:

```swift
state CollectEveryState<Value> {
    var skipEmpty: Bool
    var values: [Value] = []
    var isCompleted: Bool

    var hasValues: Bool {
       return !values.isEmpty || !skipEmpty
    }

    func collect() -> [Value]? {
        if !hasValues {
            return nil
        }
        defer { values.removeAll() }
        return values
    }
}
```

That should clean up the implementation a bit too, which is nice.

Qata: @mdiep Done 👍

Sources/Event.swift
```diff
@@ -861,6 +861,63 @@ extension Signal.Event {
...
+			let values = Atomic<CollectEveryState<Value>>(.init(skipEmpty: skipEmpty))
+			let d = SerialDisposable()
+			
+			d.inner = scheduler.schedule(after: scheduler.currentDate.addingTimeInterval(interval), interval: interval, leeway: interval * 0.1) {
+				let (currentValues, isCompleted) = values.modify { ($0.collect(), $0.isCompleted) }
+				if let currentValues = currentValues {
+					action(.value(currentValues))
+				}
+				if isCompleted {
+					action(.completed)
```
Qata: @mdiep Should I call `d.dispose()` manually here?

mdiep: Yes, I think so. Otherwise the timer would continue firing.

Sources/Event.swift
```diff
@@ -861,6 +861,63 @@ extension Signal.Event {
 			}
 		}
 	}
+	
+	internal static func collect(every interval: DispatchTimeInterval, on scheduler: DateScheduler, skipEmpty: Bool, discardsWhenCompleted: Bool) -> Transformation<[Value], Error> {
+		return { action, lifetime in
+			let values = Atomic<CollectEveryState<Value>>(.init(skipEmpty: skipEmpty))
```
mdiep: Could we also rename `values` to `state` since we have to add the `dispose` call below?

mdiep:
I think this cleaned up rather nicely. 👏 After the minor fixes noted above, we should be good to go.

Sources/Event.swift
```diff
@@ -861,6 +861,63 @@ extension Signal.Event {
 			}
 		}
 	}
+	
+	internal static func collect(every interval: DispatchTimeInterval, on scheduler: DateScheduler, skipEmpty: Bool, discardsWhenCompleted: Bool) -> Transformation<[Value], Error> {
```
mdiep: Oh, let's make this `discardWhenCompleted` (no _s_) since we're doing that in #605 as well.

mdiep:
Unfortunately, the linux build is failing on CI and I don't see why. 😐

mdiep:
This will need a merge now that #605 has been merged.

Qata:
@andersio Could you have a look when you have time

Sources/Event.swift
andersio: `d` is disposed already as part of the lifetime observer, so repeating it here is not necessary.

Qata:
@andersio Done 👍

mdiep:
Thanks @Qata!

PR #623, opened 2018-03-14
Fixes a 'false' Instrument Leaks inside Property.init(unsafeProducer:)

mishagray:
This prevents Instruments from seeing a Leak on any instance of Property.init(unsafeProducer:)

In Swift, Instruments will report a leak, when you still have a stored 'weak pointer' that is pointing to freed memory.   This PR just sets the 'box' pointer to nil as soon as the Relay signal no longer needs it.  So that the 'weak refcount' for the Box value is lowered to zero, and prevents Instruments from reporting it as a leak.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 *Please add new entries at the top.*
 
 1. Instruments will no longer detect a leak during Property.init(unsafeProducer:) (#623, kudos to @mishagray)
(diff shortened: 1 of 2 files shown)


mishagray:
Not sure this was working as intended..

Gonna close this for now until I investigate more.

PR #625, opened 2018-03-30
Add resample(on:) operator

sharplet:
This operator is similar to `sample(on:)`, but also passes through all values from the receiver. This is useful when you want to perform some operation whenever a value changes or upon some manual trigger, like interaction from the user.

- [ ] Updated CHANGELOG.md.

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -1020,6 +1020,21 @@ extension Signal {
 			.map { $0.0 }
 	}
 
+	/// Forward all values from `self`, and resend the latest value whenever
+	/// `sampler` sends a `value` event.
+	///
+	/// - parameters:
+	///   - sampler: A signal that will trigger redelivery of the latest `value`
+	///              event from `self`.
+	///
+	/// - returns: A signal that will send all values from `self`, and will resample
+	///            the latest value whenever `sampler` sends a `value` event. Completes
+	///            once both input signals have completed, or interrupts if either
+	///            input signal is interrupted.
+	public func resample(on sampler: Signal<(), NoError>) -> Signal<Value, Error> {
+		return merge(with: sample(on: sampler))
+	}
+
 	/// Forward the latest value from `samplee` with the value from `self` as a
 	/// tuple, only when `self` sends a `value` event.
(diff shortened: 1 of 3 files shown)


mdiep:
This might work better as a flag on `sample(on:)`?

```swift
func sample(on sampler: Signal<(), NoError>, passValuesThrough: Bool = false) -> Signal<Value, Error>
```

I'm not against including this functionality, but I don't think it warrants a new top-level operator.

sharplet:
So here's an example from my app that using each version:

```swift
searchOrigin.sample(on: refreshed, passThroughValues: true)

searchOrigin.resample(on: refreshed)
```

The first reads to me as two distinct actions: "sample on refreshed" (stops emitting the original values), then "pass through values" (undo that and start passing through values again). The second is more like "give me the values, and resample on refresh".

I don't think it's a very strong argument, but IMO the second version edges ahead of the first in clarity.

mdiep:
> IMO the second version edges ahead of the first in clarity

I agree there. But I think there's balance to be struck between "this is clearest when reading" and "this adds the minimal overhead to the API". We're at two ends of the spectrum right now and it'd be nice to find something in the middle.

If we're going to add a new base operator verb, then I think it should be something that's widely useful or has a number of useful variants. I'm not sure if this fits. 🤔

gfontenot:
I think this PR is pretty stale, but I'd like to voice my support for _something_ like this being included in the API. I just ran into this need in my fairly trivial app and wasn't entirely sure how to make it work. I resorted to `combineLatest(with: other).map { $0.0 }`, but the approach outlined here works closer to what I originally wanted.

mdiep:
I'd prefer to find a name with _sample_ as a prefix, but I'd be okay to merge this if it were updated.

mdiep:
Can you add a test for the `SignalProducer` variant that verifies that the producer is only started once?

gfontenot:
Yeah, I should have noted that I don't have any strong opinions on the spelling, just the utility.

andersio:
Closing due to inactivity. It also seems to me that the use case could be tackled by `withLatest(from:)`. 

```swift
refreshed.withLatest(from: searchOrigin)
```

PR #626, opened 2018-03-30
Add attempt and attemptMap overloads for ErrorConvertible

sharplet:
This allows consumers to use throwable closures with their own custom `ErrorConvertible` types, not just with `AnyError`.

- [ ] Add tests for various type inference scenarios https://github.com/ReactiveCocoa/ReactiveSwift/pull/626#pullrequestreview-108491025
- [ ] Updated CHANGELOG.md.


mdiep:
Definitely 👍 on the idea. Type inference is the only trouble spot, so tests are a good idea like @andersio mentioned.

andersio:
Closing since `antitypical/Result` is no longer a RAS dependency.

Issue #78, opened 2016-11-03
Bring back the Subject?


liscio:
A while ago, @andersio shared a gist with me that implemented a replacement for `RACSubject`. In essence, it just wraps the tuple returned by `Signal<T,E>.pipe()`, but works much nicer in practice.

Here's the gist containing his implementation: https://gist.github.com/andersio/3b4889176d6632bbb637ffd177795815/

Is there a reason that we don't offer this useful construct that we have available in ReactiveObjC? I've never loved the name, personally—calling it a `Pipe` would make far more sense—but since I integrated this code into a project I'm working on, life got a lot easier.

Thoughts?

mdiep:
I'm not sure why this wasn't included in the Swift APIs.

Where are you using it in your project?

liscio:
@mdiep At the moment I use it in places where I'd have used Subject in the past. For instance, I have a "view controller configuration" protocol that vends a `SignalProducer<UIViewController,NSError>` whenever a cell is selected. 

It's not appropriate to store a mutable property containing `SignalProducer<UIViewController,NSError>`, but sending values to a `Subject` makes more sense. When the view controller is loaded/ready, it just sits and watches for that view controller to be pushed.

When this originally surfaced, @andersio and I were speaking about the common "table view cell configuration" pattern where we'd often just store properties containing the model values to be displayed, however we found that this might actually be a better solution—there's no additional storage required as the values are simply piped through to the destination UI components.

I don't use it in _this way_ (yet), but it's on my list to investigate in the future as it seemed like a good idea to avoid building too many layers of storage.

Unit tests would be a lot nicer as well, because we use the `pipe()` pattern a lot there for testing. 

The drawback of using an `Observer` (or `Pipe`) over the `pipe()` tuple is the ability to split the visibility of the two items. For instance, the observer could be made private, and the signal public.

NachoSoto:
I don't think this is a good idea. I've seen this misused more often than not in RX frameworks.
The main reason being that IMO the same type shouldn't be used to represent both ends of the pipe, as it encourages users to expose too wide of a surface instead of only the "read only view".

NachoSoto:
I'm pretty sure that was the reason for not making the pipe method return just one object, but two, because they almost always belong to different owners and they have different purposes.

mdiep:
> The main reason being that IMO the same type shouldn't be used to represent both ends of the pipe, as it encourages users to expose too wide of a surface instead of only the "read only view".

[...]

PR #632, opened 2018-04-10
Introducing Trigger

Marcocanc:
## Description

`Trigger` provides a simple wrapper for Signals that are used to *trigger* something through `Void`.
It conforms to `BindingSource` and `BindingTargetProvider`.

## Motivation

I find myself using `MutableProperty<()>` way too often, especially for my ViewModel inputs.

## Todo

- I'm not entirely sure which verb is the most adequate for the use-case. I've used `fire` for now.
- Proper documentation
- Maybe a Playground example?


brow:
How are you using `MutableProperty<()>`? I am curious because I have none of those in my codebase, in view models or elsewhere.

Marcocanc:
@brow for example when there's a button press. It's used extensively in the [Kickstarter App](https://github.com/kickstarter/ios-oss/) and I started using it since it takes less code than creating a Signal pipe.

brow:
Ah, thanks for the example. In my own projects, I do something like:
```swift
//ViewModel.swift
protocol ViewModelInputs {
    var viewDidAppear: () -> ()
}
// ViewController.swift
self.reactive.viewDidAppear.observeValues(viewModel.viewDidAppear)
```
(I'm assuming your `reactive.viewDidAppear` is convenience method for `reactive.trigger(for: #selector(viewDidAppear))`.)

Is that equivalent? Do you see one as preferable?

brow:
Possibly relevant: https://github.com/ReactiveCocoa/ReactiveSwift/issues/78

Marcocanc:
Sure that works too, but I try to stay fully reactive when binding my ViewModels and ViewControllers, since most inputs will result in an output.

andersio:
[This appears to be a specialised `(Publish)Subject` from which we have moved away since the swing to Swift](https://github.com/ReactiveCocoa/ReactiveSwift/issues/78). I understand the ergonomic aspect of this, and admittedly I had been utilising `Action` in a similar manner. But I personally would stand by the argument of not encouraging this from an API surface perspective, and hence not incorporating this as part of the framework.

If you find creating multiple pipes a bit of a boilerplate, grouping the events as an enum and having a signal of events could potentially be a way out.

PR #633, opened 2018-04-15
feat(bindingTarget): allow optional target in binding infix operator

Ankit-Aggarwal:
#### Checklist
- [ ] Updated CHANGELOG.md.

### Description:
- Allow binding target infix operator (<~) to work on optional target.

### Use Case:
I am having optional IBOutlet due to multiple storyboard file for same controller.
Binding operator was not working with optional target and giving error, so i added support for that using extension on optional.
Now following is possible:
```
        self.disposable += self.selectGenderView?.reactive.isHidden <~ self.viewModel.showGenderSelection.map{!$0}

```

### DIFF
--- a/Sources/UnidirectionalBinding.swift
+++ b/Sources/UnidirectionalBinding.swift
@@ -37,6 +37,57 @@ public protocol BindingTargetProvider {
 	var bindingTarget: BindingTarget<Value> { get }
 }
 
+/// Binds a source to a target, updating the target's value to the latest
+/// value sent by the source, can have target to be optional
+///
+/// - note: The binding will automatically terminate when the target is
+///         deinitialized, or when the source sends a `completed` event.
+///
+/// ````
+/// let property =  MutableProperty<String?>(nil)
+/// let signal = Signal({ /* do some work after some time */ })
+/// property <~ signal
+/// ````
+///
+/// ````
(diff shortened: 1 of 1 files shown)


andersio:
With Swift 4.1, I wonder if having `Optional` conditionally conforming to `BindingTargetProvider` makes better sense.

Ankit-Aggarwal:
@andersio : Agree with you, appended extension. :)

Sources/UnidirectionalBinding.swift
```diff
@@ -37,6 +37,57 @@ public protocol BindingTargetProvider {
...
+///
+/// - returns: A disposable that can be used to terminate binding before the
+///            deinitialization of the target or the source's `completed`
+///            event.
+
+extension Optional where Wrapped: BindingTargetProvider {
+	@discardableResult
+	public static func <~
+		<Source: BindingSource, Target: BindingTargetProvider>
+		(provider1: Target?, source: Source) -> Disposable?
```
mdiep: Can we keep this name as `provider` since it's a public label?

mdiep:
Thanks for the PR! Sorry we've been slow to review it. 🙈 

This needs also some tests and a changelog entry.

mdiep:
We've taken #642 instead. Thanks again for opening a PR!

PR #635, opened 2018-04-25
Add binding operator to Signal.Observer

Marcocanc:
~As the title says, I think it would be nice to make `Observer` conform to `BindingTargetProvider`.
Any thoughts on that?~

This PR adds the binding operator to `Signal.Observer`

#### Checklist
- [x] ~Updated CHANGELOG.md.~

### DIFF
--- a/Sources/UnidirectionalBinding.swift
+++ b/Sources/UnidirectionalBinding.swift
@@ -102,6 +102,37 @@ extension BindingTargetProvider {
 	}
 }
 
+extension Signal.Observer {
+	/// Binds a source to a target, updating the target's value to the latest
+	/// value sent by the source.
+	///
+	/// - note: Only `value` events will be forwarded to the Observer.
+	///         The binding will automatically terminate when the target is
+	///         deinitialized, or when the source sends a `completed` event.
+	///
+	/// - parameters:
+	///   - target: A target to be bond to.
+	///   - source: A source to bind.
+	///
+	/// - returns: A disposable that can be used to terminate binding before the
+	///            deinitialization of the target or the source's `completed`
+	///            event.
+	@discardableResult
+	public static func <~
+		<Source: BindingSource>
+		(observer: Signal<Value, Error>.Observer, source: Source) -> Disposable?
+		where Source.Value == Value
+	{
+		return source.producer.start { [weak observer] in
+			switch $0 {
+			case .value(let val):
+				observer?.send(value: val)
+			default: break
+			}
+		}
+	}
+}
+
 /// A binding target that can be used with the `<~` operator.
 public struct BindingTarget<Value>: BindingTargetProvider {
 	public let lifetime: Lifetime
--- a/Sources/UnidirectionalBinding.swift
+++ b/Sources/UnidirectionalBinding.swift
@@ -123,12 +123,8 @@ extension Signal.Observer {
 		(observer: Signal<Value, Error>.Observer, source: Source) -> Disposable?
 		where Source.Value == Value
 	{
-		return source.producer.start { [weak observer] in
-			switch $0 {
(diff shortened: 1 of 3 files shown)


Sources/Observer.swift
```diff
@@ -124,6 +125,12 @@ extension Signal {
 	}
 }
 
+extension Signal.Observer: BindingTargetProvider {
+	public var bindingTarget: BindingTarget<Value> {
+		return BindingTarget(lifetime: lifetime, action: send)
```
sharplet: I think this is implicitly making a strong reference to `self` here. Might need a weak reference to ensure the binding doesn't keep the observer alive?

andersio: I am not quite convinced, because usually there is an owner whose lifetime the binding target can use.

sharplet: Are you saying another lifetime should be passed in when initialising the observer?

andersio: `Observer`s are generally owned by someone, and IMO it should be that someone who exposes the binding target for the pipe.

We could have a convenience that takes a lifetime to create a binding target though.

mdiep:
Makes sense. 👍 

Can you add some tests and a changelog entry?

Marcocanc:
@mdiep Will do, but there's still an ongoing discussion. (Unfortunately I buried it by making another commit)

Sources/Observer.swift
andersio: This is strongly captured `self` by the way, which goes against the contract of `<~`.

Marcocanc: I reverted the commit that fixes that in order to un-hide the discussion that was going on. Will change it again if/when we decide how to proceed with this PR.

Marcocanc:
I had two main motivations for this PR:
1. When binding ViewModels that use `Observer`s as inputs it's not very nice to have two ways of binding (`<~` operator for outputs and `observe*` for inputs). With the binding operator the target will be on the left, whereas with `observe` it will be on the right.

2. I use MVVM for UICollectionViewCells extensively. In that case I want the observer to only receive values and no terminating events (otherwise the ViewModel's pipe would break whenever the cell is reused since the binding is disposed)

@andersio Here's another approach:
```swift
extension Signal.Observer {
    public static func <~
        <Source: BindingSource>
        (observer: Signal<Value, Error>.Observer, source: Source) -> Disposable?
        where Source.Value == Value, Source.Error == NoError
    {
        return source.producer.start { [weak observer] in
            switch $0 {
            case .value(let val):
                observer?.send(value: val)
            default: break
            }
        }
    }
}
```

andersio:
@Marcocanc I think the approach in https://github.com/ReactiveCocoa/ReactiveSwift/pull/635#issuecomment-387016433 is fair. :)

Marcocanc:
@andersio I have updated the PR with the proposed approach 😊 Sorry it took me so long

Sources/UnidirectionalBinding.swift
```diff
@@ -102,6 +102,37 @@ extension BindingTargetProvider {
...
+	/// - returns: A disposable that can be used to terminate binding before the
+	///            deinitialization of the target or the source's `completed`
+	///            event.
+	@discardableResult
+	public static func <~
+		<Source: BindingSource>
+		(observer: Signal<Value, Error>.Observer, source: Source) -> Disposable?
+		where Source.Value == Value
+	{
+		return source.producer.start { [weak observer] in
```
mdiep: I think this can use `startWithValues`?

Marcocanc: Good catch!

mdiep:
Thanks for the PR! Sorry this took so long. 🙈

PR #636, opened 2018-04-27
Make non-class types be able to conforms to ReactiveExtensionProvider

ra1028:
This is very tiny changes.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 1. New operator `map(value:)` (#601, kudos to @ra1028)
 1. `SignalProducer.merge` now accepts any combination of `SignalProducerConvertible` conforming types (#610, kudos to @1028)
 1. Bag can be created with the initial elements now (#609, kudos to @ra1028)
+1. Non-class types now can be conforms to ReactiveExtensionProvider (#636, kudos to @ra1028)
 
 # 3.1.0
 1. Fixed `schedule(after:interval:leeway:)` being cancelled when the returned `Disposable` is not retained. (#584, kudos to @jjoelson)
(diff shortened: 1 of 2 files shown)


mdiep:
I know there's been some discussion of removing this requirement before. Can you share an example of a non-class type where you'd like to use it?

mdiep:
I'm not sure this is something that we want to support.

Reactive extensions only really make sense with a reference type. If you were trying to change a `struct` or `enum`, you'd have problems. I'm not sure if enabling other wrappers like this is worth it given that it might be confusing to users.

@ReactiveCocoa/reactivecocoa?

sharplet:
I have had this particular use case before when writing extensions for Kingfisher, which uses a `.kf` extension provider for its UIKit wrapper functionality. The value returned from `.kf` is a class, which explains why I didn't have this limitation.

I'm not currently opposed to loosening the restriction, although as you say it doesn't make a lot of sense for binding targets. @mdiep is that the confusion you're thinking of? That users might make a binding target to a struct and then be confused that it doesn't mutate shared state?

mdiep:
> @mdiep is that the confusion you're thinking of? That users might make a binding target to a struct and then be confused that it doesn't mutate shared state?

Yup.

andersio:
> That users might make a binding target to a struct and then be confused that it doesn't mutate shared state?

This is unlikely to happen since `BindingTarget` does not support `inout` reference. So if the value is value type, mutations attempted in the closure would be blocked by the compiler.

ra1028:
> Either way, this is a semantic error that IMO we shouldn't account for.

Agreed.

@mdiep 
How about?

PR #640, opened 2018-05-16
Update Package.swift to match Cartfile

liscio:
I noticed that the `Package.swift` definitions did not match those of the `Cartfile`s while working on a server-side project that imported the latest release of `Result`.


mdiep:
Eek! 🙈 Good catch!

PR #643, opened 2018-05-27
Rewrite names of a few test case. Disable four test cases.

andersio:
#### Checklist
~~Updated CHANGELOG.md.~~

### DIFF
--- a/Tests/ReactiveSwiftTests/FlattenSpec.swift
+++ b/Tests/ReactiveSwiftTests/FlattenSpec.swift
@@ -645,22 +645,34 @@ class FlattenSpec: QuickSpec {
 					.flatMap(.latest) { _ in Property(value: 0) }
 			}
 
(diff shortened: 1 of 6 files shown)


Tests/ReactiveSwiftTests/FlattenSpec.swift
```diff
@@ -645,22 +645,22 @@ class FlattenSpec: QuickSpec {
 					.flatMap(.latest) { _ in Property(value: 0) }
 			}
 
-			it("sould available to use contextual lookup for flatMap arbitrary error signal to arbitrary error signal") {
+			it("should available to use contextual lookup for flatMap arbitrary error signal to arbitrary error signal") {
```
mdiep: This still doesn't make sense. 😕 Maybe it should be `should be able to use…`?

PR #644, opened 2018-06-06
Xcode 10 support

ikesyo:
#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/Sources/ValidatingProperty.swift
+++ b/Sources/ValidatingProperty.swift
@@ -178,8 +178,8 @@ public final class ValidatingProperty<Value, ValidationError: Swift.Error>: Muta
 		self.init(MutableProperty(initial), with: other, validator)
 	}
 	
-	/// Create a `ValidatingProperty` that presents a mutable validating
-	/// view for an inner mutable property.
+	/// Create a `ValidatingProperty` which validates mutations before
+	/// committing them.
 	///
 	/// The proposed value is only committed when `valid` is returned by the
 	/// `validator` closure.
@@ -199,8 +199,8 @@ public final class ValidatingProperty<Value, ValidationError: Swift.Error>: Muta
 		self.init(MutableProperty(initial), with: other, validator)
 	}
 
-	/// Create a `ValidatingProperty` that validates mutations before
-	/// committing them.
+	/// Create a `ValidatingProperty` that presents a mutable validating
+	/// view for an inner mutable property.
 	///
 	/// The proposed value is only committed when `valid` is returned by the
 	/// `validator` closure.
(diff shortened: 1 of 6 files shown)


Sources/ValidatingProperty.swift
```diff
@@ -178,26 +178,6 @@ public final class ValidatingProperty<Value, ValidationError: Swift.Error>: Muta
...
-	/// The proposed value is only committed when `valid` is returned by the
-	/// `validator` closure.
-	///
-	/// - note: `inner` is retained by the created property.
-	///
-	/// - parameters:
-	///   - inner: The inner property which validated values are committed to.
-	///   - other: The property that `validator` depends on.
-	///   - validator: The closure to invoke for any proposed value to `self`.
-	public convenience init<U, E>(
```
ikesyo: The capture of the compiler error

ikesyo: The following overload will be used instead after this removal: https://github.com/ReactiveCocoa/ReactiveSwift/blob/7c25bc2e284ba54a8d81d3843a59cdd933c8f823/Sources/ValidatingProperty.swift#L128-L132

sharplet: I think this makes sense 👍

Tests/ReactiveSwiftTests/SignalProducerSpec.swift
```diff
@@ -2791,7 +2791,7 @@ class SignalProducerSpec: QuickSpec {
 					let producer = SignalProducer<Int, NoError>.never
 						.on(disposed: { disposed = true })
 
-					var replayedProducer = ImplicitlyUnwrappedOptional(producer.replayLazily(upTo: 1))
+					var replayedProducer = Optional(producer.replayLazily(upTo: 1))
```
sharplet: So we were using it as a regular optional anyway? Huh.

mdiep:
Do you want to move the headers phase before the compile sources phase as part of this?

ikesyo:
> Do you want to move the headers phase before the compile sources phase as part of this?

Addressed in eb2bae1!

Sources/ValidatingProperty.swift
andersio: This changes the semantic. [This overload](https://github.com/ReactiveCocoa/ReactiveSwift/pull/644/files/e27310de692416a4fd70edfd1896ff3245ca4b51#diff-42b514003f72de821971eab375eda51dR192) was supposed to be the primary one, and the removed one here should have been implemented in terms of that.

ikesyo: So what should be done here to avoid the compilation error? @andersio

ikesyo: Feel free pushing your changes on this branch.

PR #646, opened 2018-06-12
updating Result dependency in ReactiveSwift.podspec to 4.0'

mishagray:
This sets the cocoapod dependency to match the current dependency set in Carthage.

#### Checklist
- [ ] Updated CHANGELOG.md.


mdiep:
Thanks!

PR #656, opened 2018-07-12
add .true and .false signal producers which sends a boolean and compl…

mvnrc:
…etes immediately

#### Checklist
- [ ] Updated CHANGELOG.md.

Useful for **zip** operator for example, when you want to bypass a signal producer in a specific condition.


mdiep:
Thanks for opening a pull request, but I don't think we'd want these. Instead, you should use `init(value:)`:

```swift
SignalProducer(value: true)
```

andersio:
`ExpressibleByBooleanLiteral` could be an alternative, but I am not sure if this is a good idea.

PR #664, opened 2018-08-15
Promise draft

leonid-s-usov:
Not really a pull request, but rather a dirty draft in a sandbox, for review.
Inspired by [PromiseKit](https://github.com/mxcl/PromiseKit/blob/master/Sources/Thenable.swift)

Some highlights:

    struct Promise<Value, Error: Swift.Error>: Thenable, SignalProducerConvertible {

```
extension SignalProducer {
    func makePromise() -> Promise<Value, Error> {
        return Promise<Value, Error>(producer: self)
    }
}

func add28(to term: Int) -> Promise<Int, NoError> {
    return Promise { fulfil, _ in
        fulfil(term + 28)
        return AnyDisposable()
    }
}

let answer = SignalProducer<Int, NoError>(value: 14)
    .makePromise()
    .then { add28(to: $0) }

answer.await { (result) in
    print(result)
}
```

```
> fulfilled(42)
```

### DIFF
--- a/ReactiveSwift.playground/Pages/Sandbox.xcplaygroundpage/Contents.swift
+++ b/ReactiveSwift.playground/Pages/Sandbox.xcplaygroundpage/Contents.swift
@@ -98,34 +98,14 @@ struct Promise<Value, Error: Swift.Error>: Thenable, SignalProducerConvertible {
         }
     }
     
-    private let state: Property<State>
+    fileprivate let state: Property<State>
     
     private let (_lifetime, _lifetimeToken) = Lifetime.make()
     
     var lifetime: Lifetime { return _lifetime }
     
     var producer: SignalProducer<Value, Error> {
-        return SignalProducer { [weak state] observer, lifetime in
-            guard let state = state else {
-                observer.sendInterrupted()
-                return
-            }
-            
-            lifetime += state.producer.startWithValues {
-                guard case .resolved(let result) = $0 else {
-                    return
-                }
-                switch result {
-                case .fulfilled(let value):
-                    observer.send(value: value)
-                    observer.sendCompleted()
-                case .rejected(let error?):
-                    observer.send(error: error)
-                case .rejected(nil):
-                    observer.sendCompleted()
-                }
-            }
-        }
+        return SignalProducer(self)
     }
     
     init(_ resolver: @escaping (@escaping (Result) -> Void) -> Disposable) {
@@ -182,7 +162,26 @@ struct Promise<Value, Error: Swift.Error>: Thenable, SignalProducerConvertible {
     }
 }
 
-
+extension SignalProducer {
+    init(_ promise: Promise<Value, Error>) {
+        self.init { observer, lifetime in
+            lifetime += promise.state.producer.startWithValues {
+                guard case .resolved(let result) = $0 else {
+                    return
+                }
+                switch result {
+                case .fulfilled(let value):
+                    observer.send(value: value)
+                    observer.sendCompleted()
+                case .rejected(let error?):
+                    observer.send(error: error)
+                case .rejected(nil):
+                    observer.sendCompleted()
+                }
+            }
+        }
+    }
+}
 
 extension SignalProducer {
     func makePromise() -> Promise<Value, Error> {
--- a/ReactiveSwift.playground/Pages/Sandbox.xcplaygroundpage/Contents.swift
+++ b/ReactiveSwift.playground/Pages/Sandbox.xcplaygroundpage/Contents.swift
@@ -60,8 +60,8 @@ extension Thenable {
     
     @discardableResult
     func chain<T: Thenable>(on: Scheduler = ImmediateScheduler(), body: @escaping (PromiseResult<Value, Error>) -> T) -> Promise<T.Value, Error> where T.Error == Error {
-        return Promise<T.Value, Error> { resolve in
-            return self.await { result in
+        return Promise<T.Value, Error> { resolve, lifetime in
+            lifetime += self.await { result in
                 on.schedule {
                     let thenable = body(result)
                     thenable.await(notify: resolve)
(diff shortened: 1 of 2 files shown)


leonid-s-usov:
@RuiAAPeres WDYT?

RuiAAPeres:
Need to spend some time on this, but my initial feeling is that I don't fully agree with this direction. 

I assumed the goal was to abstract a `SignalProducer` via another Layer (Promise). If you make it comply with the `SignalProducerConvertible` protocol, I don't see any advantage on this extra layer.

leonid-s-usov:
The SignalProducerConvertible is there to be able to use all those convenience functions defined to accept objects adopting this protocol. At the end of the day, the Promise is built on top of a producer and can easily become one for the sake of polymorphism.

It's strange that this simple compliance is changing your attitude. It's pure convenience. 
In any case, one may always want to convert a "OneShot" entity to a SignalProducer, so why not manifest this with the "Convertible" interface?

ReactiveSwift.playground/Pages/Sandbox.xcplaygroundpage/Contents.swift
```diff
@@ -22,4 +22,186 @@ import Foundation
...
+        }
+    }
+    
+    private let state: Property<State>
+    
+    private let (_lifetime, _lifetimeToken) = Lifetime.make()
+    
+    var lifetime: Lifetime { return _lifetime }
+    
+    var producer: SignalProducer<Value, Error> {
```
mdiep: I think this should be `SignalProducer.init(_: Promise<Value, Error>)` instead.

ReactiveSwift.playground/Pages/Sandbox.xcplaygroundpage/Contents.swift
```diff
@@ -22,4 +22,186 @@ import Foundation
  A place where you can build your sand castles 🏖.
 */
 
+enum PromiseResult<Value, Error: Swift.Error> {
```
mdiep: We could just use `Result` for this.

ReactiveSwift.playground/Pages/Sandbox.xcplaygroundpage/Contents.swift
```diff
@@ -22,4 +22,186 @@ import Foundation
...
+    
+    var error: Error? {
+        switch self {
+        case .fulfilled(_): return nil
+        case .rejected(let error): return error
+        }
+    }
+}
+
+protocol Thenable {
```
mdiep: I'm not sure we'd get any benefit from this protocol.

ReactiveSwift.playground/Pages/Sandbox.xcplaygroundpage/Contents.swift
leonid-s-usov: Well that was my intention in the first place, but then I couldn't work around the case with `nil` error.
That's actually a questionable decision to store it this way, other implementations chose to wrap a user error into additional error enum like

```
enum PromiseError: Error {
     case nested(error)
     case interrupted
}
```

In this case standard `Result` would do. I'm open for suggestions here - including an option to `fatalError` the case of interrupted / empty source producer. Which is of course not something I'd agree for ;)

ReactiveSwift.playground/Pages/Sandbox.xcplaygroundpage/Contents.swift
leonid-s-usov: Yeah.. I just snatched it from PromiseKit, after [reading the specs](https://promisesaplus.com/) which refer to  `Thenable` as something separate from the `Promise` itself. Technically, there could be thin auxiliary classes implementing the protocol directly, without the overhead of the full `Promise`

ReactiveSwift.playground/Pages/Sandbox.xcplaygroundpage/Contents.swift
leonid-s-usov: Yeah, agreed. `state` can be made `fileprivate` and an extension for `SignalProducer` could be defined. Also it would be cleaner to make this a `let` property cause there's no reason in creating a new signal producer for every caller

leonid-s-usov: done

RuiAAPeres:
> It's strange that this simple compliance is changing your attitude. It's pure convenience.
In any case, one may always want to convert a "OneShot" entity to a SignalProducer, so why not manifest this with the "Convertible" interface?

The goal of having a `Promise` is to hide the apparent "complexity" of a `SignalProducer`/`Signal` behind an easy to digest API. From the moment you add `SignalProducerConvertible`, you are opening the pandora box, so the big advantage (simplicity) is lost. I fear I have completely misunderstood your intentions/ambitions/motivations with this new entity. 

It would be interesting to see how many people, besides you, would be interest in using it like that. **I think this is a case of less is more.**

leonid-s-usov:
Sorry, I still can't see what it changes. How can the fact that Promise is convertible back to SignalProducer compromise its simplistic API?

Let's for a moment remove this compliance. Do you agree that there should be a way to create a SignalProducer out of a Promise, given that these two are part of the same framework? Just like you have this ability with Signal and Property, both can be rendered into a SignalProducer.

If yes, then that's it, you take a Promise, convert it to a SignalProducer and you have the same pandora box. So what is different if there is no explicit conformance to SignalProducerConvertible?

I mean.. I don't really mind consider removing that protocol, I just want to understand. And it's a pity to hear that you "completely misunderstood my original intention", cause I feel like I have achieved exactly what I wanted. I have a promise now, I can construct it from scratch or convert a SignalProducer to it, I have the promise-like behavior and there's nothing that stops me from even implementing the Promises A+ (apart from the time, of course)

ReactiveSwift.playground/Pages/Sandbox.xcplaygroundpage/Contents.swift
leonid-s-usov: I have refactored the code, and now Thenable is a type-erased ThenableType, which allows for a number of nice optimisations

andersio:
ref: https://github.com/ReactiveCocoa/ReactiveSwift/pull/668#issuecomment-974140038

Issue #667, opened 2018-08-17
New API to work with single value streams


leonid-s-usov:
This issue is created in an attempt to help proceed with the discussion happening in the [Stream of one event #661](https://github.com/ReactiveCocoa/ReactiveSwift/issues/661) proposal. 

The issue is around providing a better API for cases like
```
public protocol ChatService {
    func listAll() -> S<[Conversation], ChatError>
    func fetchMessages(for conversation: Conversation.ID) -> S<[Message], ChatError>
}
```

Today RAS gives us two options to put instead of S: `Signal` and `SignalProducer`. Both options aren't ideal: `Signal` isn't fine for the reason that there are no guarantees that we will not miss the result. `SignalProducer` is the only real option but it doesn't fit the semantics of the protocol - in order for SignalProducer to fit the protocol should be changed to

```
public protocol ChatServiceAccessor {
    func allLister() -> SignalProducer<[Conversation], ChatError>
    func messagesFetcher(for conversation: Conversation.ID) -> SignalProducer<[Message], ChatError>
}
```

Last but not the least - both Signal and SignalProducer fail totally at advertising that the observer will never receive multiple events - there will always be just one value, or an error (actually, there might be also an interruption, depending on whether we decide to allow it here)
***
During the discussion in #661 there was a lot said about the wanted cold / warm nature of the suggested new API but I believe that the best way of summarising the effort there would be to copy here a set of requirements in front of the potential API

1. **We want a concrete type (or several related types) which guarantees to return only a single value, or an error**
2. **We want to have a cold single value stream to control when and how the operation is initiated**
3. **We want to have a warm single value stream to await for the value without caring about when and how the operation has been initiated, similar to how Future or Promise works**
4. **We want a clearer (and more efficient) flattenMap strategy when chaining streams of single value**
5. **We want other single value oriented operators, which would be more efficient, readable and safe**
6. **We want to be able to easily convert the new classes back to a SignalProducer for use in some generic operators**

leonid-s-usov:
@jjoelson wrote:

>  In #668, it seems the example would look almost exactly the same with a SignalProducer except the need to provide a flatten strategy is gone, on has been renamed when, and start has been renamed await. Isn't a major goal of a framework like ReactiveSwift to have a common vocabulary of types and operators that can represent streams of values over time? Bracketing off a particular subset of use cases and changing all of the names seems counter to the mission as I understand it.
> 
[...]

PR #668, opened 2018-08-18
AsyncValue POC

leonid-s-usov:
A proof of concept for a set single value stream targeted classes:
`AsyncValue` and `AsyncValueProducer`, as per #667 

`AsyncValue` has promise-like semantics, and `AsyncValueProducer` is its cold brother.

`AsyncValue` can be constructed from a SignalProducer or AsyncValueProducer and will start the producer immediately, and will cache its first value or error

Operators are defined as extensions to a protocol `AsyncValueType`, and this way they are applicable to both `AsyncValue` and `AsyncValueProducer`. Operator chain semantics depends on whether one decides to keep all chained values cold - then the resulting instance will also be cold and will need to be `start()`ed. 
`await()` will implicitly start a cold value

Like in this usage
```
var seed = 0

let seedProducer = SignalProducer<Int, NoError> { obs, _ in
    obs.send(value: seed)
}

let value = AsyncValueProducer(seedProducer)
    .map {
        // try changing the below between AsyncValue and AsyncValueProducer
        AsyncValueProducer($0 + 10)
    }
    .when(ready: { print("resolved with value \($0)") })
    
print("incrementing seed")
seed += 1

print("awaiting for a result")
    value.await(notify: { print("await notified with \($0)") })
```

### DIFF
--- a/ReactiveSwift.playground/Pages/Sandbox.xcplaygroundpage/Contents.swift
+++ b/ReactiveSwift.playground/Pages/Sandbox.xcplaygroundpage/Contents.swift
@@ -175,17 +175,25 @@ let seedProducer = SignalProducer<Int, Never> { obs, _ in
     obs.send(value: seed)
 }
 
-
-let value = AsyncValueProducer(seedProducer)
-    .map {
-        // try changing the below between AsyncValue and AsyncValueProducer
-        AsyncValueProducer($0 + 10)
-    }
-    .when(ready: { print("resolved with value \($0)") })
+let test = {
+    let value =
+        AsyncValueProducer(seedProducer)
+        .map {
+            // try changing the below between AsyncValue and AsyncValueProducer
+            AsyncValue($0 + 10)
+            
+            // NB! Returnign an AsyncValue immediately starts the mapped producer
+            // because an AsyncValue is a hot signal, and it will need to start the
+            // producer in order to try an resolve its value
+        }
+        .when(ready: { print("resolved with value \($0)") })
     
-print("incrementing seed")
-seed += 1
+    print("incrementing seed")
+    seed += 1
 
-print("awaiting result")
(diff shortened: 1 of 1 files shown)


leonid-s-usov:
Hi, I've never received any feedback on this one. What do you guys think? 
I don't think that this implementation requires any kind of interruption handling whether directly or by introducing a special `cancelled` state. If the resolving producer never provides result then the async value never completes. Can be mitigated by a timeout.

andersio:
Hi, sorry for the long hiatus on this large and important topic.

I have been drafting on and off an implementation strategy, as part of a new overhaulling initiative, to bring in both cold and hot streams-of-one based on concepts from #722. It should hopefully address value proposition and implementation/maintainance concerns of these primitives, that were brought up in previous discussions. It also make use of Swift's generic system to its full extent (and potentially upcoming features e.g. parameterized extensions). I really appreciate your devotion in prototyping these ideas, and your arguments are a valuable angle for evaluating several possible approaches. 🙇

leonid-s-usov:
So what's the API you are considering? WIll this PR be used? Any help needed?

john-twigg-ck:
I've been following this thread for some time.
We have a strong need of `Single` or the "streams-of-one" concept. 
The complexity that is forced upon our code when having to consider streams of `>=2` (or even `0` events followed by completed) is really challenging.

I look forward to this addition.

andersio:
Now that Swift Concurrency is out with backward deployment to iOS 13.0+ in the works, async functions are the natural construct for all these common "stream of one" asynchronous programming use cases. This is especially true since the language itself has to guarantee that your code awaiting a suspension point will resume **exactly** once, unlike a closure-based solution in the current FRP implementations.

So given the current landscape, expanding in this direction no longer seems valuable to pull RAS out of maintenance mode.

PR #673, opened 2018-09-10
Lifecycle

arnopoulos:
This adds the ability for a consumer of ReactiveSwift to have a custom life cycle without having to manually create the scaffolding for the `Lifetime`. In addition this has the added benefit of making `Lifetime.Token`s implicit mutual exclusion rule and makes it explicit.

#### Checklist
- [ ] Updated CHANGELOG.md.

### DIFF
--- a/ReactiveSwift.xcodeproj/project.pbxproj
+++ b/ReactiveSwift.xcodeproj/project.pbxproj
@@ -9,6 +9,16 @@
 /* Begin PBXBuildFile section */
 		02D2602A1C1D6DAF003ACC61 /* SignalLifetimeSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 02D260291C1D6DAF003ACC61 /* SignalLifetimeSpec.swift */; };
 		02D2602B1C1D6DB8003ACC61 /* SignalLifetimeSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 02D260291C1D6DAF003ACC61 /* SignalLifetimeSpec.swift */; };
+		490B642D2145F9D1008D5D9C /* Lifecycle.swift in Sources */ = {isa = PBXBuildFile; fileRef = 490B642C2145F9D1008D5D9C /* Lifecycle.swift */; };
+		490B642E2145F9D1008D5D9C /* Lifecycle.swift in Sources */ = {isa = PBXBuildFile; fileRef = 490B642C2145F9D1008D5D9C /* Lifecycle.swift */; };
+		490B642F2145F9D1008D5D9C /* Lifecycle.swift in Sources */ = {isa = PBXBuildFile; fileRef = 490B642C2145F9D1008D5D9C /* Lifecycle.swift */; };
+		490B64302145F9D1008D5D9C /* Lifecycle.swift in Sources */ = {isa = PBXBuildFile; fileRef = 490B642C2145F9D1008D5D9C /* Lifecycle.swift */; };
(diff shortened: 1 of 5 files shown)


mdiep:
Sorry, but I'm unclear (1) what problem this solves and (2) how it solves it. Could you share a little more detail please?

arnopoulos:
@mdiep Sorry about the delayed response. This solves a mutual exclusion problem with `Lifetime.Token` and simplifies creating custom life cycles. Take the following example:

```
protocol Items {
	//Things an item would need
}

class Store {
	private var (lifetime, token): (Lifetime?, Lifetime.Token?)
	var items: Property<[Items]>! {
		didSet {
			if token != nil {
				open()
			}
		}
	}

	func open() {
		let (lifetime, token) = Lifetime.make()
		self.lifetime = lifetime
		self.token = token

		items.producer.take(during: lifetime).startWithValues { (items) in
			//Do something here
		}
	}

	func close() {
		token = nil
	}
}
```

The example is simple but it shows the mutual exclusion problem and some of the problems with manual life cycle management. The first is because of potential user error and (intentional) improper access control a consumer of `Store` could hold a strong reference to `Lifetime.Token`. In addition there isn't a single code path to guarantee that your reactive components are going to be started. For example, in the example you could imagine that `Store` could have another function that opens at a specific time `open(at time: Date)`. In that function you can imagine that a user could forget to start the items signal producer which can be an issue.

What I'm proposing is something to the effect of this:

```
class Store {
	lazy var lifecycle: Lifecycle = {
		let lifecycle = Lifecycle()
		lifecycle.lifetime.producer.startWithValues { [weak self] lifetime in
			self?.items.producer.take(during: lifetime).startWithValues { items in
				//Do something here
			}
		}
		return lifecycle
	}()
	var items: Property<[Items]>! {
		didSet {
			lifecycle.updateIfValid()
		}
	}

	func open() {
		lifecycle.update()
	}

	func close() {
		lifecycle.invalidate()
	}
}
```

In addition this makes the `open(at time: Date)` example relatively trivial to implement:

```
class Store {
	private var timer: Timer?
	lazy var lifecycle: Lifecycle = {
		let lifecycle = Lifecycle()
		lifecycle.lifetime.producer.startWithValues { [weak self] lifetime in
			self?.items.producer.take(during: lifetime).startWithValues { items in
				//Do something here
			}
		}
		return lifecycle
	}()
	var items: Property<[Items]>! {
		didSet {
			lifecycle.updateIfValid()
		}
	}

	func open() {
		timer?.invalidate()
		lifecycle.update()
	}

	@available(OSX 10.12, *)
	func open(at date: Date) {
		timer = Timer(fire: date, interval: 1.0, repeats: false) { [weak self] timer in
			self?.lifecycle.update()
		}
	}

	func close() {
		lifecycle.invalidate()
	}
}
```

Another place where I and the team I work on end up using this a lot is in UI logic specifically, view controllers and reusable views such as table/collection cells.

arnopoulos:
@mdiep Have you had time to look over this PR again?

arnopoulos:
@mdiep Have you had time to look over this PR again?

mdiep:
Sorry it took me so long to look at this.

Unfortunately, I don't think this makes sense to include in ReactiveSwift. We try to focus on reactive primitives, and I don't think this qualifies. I've been doing reactive programming for a while now, and I've never seen anyone do something like this.

As an aside, I think you might be able to find a better abstraction. In particular, from the limited code samples that you've shared, it looks like it might make more sense to have a `Property<Store>` instead of a `Store` with a `Property`—or to create another type to serve as the thing being observed.

andersio:
Closing as this seems to be concluded.

PR #679, opened 2018-10-04
Add materializeResults operator

ra1028:
This operator is useful when we want to handle Result without to complete stream.
It has semantics similar to `dematerialize`.
And `dematerializeResults` which works in inverse is also added.

Use case:
```swift
// This stream doesn't complete even if an error sent.
viewModel.reactive.date.producer
    .take(duringLifetimeOf: self)
    .flatMap(.latest) { date in
        repository.getPrograms(on: date)
            .materializeResults()
    }
    .startWithValues { result in
        switch result {
        case .success(let programs)
            print(programs)

        case .failure(let error)
            print(error)
    }
```

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # master
 *Please add new entries at the top.*
 
+1. New operator `matelializeResults` and `dematerializeResults` (#679, kudos to @ra1028)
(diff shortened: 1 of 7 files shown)


mdiep:
Awesome! 👍🏻

I've wanted this for a while, but I had a hard time deciding on a name. `materializeResults` and `dematerializeResults` seem like good choices.

Issue #686, opened 2018-10-25
Question about `SerialDisposable`


DevAndArtist:
I'm currently learning the project while trying to understand every single bit of it. 

As far as I understood the design of `Disposable` types included in the `Disposable.swift` file, I think I might have found a small issue (maybe harmless, but potentially a bug?).

For example `CompositeDisposable` [will clear](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/Disposable.swift#L164-L167) the `Atomic`'s value and potentially remove all the retained references to the `Disposable` objects. Then if you try to add a new disposable to the `CompositeDisposable`, but the `CompositeDisposable` itself is already disposed, it will dispose the passed `Disposable` and reject retaining the reference into the bag.

Now if I compare that behavior with `SerialDisposable`, the `dispose()` method [does the same](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/Disposable.swift#L372-L374). However the setter on the `inner` property does not reject the value if the `SerialDisposable` is already disposed, but [retains the reference](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/Disposable.swift#L353-L357).

Is this by design or just overlooked?

I thought it would be more like this:

```diff
set(disposable) {
+  if isDisposed {
+    disposable?.dispose() // Dispose and do not retain
+  } else {
    _inner.swap(disposable)?.dispose()
+  }
-
-  if let disposable = disposable, isDisposed {
-    disposable.dispose()
-  }
}
```

DevAndArtist:
Hi @mdiep thank you for answering my question and sorry about closing / re-opening the issue (I was testing GitHawk on the iPad while browsing this repo). I do not follow your point, can you elaborate further?

The misalignment that I discovered here is that when `SerialDisposable` is already disposed you can take a different `Disposable` (disposed or not, it does not matter) and assign to inner so it will still be retained. Sure this is unlikely to happen anywhere because doing that makes simply less sense, but it's a potential source for an issue by allowing retaining a new disposable on a already disposed `SerialDisposable`.

mdiep:
What I mean is that adding a disposable to a `CompositeDisposable` is a failable operation. You call `add` and get back get back a `Disposable?` that tells you whether your disposable was added. Because `SerialDisposable` lets you assign to a property, it can't return a value that indicates whether your disposable was set as the inner disposable.

DevAndArtist:
[...]

PR #688, opened 2018-10-27
Experimental: Discarding new disposable on a disposed `SerialDisposable`

DevAndArtist:
This is an experimental PR to run tests against the [change](https://github.com/ReactiveCocoa/ReactiveSwift/issues/686). Please **DO NOT MERGE** iff this change goes against the design.

#### Checklist
- [ ] Updated CHANGELOG.md.

### DIFF
--- a/Sources/Disposable.swift
+++ b/Sources/Disposable.swift
@@ -350,10 +350,10 @@ public final class SerialDisposable: Disposable {
 		}
 
 		set(disposable) {
-			_inner.swap(disposable)?.dispose()
-
-			if let disposable = disposable, isDisposed {
-				disposable.dispose()
+			if isDisposed {
+				disposable?.dispose()
+			} else {
+				_inner.swap(disposable)?.dispose()
 			}
 		}
 	}


Sources/Disposable.swift
```diff
@@ -350,10 +350,10 @@ public final class SerialDisposable: Disposable {
...
 
 		set(disposable) {
-			_inner.swap(disposable)?.dispose()
-
-			if let disposable = disposable, isDisposed {
-				disposable.dispose()
+			if isDisposed {
+				disposable?.dispose()
+			} else {
+				_inner.swap(disposable)?.dispose()
```
andersio: We need to check `isDisposed` again after swapping, since `self` could be disposed of in parallel during the process, racing against https://github.com/ReactiveCocoa/ReactiveSwift/pull/688/files#diff-7056bb5d45375e26f2eeb3adeabac16bR373.

andersio: Change suggestions do not seem to be available in this repo.

Essentially, we need to do this after the swapping in the else path:

```
if isDisposed {
    _inner.swap(nil)?.dispose()
}
```

Because `dispose()` can be called in parallel to `inner = new` on different threads. Depending on the order observed by the system memory, it could yield a boundary case where `isDisposed` is `true` but the inner disposable is left untouched. The conflicting path may 

> I always thought that such operations should not be done in a multi-threading fashion, so it will be the users responsibility to serialize correctly.

Disposables have always meant to be thread safe, despite us not explicitly having written this as part of the contract.

mdiep: > Change suggestions do not seem to be available in this repo

Weird. I don't see a setting for it.

mdiep:
> Semantically it is the right fix

Can you elaborate on why you think this is the case? I shared some thoughts on #686, but I don't see why this is necessarily better or more correct.

andersio:
@mdiep This PR doesn’t change the semantic of `SerialDisposable` disposing of any inner disposable set after itself had been disposed of.

I would consider #686 a bug in the sense that the inner disposable has been disposed of, but the storage still retains a strong reference to the disposed inner.

PR #692, opened 2018-11-09
Update ReactivePrimitives.md

EduardLev:
#### Checklist
- [ ] Updated CHANGELOG.md.

Just getting started and was going through this file first. I'd like to contribute starting as early as I can. 

The process in question is the vending machine process. Mutual exclusivity is a property of two events, not one, so saying that a process is mutually exclusive sounds strange. Vending two snacks at the same time would be mutually exclusive. Or it could say that the process is mutually exclusive with itself, or something like that. Anyway it sounded strange as I was reading, and the line after the dash explains it anyway.


mdiep:
Thanks for the PR!

PR #697, opened 2018-12-18
Fix the `replayLazily` deadlock.

andersio:
#### Checklist
- [ ] Updated CHANGELOG.md.

### DIFF
--- a/Tests/ReactiveSwiftTests/SignalProducerSpec.swift
+++ b/Tests/ReactiveSwiftTests/SignalProducerSpec.swift
@@ -2863,6 +2863,17 @@ class SignalProducerSpec: QuickSpec {
 					expect(deinitValues) == 1
 				}
 			}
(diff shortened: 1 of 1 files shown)


Tests/ReactiveSwiftTests/SignalProducerSpec.swift
```diff
@@ -2863,6 +2863,17 @@ class SignalProducerSpec: QuickSpec {
 					expect(deinitValues) == 1
 				}
 			}
+			
+			context("recursion") {
+				it("should not deadlock when it is started again doenstream as a result of a terminating event") {
```
RuiAAPeres: you mean `downstream`?

mluisbrown:
@andersio the new test is passing without any changes to the code, so it looks like the test is not reproducing the problem 😕

PR #701, opened 2019-01-07
RFC: Async file reading & run loop based scheduling

sharplet:
This PR is in two parts: first is a possible new reactive extension for `FileHandle`’s background file reading functionality. I think it would be great for ReactiveSwift to provide first-class support for async file IO.

The second part addresses an awkward problem with wrapping this particular `FileHandle` API: it requires an active run loop in order to work. I couldn’t work out a way to construct a suitable precondition (every thread has access `RunLoop.current`, regardless of whether it’s active and has input sources attached), but similar to our other APIs that require a `DateScheduler` or `QueueScheduler` in order to work, it seemed there was an opportunity for a `RunLoopScheduler` abstraction in the same spirit. This still doesn’t guarantee that the run loop scheduler wraps an active run loop, but having to construct a `RunLoopScheduler` in the first place is a much stronger signal to users of the API. This also provides `RunLoopSchedular.main` for the most common case.

#### Checklist

- [ ] Add tests for `RunLoopScheduler`
- [ ] Add recursive scheduling detection
- [ ] Update CHANGELOG.md.

### DIFF
--- a/Sources/FoundationExtensions.swift
+++ b/Sources/FoundationExtensions.swift
@@ -22,11 +22,11 @@ extension Reactive where Base: FileHandle {
 	/// Returns a SignalProducer that reads the contents of this file in the
 	/// background.
 	///
-	/// - returns: A SignalProducer of the file's data.
+	/// - parameters:
+	///   - scheduler: A run loop scheduler to perform the asynchronous file read.
 	///
-	/// - warning: The returned producer must be started on a thread with an
-	///            active run loop, or it will never complete.
-	public func readToEndOfFile() -> SignalProducer<Data, NoError> {
+	/// - returns: A SignalProducer of the file's data.
+	public func readToEndOfFile(on scheduler: RunLoopScheduler) -> SignalProducer<Data, NoError> {
 		return SignalProducer { [base] observer, lifetime in
 			lifetime += NotificationCenter.default.reactive
 				.notifications(forName: .NSFileHandleReadToEndOfFileCompletion, object: base)
@@ -35,7 +35,7 @@ extension Reactive where Base: FileHandle {
 				.observe(observer)
 
 			base.readToEndOfFileInBackgroundAndNotify()
-		}
+		}.start(on: scheduler)
 	}
 
 	/// Returns a SignalProducer that reads the contents of the file at `url` in
@@ -43,15 +43,13 @@ extension Reactive where Base: FileHandle {
 	///
 	/// - parameters:
 	///   - url: A file URL to read.
+	///   - scheduler: A run loop scheduler to perform the asynchronous file read.
 	///
 	/// - returns: A SignalProducer of the file's data, or an error if the file
 	///            could not be opened.
-	///
-	/// - warning: The returned producer must be started on a thread with an
-	///            active run loop, or it will never complete.
-	public static func readFile(at url: URL) -> SignalProducer<Data, AnyError> {
+	public static func readFile(at url: URL, on scheduler: RunLoopScheduler) -> SignalProducer<Data, AnyError> {
 		return SignalProducer { try FileHandle(forReadingFrom: url) }
-			.flatMap(.latest) { $0.reactive.readToEndOfFile() }
+			.flatMap(.latest) { $0.reactive.readToEndOfFile(on: scheduler) }
 	}
 }
 
--- a/Sources/FoundationExtensions.swift
+++ b/Sources/FoundationExtensions.swift
@@ -18,6 +18,7 @@ import struct Result.AnyError
 
 extension FileHandle: ReactiveExtensionsProvider {}
 
+#if !os(Linux)
 extension Reactive where Base: FileHandle {
 	/// Returns a SignalProducer that reads the contents of this file in the
 	/// background.
@@ -52,6 +53,7 @@ extension Reactive where Base: FileHandle {
 			.flatMap(.latest) { $0.reactive.readToEndOfFile(on: scheduler) }
 	}
 }
+#endif
 
 extension NotificationCenter: ReactiveExtensionsProvider {}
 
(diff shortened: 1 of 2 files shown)


sharplet:
Note: I plan to add tests for this as well, but wanted to open this up for discussion about the API itself. For example: using run loop based concurrency might not be what we want, because it doesn’t compose well with GCD outside of the main queue. So something `DispatchIO` based could solve the same problem, and make use of `QueueScheduler` instead.

sharplet:
(I already have some proof-of-concept level code for a `DispatchIO` wrapper, but it’s far more complex.)

Sources/Scheduler.swift
```diff
@@ -382,6 +382,90 @@ public final class QueueScheduler: DateScheduler {
...
+			repeats: false
+		)
+		CFRunLoopAddTimer(runLoop, timer, .defaultMode)
+
+		return disposable(for: timer)
+	}
+
+	private func disposable(for timer: Timer) -> Disposable {
+		return AnyDisposable { [runLoop] in
+			CFRunLoopPerformBlock(runLoop, CFRunLoopMode.defaultMode.rawValue, timer.invalidate)
```
sharplet: This should probably check the current run loop to avoid the thread hop if possible.

sharplet:
All the CFunLoop stuff isn't compiling on Linux. I'll have to do some investigation there.

ikesyo:
Maybe you'll need to write `import CoreFoundation` explicitly.

Sources/Scheduler.swift
```diff
@@ -382,6 +382,90 @@ public final class QueueScheduler: DateScheduler {
...
+	public init(runLoop: RunLoop) {
+		self.runLoop = runLoop.getCFRunLoop()
+	}
+
+	public var currentDate: Date {
+		return Date()
+	}
+
+	public func schedule(_ action: @escaping () -> Void) -> Disposable? {
+		if CFRunLoopGetCurrent() == runLoop {
```
andersio: FIFO order needs to be maintained in case of recursive scheduling.

mdiep:
This seems like a fine thing to include. 👍🏻 (But I haven't given the code a thorough review.)

sharplet:
Thanks, please take your time!

mdiep:
Want to fix up the Linux build first? I'm curious if that will have any impact on the code.

NachoSoto:
Nice 👍

Sources/Scheduler.swift
sharplet: I’ve been thinking about this more today, and I think that because of `CFRunLoopRun()` and friends, we’ll have to use a different technique, because callers can always recursively run the run loop out of our control.

I think we might be able to maintain some thread-local state that tracks whether we’re already running inside `RunLoopScheduler`, and then potentially reschedule if that’s already the case?

But again, the problem is this call tree:

```
RunLoopScheduler.schedule(_:)         // (1) sets thread-local flag
    CFRunLoopRun()                    // (2) potentially spins indefinitely
        RunLoopScheduler.schedule(_:) // (3) we'll never be able to schedule unless we call CFRunLoopStop(_:)?
```

Another alternative may be to simply not provide any guarantees: if you’re opting into run loop scheduling, be aware of the pitfalls of recursively spinning the run loop.

sharplet:
So I started trying to get `RunLoopScheduler` to compile on Linux, and ran into a couple of issues.

The first is that `Timer` isn’t toll-free bridged on Linux right now, and there’s no access to the underlying `CFRunLoopTimer` object, so we’d essentially have to [reimplement the `Timer` class](https://github.com/apple/swift-corelibs-foundation/blob/c7b997d1a90d0d152eb4072f0776a95847065b1f/Foundation/Timer.swift) to get this to compile.

The second is that [background file reading isn’t even implemented yet](https://github.com/apple/swift-corelibs-foundation/blob/a61b058ed53b00621e7acba4c53959e3ae01a254/Foundation/FileHandle.swift#L326-L328)! 😅 I’ve pushed commit c7bc699 that disables this feature on Linux, in order to get things compiling.

Sources/FoundationExtensions.swift
```diff
@@ -16,6 +16,45 @@ import struct Result.AnyError
...
+#if !os(Linux)
+extension Reactive where Base: FileHandle {
+	/// Returns a SignalProducer that reads the contents of this file in the
+	/// background.
+	///
+	/// - parameters:
+	///   - scheduler: A run loop scheduler to perform the asynchronous file read.
+	///
+	/// - returns: A SignalProducer of the file's data.
+	public func readToEndOfFile(on scheduler: RunLoopScheduler) -> SignalProducer<Data, NoError> {
```
mdiep: `NoError` doesn't seem correct here. The documentation is pretty light on details, but the `userInfo` can contain `NSFileHandleError`, "An `NSNumber` object containing an integer representing the UNIX-type error which occurred".

Sources/FoundationExtensions.swift
```diff
@@ -16,6 +16,45 @@ import struct Result.AnyError
...
+		return SignalProducer { [base] observer, lifetime in
+			lifetime += NotificationCenter.default.reactive
+				.notifications(forName: .NSFileHandleReadToEndOfFileCompletion, object: base)
+				.take(first: 1)
+				.map { $0.userInfo![NSFileHandleNotificationDataItem] as! Data }
```
mdiep: Is this guaranteed to be present if an error occurred?

Sources/FoundationExtensions.swift
sharplet: Thanks for digging that up. I recall doing a search in the docs for error handling info at some point, but didn't find this!

sharplet:
Unfortunately I don't have the bandwidth to pick this back up again. I've pushed this code to the [async-read-file branch of my fork](https://github.com/sharplet/ReactiveSwift/tree/async-read-file), in case anyone wants to take a look.

PR #702, opened 2019-01-15
Removing Result in favor of Swift.Result

NachoSoto:
I got out of my cockpit for a day to do some coding and experiment with this.

I understand we'd be dropping support for anything prior to Swift 5.0, so this is kinda just a proof of concept until the next major version?

Code is almost compiling, I'll try to finish it tomorrow. 

#### Checklist
- [x] Update `Quick` and `Nimble`.
- [x] Fix tests.
- [ ] Updated CHANGELOG.md.
- [x] Update `Package.resolved` (not sure how)
- [x] Deprecation for `NoError` (with `Never` fixit)


Sources/ResultExtensions.swift
```diff
@@ -1,7 +1,41 @@
...
 		return .init(result: self)
 	}
 }
+
+/// An “error” that is impossible to construct.
+///
+/// This can be used to describe `Result`s where failures will never
+/// be generated. For example, `Result<Int, NoError>` describes a result that
+/// contains an `Int`eger and is guaranteed never to be a `failure`.
+public enum NoError: Swift.Error, Equatable {
```
ikesyo: `NoError` can be replaced by `Never`.

- https://github.com/apple/swift-evolution/blob/master/proposals/0215-conform-never-to-hashable-and-equatable.md
- https://github.com/apple/swift/pull/16857

NachoSoto: Oh that's right!! I'll get rid of `NoError`.

andersio:
@mdiep Do we not want to keep `NoError` as a typealias?

mdiep:
That's not a bad idea. Maybe a deprecated typealias? That would ease migration.

NachoSoto:
This is 90% I think. If anyone wants to finish this please go ahead 🙏

NachoSoto:
> That's not a bad idea. Maybe a deprecated typealias? That would ease migration.

Done!

mdiep:
I think this is ready, but I'd like to [get a 5.0.1 release](https://github.com/ReactiveCocoa/ReactiveSwift/pull/721) in first to fix the warnings in Xcode 10.2.

Tests/ReactiveSwiftTests/TestError.swift
```diff
@@ -22,16 +21,16 @@ internal extension SignalProducer {
 	/// Halts if an error is emitted in the receiver signal.
 	/// This is useful in tests to be able to just use `startWithNext`
 	/// in cases where we know that an error won't be emitted.
-	func assumeNoErrors() -> SignalProducer<Value, NoError> {
-		return self.lift { $0.assumeNoErrors() }
+	func assumeNevers() -> SignalProducer<Value, Never> {
```
andersio: This should not be renamed since the re al meaning is now lost.

NachoSoto: Agreed. The "no error" part of this method name was beyond the `NoError` type.

mdiep: Done. 👍

PR #711, opened 2019-03-03
Update Quick to 2.0.0 and Nimble to 8.0.0

ikesyo:
- https://github.com/Quick/Quick/releases/tag/v2.0.0
- https://github.com/Quick/Nimble/releases/tag/v8.0.0

This is required to build using Xcode 10.2. But we require Swift 4.2 at least now, so we must drop Swift 4.1 support.

#### Checklist
- ~~[ ] Updated CHANGELOG.md.~~

### DIFF
--- a/ReactiveSwift.xcodeproj/project.pbxproj
+++ b/ReactiveSwift.xcodeproj/project.pbxproj
@@ -1338,6 +1338,7 @@
 			buildSettings = {
 				INFOPLIST_FILE = Tests/ReactiveSwiftTests/Info.plist;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @loader_path/../Frameworks";
+				MACOSX_DEPLOYMENT_TARGET = 10.10;
 				PRODUCT_NAME = "$(PROJECT_NAME)Tests";
 			};
 			name = Debug;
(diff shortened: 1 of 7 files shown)


mdiep:
I think the deployment target of the macOS tests will need to be changed to 10.10 to match Quick/Nimble.

andersio:
CI for Xcode 9.4 can be removed.

ReactiveSwift.xcodeproj/project.pbxproj
```diff
@@ -1278,7 +1278,7 @@
 				CURRENT_PROJECT_VERSION = 1;
 				ENABLE_TESTABILITY = YES;
 				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
-				MACOSX_DEPLOYMENT_TARGET = 10.9;
+				MACOSX_DEPLOYMENT_TARGET = 10.10;
```
mdiep: Can we update the deployment target of just the test target and leave the framework's deployment target at 10.9?

ikesyo: I'll address that.

ikesyo: Fixed in 8bdcc81.

PR #715, opened 2019-03-11
Bump the version to 5.0

mdiep:
I'd like to release our changes for Xcode 10.2. Since we dropped Swift 4.1, this needs to be a breaking release.

I don't think we need to do a beta release for this, but I'm open to other opinions. 🙂

@ReactiveCocoa/reactiveswift

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # master
 *Please add new entries at the top.*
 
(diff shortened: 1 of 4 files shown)


andersio:
I don’t recall any change with huge impact being made that needs to be staged. So IMO going straight for 5.0 is fine.

PR #720, opened 2019-03-30
Make `Action` as covariant & contravariant functor on each type param

inamiy:
This PR allows to `map` on `Output` and `Error`, and `contramap` on `Input`.

#### Checklist
- [ ] Updated CHANGELOG.md.


Sources/Action.swift
```diff
@@ -70,6 +68,29 @@ public final class Action<Input, Output, Error: Swift.Error> {
...
 
+	private init(
+		execute: @escaping (Input) -> SignalProducer<Output, ActionError<Error>>,
+		deinitToken: Lifetime.Token,
+		lifetime: Lifetime,
+		events: Signal<Signal<Output, Error>.Event, NoError>,
+		disabledErrors: Signal<(), NoError>,
+		isExecuting: Property<Bool>,
+		isEnabled: Property<Bool>
+	) {
```
inamiy: ### Comparison

inamiy/ReactiveSwift@008e9a6e (experiment commit) shows the asymmetric result, which is the same as the following implementation:

```swift
extension Action {
    /// - Note: This can be reintepreted as `action.contramapInput { _ in input }.mapOutput { _ in }`
    public func convertToOpaqueActionWithError(_ input: Input) -> Action<Void, Void, Error> {
        return Action<Void, Void, Error>(enabledIf: isEnabled) { _ in
            self.apply(input)
                .map { _ in }
                .flatMapError { error in
                    switch error {
                    case .disabled:
                        return .empty
                    case .producerFailed(let innerError):
                        return SignalProducer(error: innerError)
                    }
            }
        }
    }
}
```

andersio:
LGTM. I will do documentation afterwards, and potentially rename `mapAll`.

PR #724, opened 2019-04-06
6.0.0

mdiep:
A release with no dependencies! 🎉 

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # master
 *Please add new entries at the top.*
 
+# 6.0.0
+1. Dropped support for Swift 4.2 (Xcode 9)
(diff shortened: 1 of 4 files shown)


iv-mexx:
I'm a huge fan of removing the dependency on Result!

I think it would be good to add a few words about the upgrade process from 5.x to 6.0 to the Changelog.
Right now one would have to look into the code changes of #702 to learn more so I tried to extract the changes from there.

Maybe add something along these lines:

---

* If you have used `Result` only for `ReactiveSwift`, remove all instances of `import Result`, `import enum Result.NoError` or `import struct Result.AnyError` and remove the `Result` Framework from your project.
* Replace all cases where `NoError` was used in a `Signal` or `SignalProducer` with `Never`
* Replace all cases where `AnyError` was used in a `Signal` or `SignalProducer` with `Swift.Error`
---

PR #728, opened 2019-04-21
Delete an unnecessary setup for playground

shingt:
#### Checklist

- [x] Delete an unnecessary setup for playground since it doesn't depend on `antitypical/Result` anymore.


mdiep:
Thanks!

PR #733, opened 2019-04-30
Fix Result extensions ambiguity

nekrich:
Hi.

`public protocol ResultProtocol `, `public var value: Success?` & `public var error: Failure?` brings ambiguity for those who wrote their own extensions or use [antitypical/Result](https://github.com/antitypical/Result).

This PR attempts to fix this issue.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # master
 *Please add new entries at the top.*
+1. Fixed Result extensions ambiguity (#733, kudos to @nekrich)
+
 
 # 6.0.0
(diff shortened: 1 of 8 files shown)


mdiep:
Awesome, thanks! Not sure how we missed this. 🙈

benasher44:
Just a heads up that this was a backward incompatible API change. We were using this, and our build broke after the upgrade from 6.0.0 to 6.1.0.

ikesyo:
As stated https://github.com/ReactiveCocoa/ReactiveSwift/pull/743#issuecomment-503744130 this was considered as a bug.

PR #734, opened 2019-05-03
SignalProducer: avoid recursive start of a sequence of producers

### LINKED ISSUE #729: combineLastest with many SignalProducers crashes
To reproduce, run:

```swift
let scheduler = QueueScheduler(qos: .default, name: "RACScheduler", targeting: nil)

let signals = (0..<700).map { _ -> SignalProducer<Void, NoError> in
    return SignalProducer(value: ())
}

SignalProducer
    .combineLatest(signals)
    .start(on: scheduler)
    .logEvents()
    .start()
```

Expected: SignalProducers run and doesn't crash.
Actual: Crash with EXC_BAD_ACCESS with a huge stack trace.

Using ReactiveSwift v4.0.0

Related to these closed issues #156, https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2384, https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3238 but is still reproducible.

leonid-s-usov:
Instead, generate an array of pipes which can be fed to the transformation method
and then start all producers in a loop, connecting each to the corresponding pipe

This fixes #729 

I suspect that with significantly more effort one could refactor the Signal aggregate strategy to avoid creating the set of pipes by allowing for dynamic aggregation.
However, it may happen that eventually, such refactor will result in a comparable overhead down the line.

### DIFF
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -2169,31 +2169,26 @@ extension SignalProducer {
 		return start(producers, Signal.zip)
 	}
 
-	private static func start<S: Sequence>(_ producers: S, _ transform: @escaping (ReversedCollection<[Signal<Value, Error>]>) -> Signal<[Value], Error>) -> SignalProducer<[Value], Error> where S.Iterator.Element: SignalProducerConvertible, S.Iterator.Element.Value == Value, S.Iterator.Element.Error == Error {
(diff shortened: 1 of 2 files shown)


leonid-s-usov:
@mdiep is this repository active? The fix to a bug has been filed a month ago, but not even a comment. Will this get merged?

mdiep:
Yes, it's active. This PR is on my list.

But I have a family and a (busy) job. I've been on vacation and traveling for work. I have other projects that I'm busy maintaining. And I'm trying to catch up on WWDC.

I'll get to this as soon as I can. :)

PR #735, opened 2019-05-07
Add possibility to use `and` + `or` with collection of signals/properties

olejnjak:
Usually on my projects I do use more properties to determine state of some `Property`, e.g. `isLoading` property usually depends on more properties which later on ends like this
```
action1.isExecuting.or(action2.isExecuting)...or(actionX.isExecuting)
```
Then I end up with extensions which simplify such code and I think it might be useful for more people than just me.

This PR adds static function `and` and `or` to `SignalProducer`, `PropertyProtocol` with appropriate attributes which solves such case elegantly.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 # master
 *Please add new entries at the top.*
-1. add possibility to use `and` and `or` operators with array of arguments (#735, kudos to @olejnjak)
+1. add possibility to use `all` and `any` operators with array of arguments (#735, kudos to @olejnjak)
    ```swift
-   let property = Property.or([boolProperty1, boolProperty2, boolProperty3])
+   let property = Property.any([boolProperty1, boolProperty2, boolProperty3])
    ```
 2. Fixed Result extensions ambiguity (#733, kudos to @nekrich)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # master
 *Please add new entries at the top.*
+1. add possibility to use `and` and `or` operators with array of arguments (#735, kudos to @olejnjak)
+   ```swift
+   let property = Property.or([boolProperty1, boolProperty2, boolProperty3])
+   ```
 
 # 6.0.0
 1. Dropped support for Swift 4.2 (Xcode 9)
--- a/Sources/FoundationExtensions.swift
+++ b/Sources/FoundationExtensions.swift
@@ -134,19 +134,3 @@ extension DispatchTimeInterval {
 		return result
 	}
 }
-
-extension Collection where Element == Bool {
-	/// Returns logical or over all elements or `false` if `self` is empty.
-	///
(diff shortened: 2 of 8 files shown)


Sources/FoundationExtensions.swift
```diff
@@ -134,3 +134,19 @@ extension DispatchTimeInterval {
...
+		return reduce(false) { $0 || $1 }
+	}
+	
+	/// Returns logical or over all elements or `false` if `self` is empty.
+	///
+	/// - returns: Logical or over all elements or `false` if `self` is empty
+	internal func and() -> Bool {
+		return reduce(true) { $0 && $1 }
+	}
+}
```
mdiep: I'd prefer not to add these extensions. They pollute the namespace and aren't used widely enough to justify that—nor do they add that much over `.reduce(true, &&)`.

Sources/Property.swift
```diff
@@ -411,6 +411,16 @@ extension PropertyProtocol where Value == Bool {
...
 		return self.lift(SignalProducer.and)(property)
 	}
+	
+	/// Create a property that computes a logical AND between the latest values of `properties`.
+	///
+	/// - parameters:
+	///   - property: Collection of properties to be combined.
+	///
+	/// - returns: A property that contains the logial AND results.
+	public static func and<P: PropertyProtocol, Properties: Collection>(_ properties: Properties) -> Property<Value> where P.Value == Value, Properties.Element == P {
```
mdiep: `all` seems like a better name for this

Sources/Property.swift
```diff
@@ -422,6 +432,16 @@ extension PropertyProtocol where Value == Bool {
...
 		return self.lift(SignalProducer.or)(property)
 	}
+	
+	/// Create a property that computes a logical OR between the latest values of `properties`.
+	///
+	/// - parameters:
+	///   - properties: Collection of properties to be combined.
+	///
+	/// - returns: A property that contains the logial OR results.
+	public static func or<P: PropertyProtocol, Properties: Collection>(_ properties: Properties) -> Property<Value> where P.Value == Value, Properties.Element == P {
```
mdiep: `any` seems like a better name for this

olejnjak:
Thanks for the review, all issues should be resolved now.

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # master
 *Please add new entries at the top.*
-1. Fixed Result extensions ambiguity (#733, kudos to @nekrich)
-
+1. add possibility to use `and` and `or` operators with array of arguments (#735, kudos to @olejnjak)
+   ```swift
+   let property = Property.or([boolProperty1, boolProperty2, boolProperty3])
+   ```
```
mdiep: This needs to be updated now

mdiep:
Looks good to me. I'll leave this open for a day or two in case anyone else has feedback.

mdiep:
Thanks! Sorry for the delay.

PR #739, opened 2019-06-03
Signal: using Signal.Event.promoteValue

NachoSoto:
This fixes compilation on `Xcode 11` / `Swift 5.1`.


Sources/Event.swift
```diff
@@ -1015,17 +1015,23 @@ extension Signal.Event where Value == Never {
...
-				case .value:
-					fatalError("Never is impossible to construct")
-				case let .failed(error):
-					action(.failed(error))
-				case .completed:
-					action(.completed)
-				case .interrupted:
-					action(.interrupted)
-				}
+                action(event.promoteValue())
```
mdiep: Indentation is off here

```suggestion
			action(event.promoteValue())
```

PR #743, opened 2019-06-19
Bump version to 6.1.0

sharplet:
#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/ReactiveSwift.podspec
+++ b/ReactiveSwift.podspec
@@ -1,7 +1,7 @@
 Pod::Spec.new do |s|
   s.name         = "ReactiveSwift"
   # Version goes here and will be used to access the git tag later on, once we have a first release.
-  s.version      = "6.0.0"
+  s.version      = "6.1.0"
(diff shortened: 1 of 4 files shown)


mdiep:
Looks good to me.

@ReactiveCocoa/reactiveswift Any objections to releasing this as 6.1.0? It's technically a breaking change because of #733, but I'm considering that a bug.

ReactiveSwift.podspec
```diff
@@ -21,5 +21,5 @@ Pod::Spec.new do |s|
   s.pod_target_xcconfig = {"OTHER_SWIFT_FLAGS[config=Release]" => "$(inherited) -suppress-warnings" }
 
   s.cocoapods_version = ">= 1.4.0"
```
ikesyo: `swift_versions` DSL is only available from CocoaPods 1.7.0, so this line should be updated to `">= 1.7.0"` as well.

PR #746, opened 2019-07-03
Rename filterMap to compactMap

Marcocanc:
I've renamed `filterMap` to `compactMap` so that it would be more in line with the Swift [compactMap](https://developer.apple.com/documentation/swift/sequence/2950916-compactmap) and the Combine [CompactMap Publisher](https://developer.apple.com/documentation/combine/publishers/compactmap).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # master
 *Please add new entries at the top.*
-1. Renamed `filterMap` to `compactMap` and deperacated `filterMap` (#746, kudos to @Marcocanc)
+1. Renamed `filterMap` to `compactMap` and deprecated `filterMap` (#746, kudos to @Marcocanc)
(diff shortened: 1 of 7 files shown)


Marcocanc:
Seems like the Travis build has gone through, but on Github it's still pending. Can anyone trigger a rebuild?

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # master
 *Please add new entries at the top.*
+1. Renamed `filterMap` to `compactMap` and deperacated `filterMap` (#746, kudos to @Marcocanc)
```
mdiep: _deprecated_ is misspelled here

Marcocanc: Oops, good catch. At least fixing that will retrigger the build :)

PR #750, opened 2019-08-07
Add @unknown default to silence a warning

ootsamo:
This fixes the following compiler warning:
```
Switch covers known cases, but 'DispatchTimeInterval' may have additional unknown values, possibly added in future versions.
```

Not 100% sure if `.infinity` and `.never` are the best values to return here, but I guess it's also unlikely we'll ever actually have new cases in the enum.


andersio:
This is actually debatable of whether we should fail gracefully rather than trap at runtime. It is arguable that failing gracefully is wrong because the system would no longer act in accordance to the actual demanded time interval as expected.

PR #758, opened 2019-09-19
Skip building testing libraries for SPM users.

sunshinejr:
Hey all! In this PR I'm removing testing libraries for SPM users. This is because currently SPM will both fetch & build your libraries even though they are in a target that you don't use. Because of that (and the current integration of SPM in Xcode), Xcode will build your testing dependencies (either in Xcode Preview or for normal builds) and complain (in Preview you won't ever get your UI to show and in your projects you will get runtime crash for missing XCTest etc.).

With this fix we are still able to test ReactiveSwift (using additional env variable) and skip downloading the testing libraries for normal usage.

Let me know what you think about it.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ aliases:
     script:
       - swift --version
       - swift build
-      - swift test
+      - TEST=1 swift test
     git:
       submodules: false
     env: JOB=SwiftPM
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # master
 *Please add new entries at the top.*
+1. Removed fetching testing libraries for SPM consumers. (#758, kudos to @sunshinejr)
 1. Renamed `filterMap` to `compactMap` and deprecated `filterMap` (#746, kudos to @Marcocanc)
 
 # 6.1.0
--- a/Package.swift
+++ b/Package.swift
@@ -1,18 +1,31 @@
 // swift-tools-version:5.0
 import PackageDescription
+import class Foundation.ProcessInfo
+
+let shouldTest = ProcessInfo.processInfo.environment["TEST"] == "1"
+
+func resolveDependencies() -> [Package.Dependency] {
+    guard shouldTest else { return [] }
+
+    return [
+        .package(url: "https://github.com/Quick/Quick.git", from: "2.0.0"),
+        .package(url: "https://github.com/Quick/Nimble.git", from: "8.0.0"),
+    ]
+}
+
+func resolveTargets() -> [Target] {
+    let baseTarget = Target.target(name: "ReactiveSwift", dependencies: [], path: "Sources")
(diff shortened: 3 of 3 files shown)


sunshinejr:
Hmm seems like only one test didn't pass on one environment (exceeding 1 sec) - is this something it was there before?

kaybutter:
A side effect introduced by this change is that you can no longer open the Package.swift in Xcode 11 and run the tests from there. 🤔

ikesyo:
> This is because currently SPM will both fetch & build your libraries even though they are in a target that you don't use.

I assume (or believe) Quick and Nimble are only built when running `swift test` since they are dependencies of `ReactiveSwiftTests` `.testTarget` and not of `ReactiveSwift` `.target`.

sunshinejr:
@ikesyo yeah that's what I assumed as well but you can try it yourself by creating a new Xcode 11 project with SwiftUI & add ReactiveSwift as an SPM package and test. First, you will see all the dependencies on the left (e.g. Quick, Nimble and their dependencies - maybe for resolution). But then you will see that Xcode Previews will try to actually build it and fail - so once you import ReactiveSwift into your project then you won't be able to use Xcode Previews.

Additionally I saw reports (and had it happened to me to) that it will actually crash on runtime with a `XCTest` linking problem in the main target.

Of course you could ignore it and rely on Xcode/SPM fixing it but I think this is worth looking into this PR as a short-term solution until its fixed. The drawback is that you will need to do one additional step in the development process.

ikesyo:
I just tried with Xcode 11 and only ReactiveSwift was built as a dependency as expected.

image

sunshinejr:
@ikesyo I was talking about Xcode Previews, these are available when you open a canvas (option+cmd+enter) in a SwiftUI-compatible view:
```swift
struct Test: View {
    var body: some View {
        Text("test")
    }
}

#if DEBUG
struct Test_Previews : PreviewProvider {
    static var previews: some View {
        Group {
            Test()
        }
    }
}
#endif
```

Zrzut ekranu 2019-09-27 o 10 47 33
(you can also use project from the FB https://github.com/feedback-assistant/reports/issues/42)

ikesyo:
I don't have Catalina environment yet so I can't examine it locally.

sunshinejr:
@ikesyo alright, no worries. We are maintaining [our own fork](https://github.com/Moya/ReactiveSwift/tree/6.1.0) for that matter at Moya for now if anyone is interested as well.

sunshinejr:
@ikesyo I just found an alternative approach to that problem - using Rocket for the release process, which can remove testing dependencies just for the release tag. Do you think we could consider this approach instead? https://github.com/RxSwiftCommunity/RxOptional/pull/83

andersio:
Closed in favor of #784

Issue #771, opened 2020-02-01
Starting a SignalProducer with debounce twice doesn't deliver values to the second observer


gpambrozio:
This is related to https://github.com/ReactiveCocoa/ReactiveSwift/issues/718.

I just found something similar on my project and I believe `debounce` to be broken when you start a producer twice. With latest master, I added the following tests to `SignalProducerSpec.swift` and they both fail (just the second expectation of each `it`, not the first):

```
describe("debounce") {
	context("property - starting twice") {
		it("should deliver the same values") {
			let p1 = MutableProperty(1)
			let p2 = MutableProperty(2)

			let scheduler = TestScheduler()
			let combined = Property.combineLatest(p1, p2)
			let prod = combined
				.producer
				.skipRepeats { $0 == $1 }
				.map { $0 + $1 }
				.debounce(0.1, on: scheduler)

			var v1 = [Int]()
			var v2 = [Int]()
			prod.startWithValues { (v) in
				v1.append(v)
			}
			prod.startWithValues { (v) in
				v2.append(v)
			}

			scheduler.advance(by: .milliseconds(200))
			p1.value = 3
			scheduler.advance(by: .milliseconds(200))
			p2.value = 4
			scheduler.advance(by: .milliseconds(200))
			p2.value = 4
			scheduler.advance(by: .milliseconds(200))

			expect(v1).to(equal([3, 5, 7]))
			expect(v2).to(equal([3, 5, 7]))
		}
	}

	context("pipe - starting twice") {
		it("should deliver the same values") {
			let p1 = SignalProducer<Int, Never>.pipe()
			let p2 = SignalProducer<Int, Never>.pipe()

			let scheduler = TestScheduler()
			let combined = SignalProducer.combineLatest(p1.0, p2.0)
			let prod = combined
				.skipRepeats { $0 == $1 }
				.map { $0 + $1 }
				.debounce(0.1, on: scheduler)

			var v1 = [Int]()
			var v2 = [Int]()
			prod.startWithValues { (v) in
				v1.append(v)
			}
			prod.startWithValues { (v) in
				v2.append(v)
			}

			p1.1.send(value: 1)
			p2.1.send(value: 2)
			scheduler.advance(by: .milliseconds(200))
			p1.1.send(value: 3)
			scheduler.advance(by: .milliseconds(200))
			p2.1.send(value: 4)
			scheduler.advance(by: .milliseconds(200))
			p2.1.send(value: 4)
[...]

PR #772, opened 2020-02-02
Fix debounce when started twice

gpambrozio:
Fixes issue #771. Thanks @andersio for the quick solution on this.

I also added tests that show the solution works. The tests were added at a commit before the change that fixed it so it's easy to verify that tests fail before the change but not after.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # master
 *Please add new entries at the top.*
 
+1. Fixed `SignalProducer.debounce` operator that, when started more than once, would not deliver values on producers started after the first time. (#772, kudos to @gpambrozio)
 1. `FlattenStrategy.throttle` is introduced. (#713, kudos to @inamiy)
 1. Updated `README.md` to reflect Swift 5.1 compatibility and point snippets to 6.1.0 (#763, kudos to @Marcocanc)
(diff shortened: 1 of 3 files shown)


gpambrozio:
@andersio not sure why this test failed ci:
```
/Users/travis/build/ReactiveCocoa/ReactiveSwift/Tests/ReactiveSwiftTests/SignalProducerSpec.swift:1060: error: -[ReactiveSwiftTests.SignalProducerSpec zip, can deal with hundreds of producers] : failed - Waited more than 1.0 second
```

It seems go only fail sometimes when I run the tests locally, might be indicative of a performance issue or just poor cpu management at the time it ran.

Adding a `(timeout: 2)` to that test makes it succeed all the time for me, even when my machine is building another project at the same time but I don't want to push this "fix" without getting some thoughts on it first.

andersio:
@gpambrozio It is general flakiness of asynchronous test IMO. It can be rewritten it in terms of `TestScheduler` at a later point.

gpambrozio:
@andersio do you want me to add the `timeout` change to this PR?

andersio:
@gpambrozio It wouldn't be necessary. 👍

PR #773, opened 2020-02-13
Increased performance of joining signals by a factor of around 5

rocketnik:
this becomes relevant, when joining 1000 signals and more

#### Checklist
✅ Updated CHANGELOG.md.

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -1680,7 +1680,7 @@ extension Signal {
 					return true
 				}
 
-				_haveAllSentInitial = values.reduce(true) { $0 && !($1 is Placeholder) }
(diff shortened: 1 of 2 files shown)


Sources/Signal.swift
```diff
@@ -1680,7 +1680,7 @@ extension Signal {
 					return true
 				}
 
-				_haveAllSentInitial = values.reduce(true) { $0 && !($1 is Placeholder) }
+				_haveAllSentInitial = values.allSatisfy{ !($0 is Placeholder) }
```
andersio:
```suggestion
				_haveAllSentInitial = values.allSatisfy { !($0 is Placeholder) }
```

PR #774, opened 2020-02-13
Joining empty list of producers fix

rocketnik:
#### Checklist
✅ Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # master
 *Please add new entries at the top.*
 
+1. Joining an empty list of producers results in sending an event on the joined signal which contains the empty list, before the signal is completed. This becomes relevant, when the list of producers is calculated from some other Signal and the signal resulting from the joined producers is observed. If no event is sent only when the producers list is empty, then the observer gets staleld and e.g. the ui won't update.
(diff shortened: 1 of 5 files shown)


andersio:
Both `combineLatest` and `zip` have been designed to consistently emit _only_ array of N elements, given a sequence of N producers, and are documented to return an empty producer if the sequence of producer is empty. So we cannot change the default behavior silently here, though I do see the ergonomic value here.

The best option, IMO, would be to make it an opt-in option, e.g. `noUpstreamSentinel: [S.Iterator.Element.Value]?` on `combineLatest` and `zip` that accepts a `Sequence`.

rocketnik:
@andersio please check again.

I agree with not breaking old code and went with your idea.

For private functions, I made this parameter mandatory, while for public ones I gave it a default value of `nil` resulting in the exact same behavior as before.

Please notice the implementation of `func all` and `func and`, for which I have provided a sentinel (`[]`). It looked like the old implementation was expecting an empty list to be returned already, but actually, this could not have happened.

rocketnik:
Resolved conflict. @andersio Please recheck, when you find the time.

rocketnik:
Any update on this?

rocketnik:
Any update on this?

PR #783, opened 2020-05-22
Rename `FlattenStrategy.throttle` to `dropOnBackpressure`.

andersio:
Rename `FlattenStrategy.throttle` to a more common term of art ["drop on backpressure"](https://github.com/ReactiveX/RxJava/wiki/Backpressure-(2.0)#onbackpressuredrop).

#### Checklist
- [x] Updated CHANGELOG.md.


andersio:
@sergdort @mluisbrown @Marcocanc Thoughts?

mluisbrown:
This might mislead people to believe that RAS supports backpressure like Combine, which it doesn’t.

andersio:
Hmm, it is valid to keep a clear distinction from reactive pull support. Though I am more leaning towards that _backpressure_ is a general scenario that can emerge regardless of the operating model.

mluisbrown:
What’s the motivation to change the name? What’s wrong with .throttle?

nkristek:
In my experience, `throttle` is often mixed up with `debounce` and I see why this naming makes the intention of the operator more clear.

mluisbrown:
> In my experience, throttle is often mixed up with debounce

That is true. However that is in relation to the `throttle(_, on:)` and `debounce(_, on:)` operators, not flattening strategies. But that actually provides a good argument IMO to change the `.throttle` flatten strategy name, to make it distinct from the operator.

Having now seen the RxJava [link](https://github.com/ReactiveX/RxJava/wiki/Backpressure-(2.0)#onbackpressuredrop) that @andersio provided, which provides prior art, I'm inclined to support this name change 👍

Sources/Deprecations+Removals.swift
```diff
@@ -1,6 +1,12 @@
 import Foundation
 import Dispatch
 
+extension FlattenStrategy {
+	/// Renamed to `dropOnBackpressure`.
+	@available(*, deprecated, renamed:"dropOnBackpressure")
```
Marcocanc, line 6: Not sure how strict we are on space after colon, but I just noticed this one.

andersio: We've been ad hocly following Swift stdlib style, and there is no space after colon in this context.

andersio:
@mluisbrown 

It was originally proposed as `flatMap(.first)`, but there are concerns about the room of misinterpretation of `first` with `race` and similarly named operators, and not reflecting the "start if no inner is running" aspect of it. So `throttle` was then chosen, as it resembles the semantics of `flatMap(.first)` but in the time domain.

It is only then, when I research for backpressure support in other FRP implementations, I encountered `onBackpressureDrop` as a prior art. IMO it is especially worth adopting because relating to the concept of backpressure is perhaps easier than trying to project `throttle` — an established time domain operation — into another domain.

Sources/Flatten.swift
```diff
@@ -12,7 +12,7 @@ public struct FlattenStrategy {
 		case concurrent(limit: UInt)
 		case latest
 		case race
-		case throttle
+		case dropOnBackpressure
```
sergdort, line 15: `.first` 🙈

mluisbrown: giphy

Marcocanc: I prefer `.first` as well

PR #794, opened 2020-06-10
Stop declaring Swift 5.0 support.

andersio:
We have adopted `@propertyWrapper` which requires Swift 5.1+.

#### Checklist
- ~~Updated CHANGELOG.md.~~


mluisbrown:
Doesn't it require Swift 5.2? In Package.swift you have `// swift-tools-version:5.2` which forces the use of Swift 5.2. I'm currently having that issue in the ReactiveSwift TCA, so I have to declare only Swift 5.2 support (and therefore only Xcode 11.4 and 11.5 support).

If 5.2 is not required, it would be great to change the Package declaration to reflect this too 🙏

andersio:
@mluisbrown It was bumped to `swift-tools-version:5.2` because 5.2 contains SwiftPM fixes around test dependencies. #784

mluisbrown:
ok, so the SPM version has to remain 5.2 😢  Fair enough.

PR #797, opened 2020-08-15
Unserialized and Reentrant `Signal`s

andersio:
1. `Signal` now offers an unserialized variant and a reentrant variant for advanced users. (#797)

   These new primitives would be adopted by ReactiveCocoa UIKit bindings to improve interoperability with RAC Loop, to tackle some legitimate recursive delivery scenarios (e.g. around first responder management), and also to reduce fine-grained locking across ReactiveSwift and ReactiveCocoa.

   Note that the default behavior of `Signal` has not been changed — event serialization remains the default behavior. `MutableProperty` is also not gonna be reentrant, because there is no sane reentrancy model that supports access to the current value and multicasting [while being compliant to the Property contract](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Documentation/APIContracts.md#events-must-be-synchronously-emitted-after-the-mutation-is-visible).

1. Fine-grained locking has been scaled down across Signal, SignalProducer and Property. (#797)

    [Observers are naturally inheriting mutual exclusion from upstream, because the upstream has to serialize the events to be compliant to the Event contract](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Documentation/APIContracts.md#events-are-serial). So for most transformations involving only one event stream (Signal/Producer), the locks for serializing events are redundant. This is also the same premise that RAS 2.0 producer optimizations are built upon.

    This similarly applies to Properties — both the read-only and the mutable versions have their own locks for mutual exclusion. So for them, the only value which the default event serialization in `Signal`/`SignalProducer` brings is wasting CPU time.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -80,6 +80,8 @@ public final class Signal<Value, Error: Swift.Error> {
 			generator(Observer(action: self.send, interruptsOnDeinit: true), Lifetime(disposable))
 		}
 
+		@_specialize(kind: partial, where SendLock == Lock)
+		@_specialize(kind: partial, where SendLock == NoLock)
 		private func send(_ event: Event) {
 			if event.isTerminating {
 				// Recursive events are disallowed for `value` events, but are permitted
--- a/Sources/Signal.swift
+++ b/Sources/Signal.swift
@@ -43,9 +43,14 @@ public final class Signal<Value, Error: Swift.Error> {
 	///            | Other observers |                                 | Other observers |
 	///            -------------------                                 -------------------
 	/// ```
-	private let core: Core
+	private let core: CoreBase
 
-	private final class Core {
+    private class CoreBase {
+        func observe(_ observer: Observer) -> Disposable? { fatalError() }
+        func signalDidDeinitialize() { fatalError() }
+    }
+
+    private final class Core<SendLock: LockProtocol>: CoreBase {
 		/// The disposable associated with the signal.
 		///
 		/// Disposing of `disposable` is assumed to remove the generator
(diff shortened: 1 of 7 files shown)


Sources/Signal.swift
```diff
@@ -285,8 +295,61 @@ public final class Signal<Value, Error: Swift.Error> {
...
+	/// - warning: The input observer **is not thread safe**.
+	///
+	/// - note: The disposable returned from the closure will be automatically
+	///         disposed if a terminating event is sent to the observer. The
+	///         Signal itself will remain alive until the observer is released.
+	///
+	/// - parameters:
+	///   - generator: A closure that accepts an implicitly created observer
+	///                that will act as an event emitter for the signal.
+	public static func nonSerializing(_ generator: (Observer, Lifetime) -> Void) -> Signal {
```
sharplet, line 336: "Non-serialising" reads a bit awkwardly to me. How about "unserialized"? It also shares a prefix with "unsafe", which feels at least helpfully coincidental.

Sources/Signal.swift
```diff
@@ -295,8 +358,28 @@ public final class Signal<Value, Error: Swift.Error> {
 	/// - parameters:
 	///   - generator: A closure that accepts an implicitly created observer
 	///                that will act as an event emitter for the signal.
-	public init(_ generator: (Observer, Lifetime) -> Void) {
-		core = Core(generator)
+	public static func reentrant(_ generator: (Observer, Lifetime) -> Void) -> Signal {
```
sharplet, line 361: Should this also use "unserialized"? I wonder if it would be valuable to add the terms "reentrant" and "unserialized" to a glossary somewhere to make it clear when each is used and why?

Tests/ReactiveSwiftTests/SignalSpec.swift
```diff
@@ -200,6 +200,51 @@ class SignalSpec: QuickSpec {
 			}
 		}
 
+		describe("reentrant") {
+			#if arch(x86_64) && canImport(Darwin)
```
sharplet, line 204: What's the reason for restricting this text to x86?

andersio: Catching deadlock by intercepting signals is supported only in iOS simulators.

> https://github.com/mattgallagher/CwlPreconditionTesting#cwlpreconditiontesting
>
> A Mach exception handler, written in Swift and Objective-C, that allows EXC_BAD_INSTRUCTION (as raised by Swift's assertionFailure/preconditionFailure/fatalError) to be caught and tested.
>
> NOTE: the iOS code runs in the simulator only. It is for logic testing and cannot be deployed to the device due to the Mach exception API being private on iOS.

dnadoba: What about Macs with an M1?

andersio: It is more a question for Nimble/CwlPreconditionSignal. If they support catching signals on M1, we can lift this.

mluisbrown:
> to tackle some legitimate recursive delivery scenarios (e.g. around first responder management)

@andersio any update on when this might get merged and into a release? It would be extremely useful for [RAS-TCA](https://github.com/trading-point/reactiveswift-composable-architecture), for the same reasons as it would be useful in Loop.

maximkrouk:
[+1](https://github.com/ReactiveCocoa/ReactiveSwift/pull/797#issuecomment-826761568)

mluisbrown:
Any update on when this might get merged @andersio ?

PR #801, opened 2020-08-29
Add variadic sugar for boolean static methods.

fortmarek:
Hi 👋 

This is my first contribution, albeit a small one.

What I'd like to add is variadic argument support for static intializers of `Signal`s, `SignalProducer`s and `Properties` where `Value == Boolean`.
So, similar to how we write `SignalProducer.combineLatest(producer1, producer2)` we could write `SignalProducer.any(producer1, producer2)` - right now we have to do `SignalProducer.any([producer1, producer2])` which seems unnecessarily verbose.

I have also change a few test cases to leverage this new feature

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # master
 *Please add new entries at the top.*
 
+1. Add variadic sugar for boolean static methods such as `Property.any(boolProperty1, boolProperty2, boolProperty3)` (#801, kudos to @fortmarek)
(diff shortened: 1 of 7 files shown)


RuiAAPeres:
I am slightly on the fence about this introduction, since the implementation, as-is is quite shallow. I will defer this to @andersio.

andersio (approved):

andersio:
@RuiAAPeres I think it's alright, basically sugar over the existing variant taking a `Collection`. Doesn't introduce new logic, and doesn't introduce overload conflicts.

PR #802, opened 2020-09-14
bump package target to iOS 9 and update testing packages

petrpavlik:
When `ReactiveSwift` is imported as a `swift package` using Xcode 12, it generates following warning `deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 14.0.99.` This PR fixes the warning by bumping the min. deployment target to iOS 9 when importing this dependency as a swift packages

I've also updated `Quick` and `Nimble` dependencies to their latest versions. The `ReactiveCocoa` counterpart PR is here https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3712.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,13 @@
 # master
 *Please add new entries at the top.*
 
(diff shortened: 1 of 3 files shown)


NachoSoto (approved):

RuiAAPeres:
Ping @andersio. ❤️

PR #808, opened 2020-10-27
Fix a `SignalProducer.lift` issue which may leak intermediate signals. 

andersio:
`SignalProducer.lift` might leak intermediate signals, if the upstream has been terminated, way before a downstream interruption is raised.

#### Checklist
- [x] Updated CHANGELOG.md.


Marcocanc (approved):
Good catch!

PR #810, opened 2020-12-21
Adds the `interval` operator.

mluisbrown:
Adds the `interval` operator with functionality similar to that of the operator with the same name in [ReactiveX](http://reactivex.io/RxJava/javadoc/rx/Observable.html#interval-long-java.util.concurrent.TimeUnit-rx.Scheduler-).

Allows specifying any `Sequence`, including infinite sequences such as `0...` (which is the sequence used if none is provided, to mimic the ReactiveX behaviour).

This can be useful, for example, to make a "typewriter" effect of characters appearing on the screen at regularly spaced intervals.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/Tests/ReactiveSwiftTests/SignalProducerSpec.swift
+++ b/Tests/ReactiveSwiftTests/SignalProducerSpec.swift
@@ -1257,6 +1257,70 @@ class SignalProducerSpec: QuickSpec {
 			}
 		}
 
+		describe("interval") {
+			it("should send the next sequence value at the given interval") {
+				let scheduler = TestScheduler()
+				let producer = SignalProducer.interval("abc", interval: .seconds(1), on: scheduler)
+
+				var isDisposed = false
+				var values: [Character] = []
+				producer
+					.on(disposed: { isDisposed = true })
+					.startWithValues { values.append($0) }
+
+				scheduler.advance(by: .milliseconds(900))
+				expect(values) == []
+
+				scheduler.advance(by: .seconds(1))
+				expect(values) == ["a"]
+
+				scheduler.advance()
+				expect(values) == ["a"]
+
+				scheduler.advance(by: .milliseconds(200))
+				expect(values) == ["a", "b"]
+
+				scheduler.advance(by: .seconds(1))
+				expect(values) == ["a", "b", "c"]
+
+				scheduler.advance(by: .seconds(1))
+				expect(isDisposed) == true
+			}
+
+			it("shouldn't overflow on a real scheduler") {
+				let scheduler = QueueScheduler.makeForTesting()
+				let producer = SignalProducer.interval("abc", interval: .seconds(3), on: scheduler)
+				producer
+					.start()
+					.dispose()
+			}
+
+			it("should dispose of the signal when disposed") {
+				let scheduler = TestScheduler()
+				let producer = SignalProducer.interval("abc", interval: .seconds(1), on: scheduler)
+				var interrupted = false
+
+				var isDisposed = false
+				weak var weakSignal: Signal<Character, Never>?
+				producer.startWithSignal { signal, disposable in
+					weakSignal = signal
+					scheduler.schedule {
+						disposable.dispose()
+					}
+					signal.on(disposed: { isDisposed = true }).observeInterrupted { interrupted = true }
+				}
+
+				expect(weakSignal).to(beNil())
+				expect(isDisposed) == false
+				expect(interrupted) == false
+
+				scheduler.run()
+				expect(weakSignal).to(beNil())
+				expect(isDisposed) == true
+				expect(interrupted) == true
+			}
+		}
+
 		describe("throttle while") {
 			var scheduler: ImmediateScheduler!
 			var shouldThrottle: MutableProperty<Bool>!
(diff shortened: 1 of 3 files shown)


andersio:
We have a more generalized `collect(every:)` operator, except for the sequential number generating part. Could you share your use case if you deemed it can only be achieved by `interval()`?

mluisbrown:
The `collect(every:on:)` operator is not quite the same thing, and I'm not sure it can be used to replicate what can be achieved using `interval`.  For example:

```swift
SignalProducer("abcde")
    .collect(every: .milliseconds(250), on: QueueScheduler.main, discardWhenCompleted: false)
    .startWithValues {
        print("\(Date()): \($0)")
}
```

will print `2021-01-04 12:13:03 +0000: ["a", "b", "c", "d", "e"]`, ie all the values a collected and emitted at the same time.

Whereas:

```swift
SignalProducer.interval("abcde", interval: .milliseconds(250), on: QueueScheduler.main)
    .startWithValues {
        print("\(Date()): \($0)")
}
```

will print each value in the sequence at 250ms intervals. One use case for this is for animating the appearance of text in the UI, as in this example:


https://user-images.githubusercontent.com/4175766/103536065-bf2fd080-4e89-11eb-9e73-8fa1a803081b.mp4

mluisbrown:
The alternative to using `interval` for the use case above is something like this:

```swift
SignalProducer(title)
    .flatMap(.concat) { char in
        SignalProducer(value: char)
            .delay(0.05, on: QueueScheduler.main)
    }
```

mluisbrown:
Other than that, `interval` exists as an operator in other FRP libraries so it would be nice to have it here also.

andersio:
@mluisbrown 

Seems like I might have misunderstood what `interval()` does in Rx.

Judging by your description, it seems `SignalProducer.Type.timer` should satisfy your need.

```swift
SignalProducer
    .timer(.milliseconds(5), on: QueueScheduler.main)
    .map { date in (date, "abcde") }
```

mluisbrown:
@andersio 

>  it seems SignalProducer.Type.timer should satisfy your need

Wouldn't the example you provided just emit the value `"abcde"` every 50 milliseconds? What I'm trying to do is emit each element of the sequence at the defined interval, until it ends (or for ever, in the case of an infinite sequence).

| time | value |
-------|-------
| 0.050 |  "a" |
| 0.100 |  "b" |
| 0.150 |  "c" |
| 0.200 |  "d" |
| 0.250 |  "e" |

FWIW: I tried creating this using `timer` and was unable to, ending up with the solution using `flatMap` and `delay` until I created a more generic `interval` solution.

andersio:
Argh, missed the part of emitting elements in the collection one by one. Somehow thought it was repeating the string literal regularly.

I think `zip` would help in this use case?

```
SignalProducer.zip(
    SignalProducer.timer(.milliseconds(5), on: QueueScheduler.main),
    SignalProducer("Hello world!") // `SignalProducer<Character, Never>
)
.scan(into: "") { $0.append($1) }
```

Should work alright for small arrays/sequences.

mluisbrown:
The `zip` with `timer` and `scan` approach would work. So would the `flatMap` and `delay` approach I mentioned above. There are probably a few other approaches that would work. 

The idea of adding the `interval` operator though is to have an efficient solution that can work for any sequence of any length, including lazy sequences. 

Given there is prior art for this in Rx, what is the objection to adding it?

Tests/ReactiveSwiftTests/SignalProducerSpec.swift
```diff
@@ -1257,6 +1257,70 @@ class SignalProducerSpec: QuickSpec {
...
+				scheduler.advance(by: .seconds(1))
+				expect(values) == ["a", "b", "c"]
+
+				scheduler.advance(by: .seconds(1))
+				expect(isDisposed) == true
+			}
+
+			it("shouldn't overflow on a real scheduler") {
+				let scheduler = QueueScheduler.makeForTesting()
+				let producer = SignalProducer.interval("abc", interval: .seconds(3), on: scheduler)
```
andersio, line 1292: Maybe provide an astronomically large sequence with `repeatElement`?

andersio:
@mluisbrown There is no objection. =)

Tests/ReactiveSwiftTests/SignalProducerSpec.swift
mluisbrown: Done 👍

mluisbrown:
> other than RxSwift parity (which has never been a USP of RAS). :)

😄 I just meant that the fact that it already exists in ReactiveX is one more factor in evaluating the usefulness.

FWIW [`interval`](https://github.com/ReactiveX/RxSwift/blob/0efa6d1482ddaea12d63f4f17567daa4744c7420/RxSwift/Observables/Timer.swift#L19) in RxSwift is not really the same thing, and is basically just a version of `timer`. 

The prior art I was referring to was from [RxJava](http://reactivex.io/RxJava/javadoc/rx/Observable.html#interval-long-java.util.concurrent.TimeUnit-rx.Scheduler-)

andersio (approved):

RuiAAPeres:
Nice. 🤠

PR #819, opened 2021-02-24
Fixed a deprecation warning where some protocols were inheriting from class instead of AnyObject per Xcode's suggestion

apps4everyone:
This is only to remove five compiler warnings.

I was updating my iOS app to be able to run on Xcode 12.5 (beta) and noticed that Apple is now flagging the use of class inherited protocols as deprecated in favor of inheriting from AnyObject. In my iOS app, we flag warnings as errors so now our app won't compile. I wanted to be able to pitch in and help modernize the code so I went ahead and fixed this deprecation here.

After a little bit of research, I came across this forum Swift.org [post](https://forums.swift.org/t/class-only-protocols-class-vs-anyobject/11507/4) which seems to point to the discussion where this deprecation discussion started.

### DIFF
--- a/Sources/Disposable.swift
+++ b/Sources/Disposable.swift
@@ -8,7 +8,7 @@
 
 /// Represents something that can be “disposed”, usually associated with freeing
 /// resources or canceling work.
-public protocol Disposable: class {
+public protocol Disposable: AnyObject {
 	/// Whether this disposable has been disposed already.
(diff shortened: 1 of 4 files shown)


Marcocanc (approved):

apps4everyone:
@andersio would be a 6.5.1 release possible? thx

PR #828, opened 2021-04-24
Make advance by TimeInterval in TestScheduler possible

carsten-wenderdel:
The `currentDate` in `TestScheduler` could already be advanced by
`DispatchTimeInterval`. In unit tests the developer might want to
reuse an existing constant, which often is a `Double`/`TimeInterval`.

This PR adds a second method with exactly the same signature as the
existing one; only the type of the argument is different.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # master
 
+1. `TestScheduler` can `advance` by `TimeInterval`. (#828)
+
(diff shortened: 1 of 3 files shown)


andersio (approved):

mluisbrown (approved):
Nice. This would remove some of the frustrations in migrating TCA code to [RAS-TCA](https://github.com/trading-point/reactiveswift-composable-architecture), involving converting all TimeIntervals to DispatchTimeIntervals.

PR #826, opened 2021-03-24
Update Carthage xcconfig to 1.1

MikeChugunov:
resolves #825 

#### Checklist
- [x] Updated CHANGELOG.md.

PR #832, opened 2021-06-18
Use official xcconfigs repository

iby:
This is a clean up PR for #826.

#### Checklist
- [x] Updated CHANGELOG.md.


iby:
Looks like this is all good? Can we roll this out in 6.6.1? Then I could include it in ReactiveCocoa/ReactiveCocoa#3728 – I'm guessing the build there is failing because of missing Apple Silicon support in the current Carthage release. Can I be of help with release notes or anything else?

PR #834, opened 2021-07-14
UnidirectionalBinding should return non optionals

NicholasTD07:
This fixes the `<~` operator returning optional values even though it should be able to just return non optional ones.

#### Checklist
- [x] Updated CHANGELOG.md.


RuiAAPeres:
This is a breaking change and likely need a major release. @andersio thoughts?

andersio (approved):

PR #841, opened 2021-10-05
Signal.Event: changed to `@frozen`

NachoSoto:
This `enum` isn't expected to change. It's important to mark it as `@frozen`
so that clients don't get warnings whem compiling with `BUILD_LIBRARY_FOR_DISTRIBUTION`,
which might be required in a future version of Carthage supporting Catalyst.
(See https://github.com/Carthage/Carthage/pull/3235).

#### Checklist
- [ ] Updated CHANGELOG.md.

### DIFF
--- a/Sources/Event.swift
+++ b/Sources/Event.swift
@@ -6,6 +6,7 @@ extension Signal {
 	///
 	/// Signals must conform to the grammar:
 	/// `value* (failed | completed | interrupted)?`
+	@frozen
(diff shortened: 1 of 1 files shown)


andersio (approved):

andersio:
It looks like by Library Evolution rules, adding or removing `@frozen` is always ABI breaking. So this would require a major version bump.

PR #847, opened 2021-11-12
Add `asyncStream` and `asyncThrowingStream` for `Signal` and `SignalProducer`

Marcocanc:
This is a proposal to add asyncStream and asyncThrowingStream in order to bridge into Swift Concurrency.
I find this especially useful for writing non-blocking async tests.

A couple of things I'd love to hear your thoughts on (apart from the usual review):

General thoughts on the idea
Where to place these extensions within the project and naming of files
Raising of CI platform to Xcode 11 (due to minimum Swift and testing target version requirements)
Lack of support for Swift Concurrency in Quick/Nimble (leading to tests written with XCTest)

### DIFF
--- a/Sources/Signal+SwiftConcurrency.swift
+++ b/Sources/Signal+SwiftConcurrency.swift
@@ -1,0 +1,52 @@
+//
+//  Signal+SwiftConcurrency.swift
+//  ReactiveSwift
+//
+//  Created by Marco Cancellieri on 2021-11-11.
+//  Copyright (c) 2021 GitHub. All rights reserved.
+//
+#if compiler(>=5.5) && canImport(_Concurrency)
+import Foundation
+
+@available(macOS 12, iOS 15, watchOS 8, tvOS 15, *)
+extension Signal {
+    public var asyncThrowingStream: AsyncThrowingStream<Value, Swift.Error> {
+        AsyncThrowingStream<Value, Swift.Error> { continuation in
+            let disposable = observe { event in
+                switch event {
+                case .value(let value):
+                    continuation.yield(value)
+                case .completed, .interrupted:
+                    continuation.finish()
+                case .failed(let error):
+                    continuation.finish(throwing: error)
+                }
+            }
+            continuation.onTermination = { @Sendable termination in
+                disposable?.dispose()
+            }
+        }
+    }
+}
+
+extension Signal where Error == Never {
+    @available(macOS 12, iOS 15, watchOS 8, tvOS 15, *)
+    public var asyncStream: AsyncStream<Value> {
+        AsyncStream<Value> { continuation in
+            let disposable = observe { event in
+                switch event {
+                case .value(let value):
+                    continuation.yield(value)
+                case .completed, .interrupted:
+                    continuation.finish()
+                case .failed:
+                    fatalError("Never is impossible to construct")
+                }
+            }
+            continuation.onTermination = { @Sendable termination in
+                disposable?.dispose()
+            }
+        }
+    }
+}
+#endif
--- a/Sources/SignalProducer+SwiftConcurrency.swift
+++ b/Sources/SignalProducer+SwiftConcurrency.swift
@@ -1,0 +1,52 @@
+//
+//  SignalProducer+SwiftConcurrency.swift
+//  ReactiveSwift
+//
+//  Created by Marco Cancellieri on 2021-11-11.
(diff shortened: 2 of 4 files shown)


Sources/Signal+SwiftConcurrency.swift
```diff
@@ -0,0 +1,52 @@
...
+//  Signal+SwiftConcurrency.swift
+//  ReactiveSwift
+//
+//  Created by Marco Cancellieri on 2021-11-11.
+//  Copyright (c) 2021 GitHub. All rights reserved.
+//
+#if compiler(>=5.5) && canImport(_Concurrency)
+import Foundation
+
+@available(macOS 12, iOS 15, watchOS 8, tvOS 15, *)
```
mluisbrown, line 11: With Xcode 13.2 (currently in beta), Swift Concurrency has been back ported to iOS 13 and macOS 10.15 so if we wait for that we might be able to bring this availability check down:

```
#if compiler(>=5.5.2) && canImport(_Concurrency)
@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
```

Sources/SignalProducer+SwiftConcurrency.swift
```diff
@@ -0,0 +1,52 @@
+//
+//  SignalProducer+SwiftConcurrency.swift
+//  ReactiveSwift
+//
+//  Created by Marco Cancellieri on 2021-11-11.
+//  Copyright (c) 2021 GitHub. All rights reserved.
+//
+#if compiler(>=5.5) && canImport(_Concurrency)
```
mluisbrown, line 8: See above comment.

mluisbrown (commented):
Looks awesome 👏

Sources/Signal+SwiftConcurrency.swift
Marcocanc: Good point, might be worth the wait!

Tests/ReactiveSwiftTests/SwiftConcurrencyTests.swift
```diff
@@ -109,7 +109,7 @@ class SwiftConcurrencyTests: XCTestCase {
 	}
 }
 // Extension to allow Throw assertion for async expressions
-@available(macOS 12, iOS 15, watchOS 8, tvOS 15, *)
+@available(macOS 12, iOS 15, watchOS 8, tvOS 15, macCatalyst 15, *)
```
mluisbrown, line 112: Doesn't this need to be added in the Signal and SP extensions also?

Marcocanc: It's everywhere now, but it somehow still doesn't want to build

Marcocanc:
Interestingly the macCatalyst issue is known in Xcode 12.2 Beta
https://developer.apple.com/documentation/xcode-release-notes/xcode-13_2-release-notes
There's a workaround, but probably worth waiting rather than modifying the SDK in the CI.

Sources/Signal+SwiftConcurrency.swift
```diff
@@ -0,0 +1,52 @@
...
+//
+//  Created by Marco Cancellieri on 2021-11-11.
+//  Copyright (c) 2021 GitHub. All rights reserved.
+//
+#if compiler(>=5.5) && canImport(_Concurrency)
+import Foundation
+
+@available(macOS 10.15, iOS 13, watchOS 6, tvOS 13, macCatalyst 13, *)
+extension Signal {
+	public var asyncThrowingStream: AsyncThrowingStream<Value, Swift.Error> {
```
NachoSoto, line 13: I think we should use `AsyncStream<Result<Value, Error>>` to maintain the error type.

Sources/Signal+SwiftConcurrency.swift
```diff
@@ -0,0 +1,52 @@
...
+		AsyncThrowingStream<Value, Swift.Error> { continuation in
+			let disposable = observe { event in
+				switch event {
+				case .value(let value):
+					continuation.yield(value)
+				case .completed, .interrupted:
```
gshahbazian, line 19: Should the `interrupted` event throw a swift `CancellationError`?

NachoSoto: Good point

Sources/Signal+SwiftConcurrency.swift
NachoSoto: Bump

PR #855, opened 2022-10-03
Add primary associated types to SignalProducerConvertible & SignalProducerProtocol

braker1nine:
Adds [primary associated types](https://github.com/apple/swift-evolution/blob/main/proposals/0346-light-weight-same-type-syntax.md) to `SignalProducerConvertible` and `SignalProducerProtocol`

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # master
 *Please add new entries at the top.*
 
+1. Add primary associated types to SignalProducerConvertible & SignalProducerProtocol (#855, kudos to @braker1nine)
 1. Refactor Github Actions to cover more swift versions (#858, kudos to @braker1nine)
 1.Use `OSAllocatedUnfairLock` instead of `os_unfair_lock` on supported Apple platforms (#856, kudos to @mluisbrown)
 
--- a/Sources/SignalProducer.swift
+++ b/Sources/SignalProducer.swift
@@ -538,6 +538,7 @@ extension SignalProducer where Error == Swift.Error {
 	}
 }
 
+#if swift(>=5.7)
 /// Represents reactive primitives that can be represented by `SignalProducer`.
 public protocol SignalProducerConvertible<Value, Error> {
 	/// The type of values being sent by `self`.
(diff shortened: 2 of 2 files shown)


NachoSoto (approved):

mluisbrown (approved):

NachoSoto:
This breaks compilation with older compiler versions though.

braker1nine:
Ahhh, I was afraid of that. Let me see if I can handle that better 🤔

braker1nine:
Made a first pass at this. I'm not going to lie, I don't _love_ it. Requires a bunch of repeated code...

Looks like the original proposal had the concept of a [`primary` keyword](https://github.com/apple/swift-evolution/blob/main/proposals/0346-light-weight-same-type-syntax.md#annotate-regular-associatedtype-declarations-with-primary), which would have made this much nicer. Unfortunately, from what I can tell that didn't make it into the final accepted version. 😞

NachoSoto:
I think this could work. We need to make sure CI tests run on Xcode 14 to test this too.

braker1nine:
Tweaked the test.yml to run on both macOS-12 and macOS-11. My GitHub Actions knowledge is a little limited so let me know if I did anything dumb.

braker1nine:
Just following up on this. @NachoSoto @mluisbrown is there anything you need me to do to help get this merged

mluisbrown:
@braker1nine we should try and have CI passing. There are issues with macCatalyst but the rest should pass ok.

braker1nine:
Oh sheesh. I missed that. Sorry @mluisbrown!

braker1nine:
I made a bit of a tweak to the testing structure on this one to try to increase the coverage of Swift versions

- Testing on Xcode with both 13.4.1 and 14.1, so we hit both Swift 5.6 and 5.7
- Testing Swift 5.2 (the minimum supported version for ReactiveSwift I think?), 5.6, and 5.7 on Linux SwiftPM. Also changed that one to use the setup-swift action instead of pulling it in via the swift version manager script

NachoSoto:
What do you think of opening a separate PR with the CI changes? We can merge that right away then I'll take a look at this one 🙏

mluisbrown:
> What do you think of opening a separate PR with the CI changes? We can merge that right away then I'll take a look at this one 🙏

@NachoSoto we also need to address the fact that Mac Catalyst CI jobs are simply not working. IMO we should just disable the Mac Catalyst job until we have a better idea of what the issue is, as it's going to cause issues with all PRs. 

What happens is that the job starts and then just hangs for ever until GitHub Actions decides to cancel it (hours later). If I run the exact same `xcodebuild` command locally it runs just fine, so there is no clue as to what the problem is. 

The issue is affecting this PR and also [this one](https://github.com/ReactiveCocoa/ReactiveSwift/pull/857) and affected [this one](https://github.com/ReactiveCocoa/ReactiveSwift/pull/856) too. @p4checo tried using Xcode 14.1 in his PR to see if it made a difference and it didn't 🤷‍♀️

braker1nine:
> What do you think of opening a separate PR with the CI changes? We can merge that right away then I'll take a look at this one 🙏

Yeah I can absolutely do that.

> @NachoSoto we also need to address the fact that Mac Catalyst CI jobs are simply not working. IMO we should just disable the Mac Catalyst job until we have a better idea of what the issue is, as it's going to cause issues with all PRs.

@mluisbrown  Would it be worth doing this in that same PR?

mluisbrown:
> Would it be worth disabling Catalyst in that same PR?

@braker1nine yes, absolutely 👍

braker1nine:
Opened up #858. I didn't disable Catalyst in there, but I could add that

NachoSoto (approved):

PR #857, opened 2022-11-08
Add `async` helpers to Schedulers

p4checo:
Credits to @pointfreeco for the work done in:
https://github.com/pointfreeco/combine-schedulers

## Changes

- Add new `async` APIs to `DateScheduler` to enable waiting for them in asynchronous contexts:
  + `sleep(for:leeway:) async`
  + `sleep(until:leeway:) async`
  + `timer(interval:leeway:) async`

- Add new `async` variants `TestScheduler`'s `advance` and `run` APIs, to enable waiting for them in asynchronous contexts:
  + `advance()` async`
  + `advance(by:) async` (`DispatchTimeInterval` and `TimeInterval`)
  + `advance(to:) async`
  + `run() async`

- Refactor `TestScheduler.advance(to:)`'s loop condition to be based on `_currentDate` instead of `scheduledActions.count` to be consistent with the new `async` variant.

## Tests

- I've not added any tests for now, as I would like to know if the solution/approach is accepted first.
- Additionally, current Quick/Nimble setup apparently doesn't support async contexts yet as it's only available from Quick 6 (RAS is still on Quick 4). I can add them as vanilla async XCTests though.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 # master
 *Please add new entries at the top.*
 
+1. Add `async` helpers to Schedulers (#857, kudos to @p4checo)
 1. Add primary associated types to SignalProducerConvertible & SignalProducerProtocol (#855, kudos to @braker1nine)
-1. Refactor Github Actions to cover more swift versions (#858, kudos to @braker1nine)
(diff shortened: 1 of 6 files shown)


mluisbrown (approved):
Nice job 👍

PR #858, opened 2022-11-09
Refactor Github Actions to cover more swift versions

braker1nine:
- Test against Xcode 14.1 and 13.4.1
- Use setup-swift action to install specific swift versions (replaces swiftenv script)
- Test against Swift 5.2 and 5.7

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,7 +69,7 @@ jobs:
   swiftpm-linux:
     strategy:
       matrix:
-        swift: ["5.2", "5.6", "5.7"]
+        swift: ["5.2", "5.7"]
 
     name: SwiftPM Linux
     runs-on: ubuntu-18.04
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -79,6 +79,7 @@ jobs:
       - name: Install Swift
         run: |
           eval "$(curl -sL https://swiftenv.fuller.li/install.sh)"
+          swiftenv global ${{ matrix.swift }}
       - name: Checkout
         uses: actions/checkout@v2
       - name: Recover cached dependencies
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,6 +8,8 @@ jobs:
         fail-fast: false
         matrix:
           destination: [macOS, iOS, tvOS, watchOS]
+    env:
+      DEVELOPER_DIR: /Applications/Xcode_14.1.app/Contents/Developer
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,9 +8,6 @@ jobs:
         fail-fast: false
         matrix:
           destination: [macOS, iOS, tvOS, watchOS]
-          xcode: ["14.1", "13.4.1"]
-    env:
-      DEVELOPER_DIR: /Applications/Xcode_${{ matrix.xcode }}.app/Contents/Developer
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,10 +73,12 @@ jobs:
 
     name: SwiftPM Linux
     runs-on: ubuntu-18.04
+    env:
+      SWIFT_VERSION: ${{ matrix.swift }}
     steps:
-      - uses: swift-actions/setup-swift@v1
-        with:
-          swift-version: ${{ matrix.swift }}
+      - name: Install Swift
+        run: |
+          eval "$(curl -sL https://swiftenv.fuller.li/install.sh)"
       - name: Checkout
         uses: actions/checkout@v2
       - name: Recover cached dependencies
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,6 +8,9 @@ jobs:
         fail-fast: false
         matrix:
           destination: [macOS, macCatalyst, iOS, tvOS, watchOS]
+          xcode: ["14.1", "13.4.1"]
+    env:
+      DEVELOPER_DIR: /Applications/Xcode_${{ matrix.xcode }}.app/Contents/Developer
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


braker1nine:
A couple questions I had on these...
1. Is running against 5.2, 5.6, and 5.7 excessive? Would it be sufficient to just run against 5.2 and 5.7?
2. We could replace "5.7" with "5", which would use the latest 5.X version. Does that feel more maintainable? Or is it best to keep using an explicit version?

mluisbrown:
> Is running against 5.2, 5.6, and 5.7 excessive? Would it be sufficient to just run against 5.2 and 5.7?

IMO just 5.2 and 5.7 is sufficient. Probably best to keep it explicit.

.github/workflows/test.yml
```diff
@@ -64,12 +67,16 @@ jobs:
             ONLY_ACTIVE_ARCH=YES
 
   swiftpm-linux:
+    strategy:
+      matrix:
+        swift: ["5.2", "5.6", "5.7"]
```
mluisbrown, line 72: Let's change this to just `5.2` and `5.7` (explicitly so it's clear which exact versions a being used). Otherwise looks great 👍

mluisbrown (approved):

braker1nine:
Well that's weird. 

Removed Mac Catalyst. All tests pass
Removed Swift 5.6. Swift 5.2 tests fail

mluisbrown:
It looks like the previous `swiftenv` solution specifically installed shims and put them on the `PATH`. If you set the `SWIFT_VERSION` environment variable before running the swiftenv script it will install a specific version, so you could still use `swiftenv` to install different versions on Linux.

braker1nine:
> Could this be related to the change in the way that Swift is being installed on Linux? The only link I found related to this is this, and it's not particularly helpful: https://forums.swift.org/t/error-no-such-module-swiftdispatchoverlayshims/34771

Ah wonderful. Nothing like a flaky build ...

> If you set the SWIFT_VERSION environment variable before running the swiftenv script it will install a specific version, so you could still use swiftenv to install different versions on Linux.

I didn't realize that was a feature of swiftenv. I'll roll it back to that and give it a try. When I was toying around with swiftenv locally I noticed it hadn't been updated in quite a while and it's definitions of swift versions had almost nothing post 5.0 😓

braker1nine:
Maybe it's just generally flaky, but I do wonder if it has something to do with the way the builds are being cached. 

Anecdotally, it worked fine when we didn't have a cache key hit the first time. And failed the second time when we did and copied the cached build files in `.build`

mluisbrown:
@NachoSoto it looks like CI is now passing, but the "expected checks" configured for the repo no longer match the actual checks that are run. I'm not an admin on the repo so I can't change this.

braker1nine:
Tests passed. But it's not actually using Swift 5.2. It's installing it but still using the built in Swift 5.7

mluisbrown:
I have a local Linux VM, I'll see if I can get this working there, selecting the correct version of Swift. Testing this via CI is going to take forever!

braker1nine:
> I have a local Linux VM, I'll see if I can get this working there, selecting the correct version of Swift. Testing this via CI is going to take forever!

😅 yeah it is. I was close to commenting out all the other tests until I could get this one working. Thanks @mluisbrown! Ping me if there's anything I can do to help with that

mluisbrown:
Ok, I wasn't able to make any progress. I'm on an M1 Mac and on my Ubuntu VM swiftenv is not getting the correct download links for ARM64 versions of Swift. In any case, I don't think there are ARM64 versions of Swift 5.2 so I wouldn't be able to test it. Maybe using `swiftenv` was a bad idea, but I can't think of a solution :/  

Do we really need to test on Swift 5.2?

braker1nine:
> Do we really need to test on Swift 5.2?

Yeah I suppose that's an interesting question. The reason I added it was because the library wasn't successfully building with Swift 5.2. Currently we've got 5.2 listed as the swift tools version in the package manifest. It just seems to me that we should be ensuring that any changes to the library build on whatever swift version we've specified there?

I wonder if it would be worth going back to the setup-swift action and removing the caching to see if it builds more consistently. The caching doesn't seem to be saving a TON of time from what I can tell? And frequently misses

mluisbrown:
Ok, maybe try the setup-swift action again 👍

braker1nine:
Just pushed it with setup swift. I thought about this as I was making the change... 

The swift version needs to be included in the cache key. I doubt swift would be happy using cache build/dependency files from different versions of Swift? 🤞

braker1nine:
I do wonder if testing on 13.4.1 and 14.1 on Xcode is overkill. The main purpose when I did what was to hit versions of swift with primary associated types and without them. Since we have the linux SPM hitting multiple versions of swift. Maybe having MacOS just hit each target platform would be sufficient.

mluisbrown:
Nice, CI passed! 

> I do wonder if testing on 13.4.1 and 14.1 on Xcode is overkill
Probably. 

> Maybe having MacOS just hit each target platform would be sufficient.
I would choose iOS instead of macOS, as it's almost certainly the most used platform for this repo.

In the meantime we need an Admin to alter the expected checks for CI.

.github/workflows/test.yml
```diff
@@ -8,9 +8,6 @@ jobs:
         fail-fast: false
         matrix:
           destination: [macOS, iOS, tvOS, watchOS]
-          xcode: ["14.1", "13.4.1"]
-    env:
-      DEVELOPER_DIR: /Applications/Xcode_${{ matrix.xcode }}.app/Contents/Developer
```
mluisbrown: I would change this to fix the version of Xcode at 14.1, so that we are always in control of the version, and not caught out if/when GitHub updates the version on CI.

So we could keep:
```yaml
    env:	
       DEVELOPER_DIR: /Applications/Xcode_14.1.app/Contents/Developer
```

braker1nine:
Ugh. Bummer. Pulled from cache. Build failed 😞

mluisbrown:
Let's get rid of the caching for the Linux jobs then 👍

NachoSoto:
Oh I thought I was an admin, but I guess I'm not so I can't change the required tests.

NachoSoto (approved):

mluisbrown:
@andersio or @RuiAAPeres can one of you change the CI checks to match the new ones (see the checks that passed) and then merge please? 🙏

PR #859, opened 2022-11-17
Fixes some issues related to locking.

mluisbrown:
* Fixes issues related to availability of `withLock`
* Simplified OS checking around the use of `OSAllocatedUnfairLock` now that macOS 13 and Xcode 14 have shipped.
* Bump min OS versions to iOS 10, macOS 10.12, tvOS 10, watchOS 3 so we can remove some cumbersome availability checks.

#### Checklist
- [x] Updated CHANGELOG.md.


Sources/Atomic.swift
```diff
@@ -140,34 +136,31 @@ internal class Lock: LockProtocol {
 	}
 	#endif
 
-	#if compiler(>=5.7)
-		#if !targetEnvironment(macCatalyst)
```
mluisbrown, line 144: Now that macOS 13 is released, we can remove this silly macCatalyst dance and simplify the conditions a lot.

NachoSoto (approved):

RuiAAPeres (approved):

PR #860, opened 2022-11-21
Guard use of OSAllocatedUnfairLock with a Swift 5.7 version check to allow building on Xcode 13.x

mluisbrown:
The changes made in #859 went a little too far in removing the Swift version check, without which building fails on Xcode 13.x, so I added it back. 

* Also updated the Schemes to not use deprecated options.
* Also made some minor CI fixes to avoid warnings.


mluisbrown:
Closing as unable to resolve CI issues 😕 Will create a new PR with just the code change.

PR #866, opened 2023-01-05
Explicit extension safety flag

lickel:
This explicitly declares that ReactiveSwift only uses extension safe APIs (and/or APIs in an extension-safe way).
The code was already extension safe; this just forward declares it and enforces it at compile time.

If I can enforce this here, then I can enforce it in [Workflow](https://github.com/square/Workflow-swift) which is heavily dependent on this library.
Without this explicit conformance, it is not straightforward to write CI validation that these libraries remain extension safe.

There is zero harm in explicit conformance. Non-extension applications can call into extension-safe libraries without any restrictions.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 *Please add new entries at the top.*
 
 1. Bumped deployment target to iOS 11, tvOS 11, watchOS 4, macOS 10.13, per Xcode 14 warnings
-1. Explicitly declare `APPLICATION_EXTENSION_API_ONLY`
+1. Explicitly declare `APPLICATION_EXTENSION_API_ONLY` for CocoaPods
 
 # 7.1.0
 1. Add CI Release jobs on tag push (#862, kudos to @p4checo)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 *Please add new entries at the top.*
 
 1. Bumped deployment target to iOS 11, tvOS 11, watchOS 4, macOS 10.13, per Xcode 14 warnings
+1. Explicitly declare `APPLICATION_EXTENSION_API_ONLY`
 
 # 7.1.0
 1. Add CI Release jobs on tag push (#862, kudos to @p4checo)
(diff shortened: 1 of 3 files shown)


ReactiveSwift.xcodeproj/project.pbxproj
```diff
@@ -1512,6 +1512,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = D047262919E49FE8006002AA /* Debug.xcconfig */;
 			buildSettings = {
+				APPLICATION_EXTENSION_API_ONLY = YES;
```
NachoSoto, line 1515: Which target did you set this on? It's already `YES` on each of the 4 targets.
Screenshot 2023-01-05 at 10 13 16

lickel: Oh sorry, I didn't have the XCConfig submodule setup correctly. I've reverted this.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 *Please add new entries at the top.*
 
 1. Bumped deployment target to iOS 11, tvOS 11, watchOS 4, macOS 10.13, per Xcode 14 warnings
+1. Explicitly declare `APPLICATION_EXTENSION_API_ONLY`
```
NachoSoto, line 5: Cool, maybe mention that this is specific to the `CocoaPods` installation?

NachoSoto (approved):

CHANGELOG.md
lickel: Done!

lickel:
@NachoSoto once this lands, would you be willing to publish a new release? 

That way I would be unblocked in making/landing the change to that other repo I mentioned.
And, thanks!

NachoSoto:
I don't think I have privileges to create a CocoaPods release, sorry. But hopefully somebody else can help with that.

lickel:
CC @mluisbrown @andersio re: the possibility of publishing a new release

mluisbrown:
@lickel 7.1.1 has been release. I did notice (too late to do anything about it) that the Podspec still has iOS 10 as the minimum deployment target, so you will still get warnings about that.

lickel:
Thanks! And ugh whoops sorry about missing that one.

PR #869, opened 2023-01-06
Fix CI release git tag push trigger

p4checo:
Apparently GH action's push event filters accept glob patterns and *not* regexes :facepalm:

Source:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore

That means that the existing tag filter string being a regex wasn't working properly and as such wasn't triggering the expected builds on tag push.

To address this, the `on.push.tags` were updated to have separate entries for normal vs `rc` tags, and also tweaked the tag-only jobs' filters to rely on `github.ref_type` instead of inspecting `github.ref`.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -3,7 +3,8 @@ on:
     branches:
       - master
     tags:
-      - '[0-9]+\.[0-9]+\.[0-9]+(-rc\.[0-9]+)*'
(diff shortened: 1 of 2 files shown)


mluisbrown (approved):
You have conflicts.

PR #871, opened 2023-04-10
Remove Result mention from Playground

michalsrutek:
As https://github.com/ReactiveCocoa/ReactiveSwift/pull/702 has been merged a looong time ago, the mention of `Result` should be omitted.


mluisbrown (approved):
Good spot 👍

PR #875, opened 2023-06-24
Add support for `xrOS`

NachoSoto:
#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # master
 *Please add new entries at the top.*
 
+1. Add support for VisionOS (#875, kudos to @NachoSoto)
 1. Fix minimum deployment target of iOS 11 in CocoaPods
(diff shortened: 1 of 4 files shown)


braker1nine:
@NachoSoto would it be preferable to use `visionOS` in place of `xrOS`? Since that's how Apple refers to it in Xcode. Seems like this would result in less confusion?

Screenshot 2023-06-29 at 4 53 47 PM

NachoSoto:
It might change in a future beta, but as of now `xrOS` is the name of the platform (similar to `iOS`/`watchOS`/`tvOS`/`macOS`.

braker1nine:
Ah I see what you mean. I can't say that doesn't make sense.

I hope that changes. I'm sure we'll all figure it out. But I feel like we're about to see schism in how different libraries refer to it

mluisbrown (approved):

PR #878, opened 2023-09-14
Use binary search for removing elements from Bag

nickoto:
Given that tokens are always added to the end of the array and have a monotonically increasing value, this list is always sorted, so we can use a binary search to improve performance if this list gets large.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -68,7 +68,7 @@ jobs:
   swiftpm-linux:
     strategy:
       matrix:
-        swift: ["5.2", "5.7"]
+        swift: ["5.7"]
 
     name: SwiftPM Linux
     runs-on: ubuntu-22.04
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -71,7 +71,7 @@ jobs:
         swift: ["5.2", "5.7"]
 
     name: SwiftPM Linux
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-22.04
     steps:
       - name: Setup Swift version
         uses: swift-actions/setup-swift@v1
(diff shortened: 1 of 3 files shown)


mluisbrown:
@nickoto did you run any tests to validate the performance improvement, and how significant it is? It would be useful to have some metrics.

nickoto:
Yes; in our particular scenario instruments was showing that for this particular operation in our app that `Bag.remove` was by far the hot path, taking several seconds to complete removing a large number of elements from the bag. After this change the removal process time went to several milliseconds in the exact same scenario. (In this case, the bag contained thousands of elements and the operation was removing half of them at once in no particular order).

While I can't test this over all workloads, in any situation where many items would be removed from the bag at once e.g. a large teardown, this would be a benefit. In situations where the bag is very tiny the complexity cost is negligible.

mluisbrown (approved):
Looks like a good improvement 👍

mluisbrown (approved):

nickoto:
@mluisbrown Since the runner for ubuntu 18.04 is deprecated we can no longer run the SwiftPM tests for 5.2 as there is no binary release of swift 5.2 on that platform (it only has releases for 16.04 and 18.04). That's why the log is showing a 404 - it fails to download a release that doesn't exist.

Not sure why the 5.7 runner was cancelled though -- it looks like it was passing up until it cancelled.

mluisbrown:
@nickoto We should drop the 5.2 tests then. I'm not sure if you have the rights to do that. If not let me know and I'll sort it out.

nickoto:
I can drop it from the yml but I'm not sure if there are other steps required.

Issue #887, opened 2025-01-15
Add primary associated types to applicable types


braker1nine:
These were added to `SignalProducerConvertible` and `SignalProducerProtocol` in https://github.com/ReactiveCocoa/ReactiveSwift/pull/855

But I think it would be nice to add them to the property types. The list that's in my head is...
- `PropertyProtocol`
- `MutablePropertyProtocol`
- `ComposableMutablePropertyProtocol`?
- `BindingTargetProvider`
- `BindingSource`?

Open to any others I might be missing...

PR #888, opened 2025-01-18
Add primary associated types to property and binding protocols

braker1nine:
Including the following protocols 
- PropertyProtocol
- MutablePropertyProtocol
- ComposableMutablePropertyProtocol
- BindingSource
- BindingTargetProvider

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/Sources/UnidirectionalBinding.swift
+++ b/Sources/UnidirectionalBinding.swift
@@ -10,17 +10,31 @@ precedencegroup BindingPrecedence {
 
 infix operator <~ : BindingPrecedence
 
+#if swift(>=5.7)
+/// Describes a source which can be bound.
+public protocol BindingSource<Value>: SignalProducerConvertible where Error == Never {}
+#else
 /// Describes a source which can be bound.
 public protocol BindingSource: SignalProducerConvertible where Error == Never {}
+#endif
 extension Signal: BindingSource where Error == Never {}
 extension SignalProducer: BindingSource where Error == Never {}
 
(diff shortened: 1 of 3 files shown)


braker1nine:
I'm still very irritated that I don't have a solution that doesn't involve duplicating everything in the protocol...

braker1nine:
Resolves #887

Sources/UnidirectionalBinding.swift
```diff
@@ -10,17 +10,31 @@ precedencegroup BindingPrecedence {
...
+public protocol BindingSource<Value>: SignalProducerConvertible where Error == Never {}
+#else
 /// Describes a source which can be bound.
 public protocol BindingSource: SignalProducerConvertible where Error == Never {}
+#endif
 extension Signal: BindingSource where Error == Never {}
 extension SignalProducer: BindingSource where Error == Never {}
 
+#if swift(>=5.7)
+/// Describes an entity which be bond towards.
```
mluisbrown, line 24:
```suggestion
/// Describes an entity which can be bound towards.
```

This typo is also in the original below

mluisbrown (approved):
Looks good  👍 . If you could just fix the one typo (which already existed in the original definition).

Sources/UnidirectionalBinding.swift
braker1nine: Done. Fixed it in a couple other spots that seemed incorrect as well

